PR #185, opened 2020-04-16
Preferably compact SSTables with the range tombstones that are being read the most

j-baker:
STCS has a notion of hotness - it first compacts the tables that are
being read the most, because it assumes that compacting them will lead
to the biggest win.

LCS does not do this, and most of our issues at this point relate to
range tombstones being read.

Range tombstones can only really be dropped when they reach the highest
level, so when there's a really hot SSTable in a low level, there's no
real forcing function for C* to improve things.

Here, we alter the compaction strategy (behind a feature flag). We say:

1. Let's look for the top 10 SSTables across all keyspaces which are
responsible for the most tombstone reads (asserting that we must have
read at least 1MM tombstones from each table for it to be considered).
2. When we try to compact a table, if it's in this list of worst tables,
we choose to compact that one first, because it's likely we'll get a
bigger win from compacting it.
3. If in Level 0, we look for a level 0 task, and then compact this with
that.
4. If in level n (n not largest), we look for overlapping tables in the
next level, and compact with those, compacting into n+1.
5. If in the top level, we basically just compact the table with itself.

When the compaction is done, another compaction will be triggered in
order to compact any _other_ sstables that need compacting.

### DIFF
--- a/src/java/com/palantir/cassandra/TombstoneHotness.java
+++ b/src/java/com/palantir/cassandra/TombstoneHotness.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


Sam-Kramer (changes requested):
A couple of Q's/suggestions, I'd be happy to try and implement them if you'd like @j-baker on a separate branch of this.

src/java/com/palantir/cassandra/TombstoneHotness.java
```diff
@@ -0,0 +1,92 @@
...
+        for (int i = 0; i < NUMBER_OF_TABLES_ELIGIBLE_FOR_COMPACTION && !sstables.isEmpty(); i++) {
+            result.add(sstables.remove().getKey());
+        }
+        return result.build();
+    }
+
+    public static Set<SSTableReader> maybeGetHotSstables(Iterable<SSTableReader> readers) {
+        Set<SSTableReader> result = new HashSet<>();
+        Set<SSTableReader> hottest = hottestSstables.get();
+        for (SSTableReader reader : readers) {
```
Mattheo28, line 74: why bother taking in SStables here and looping through them, if we always call it with all SStables? Can't we save ourselves this extra loop through all SStables and just return `hottestSstables`?

PR #208, opened 2020-09-01
Fix CQLSH

Sam-Kramer:
CQLSH script has left over params from the C* 2.x days


leonz:
Should we submit this upstream instead?

Sam-Kramer:
Pls ignore, this was due to mixing of bad JARs D:

PR #210, opened 2020-09-02
Publish script for mavenLocal

Sam-Kramer:
Makes things easier :)


leonz (approved):

leonz:
Might want to add `set -e` to fail on error.

PR #214, opened 2020-09-28
Only allow bootstrap when signal received through API

Mattheo28:
We want to only allow bootstraps to begin once we have received a signal through the API that we can begin, to prevent manual bootstraps.

Would suggest going through this commit by commit, as the first commit has the actual logic, and the second commit is just fixing the tests to call the method to allow for bootstrapping.

### DIFF
--- a/palantir-cassandra-jmx-api/build.gradle
+++ b/palantir-cassandra-jmx-api/build.gradle
@@ -38,10 +38,3 @@ task copySourceFiles(type: Copy) {
 
 tasks.withType(JavaCompile).all { dependsOn copySourceFiles }
 
-repositories {
-    jcenter()
-}
-
-dependencies {
-    compile 'com.google.guava:guava:18.0'
-}
--- a/palantir-cassandra-jmx-api/build.gradle
+++ b/palantir-cassandra-jmx-api/build.gradle
@@ -38,3 +38,10 @@ task copySourceFiles(type: Copy) {
 
 tasks.withType(JavaCompile).all { dependsOn copySourceFiles }
 
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile 'com.google.guava:guava:18.0'
+}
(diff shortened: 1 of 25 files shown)


src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -723,4 +731,50 @@
...
+                monitor.leave();
+            }
+        }
+
+        public void awaitBootstrappable()
+        {
+            boolean isBootstrappable;
+            try
+            {
+                isBootstrappable = monitor.enterWhen(isAllowedToBootstrap, 5, TimeUnit.MINUTES);
```
leonz: I would propose that we just block indefinitely, rather than wait 5 minutes then fail.

palantir-cassandra-jmx-api/build.gradle
```diff
@@ -38,3 +38,10 @@ task copySourceFiles(type: Copy) {
 
 tasks.withType(JavaCompile).all { dependsOn copySourceFiles }
 
+repositories {
```
leonz: We should move BootstrapManager into an internal class so that this dependency isn't exposed on the API.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -819,6 +820,8 @@ private void joinTokenRing(int delay) throws ConfigurationException
         boolean dataAvailable = true; // make this to false when bootstrap streaming failed
         if (shouldBootstrap())
         {
+            setMode(Mode.WAITING_TO_BOOTSTRAP, "Awaiting start bootstrap call", true);
```
leonz, line 823: Does this block new clusters from starting up?  I'm worried based on tests that the answer is yes...which we don't want.

Mattheo28: Nop, because we only wait for the signal after the `shouldBootstrap()` call, which is false for new clusters because the nodes are in the seeds

zpear (approved):

leonz (approved):

PR #216, opened 2020-10-12
[Cassandra 3] Cheap Reads

leonz:
Port of https://github.com/palantir/cassandra/pull/9
Original author: @j-baker

### DIFF
--- a/src/java/org/apache/cassandra/db/SinglePartitionReadCommand.java
+++ b/src/java/org/apache/cassandra/db/SinglePartitionReadCommand.java
@@ -428,6 +428,11 @@ public class SinglePartitionReadCommand extends ReadCommand
         return clusteringIndexFilter;
     }
 
+    @Override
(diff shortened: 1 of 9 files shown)


src/java/org/apache/cassandra/db/SinglePartitionReadCommand.java
```diff
@@ -428,6 +428,11 @@ public ClusteringIndexFilter clusteringIndexFilter(DecoratedKey key)
         return clusteringIndexFilter;
     }
 
+    @Override
+    public boolean isCheap() {
+        return limits().count() <= 1;
```
leonz, line 433: The only notable change in this port is that cheapness is determined by the new limit feature (rather than pulling a count from the filter).

Sam-Kramer:
LGTM

Sam-Kramer (approved):

PR #218, opened 2020-11-02
Allow snapshots during bootstrap

zpear:
Allow snapshots during bootstrap to support incremental upload of bootstrapping node immutable files during backup

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -1365,6 +1365,11 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
         setMode(Mode.NORMAL, false);
     }
 
+    @VisibleForTesting
+    void setOperationModeJoining() {
+        setMode(Mode.JOINING, false);
+    }
+
     @Override
     public Set<Map<String, String>> getNonTransientErrors() {
         return ImmutableSet.copyOf(nonTransientErrors);
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -171,7 +171,8 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     /* the probability for tracing any particular request, 0 disables tracing and 1 enables for all */
     private double traceProbability = 0.0;
 
-    private static enum Mode { STARTING, NORMAL, JOINING, LEAVING, DECOMMISSIONED, MOVING, DRAINING, DRAINED, ZOMBIE, NON_TRANSIENT_ERROR, WAITING_TO_BOOTSTRAP }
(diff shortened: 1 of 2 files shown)


test/unit/org/apache/cassandra/service/StorageServiceServerTest.java
```diff
@@ -122,9 +136,41 @@ public void testGetAllRangesEmpty()
...
+
+        InetAddress bootstrappingNode = hosts.get(1);
+        Collection<Token> bootstrapTokens = Collections.singleton(keyTokens.get(3));
+        Gossiper.instance.injectApplicationState(bootstrappingNode, ApplicationState.TOKENS, valueFactory.tokens(bootstrapTokens));
+        ss.onChange(bootstrappingNode,
+                    ApplicationState.STATUS,
+                    valueFactory.bootstrapping(bootstrapTokens));
+
+        assertFalse(tmd.isMember(bootstrappingNode));
+        assertFalse(tmd.isLeaving(bootstrappingNode));
```
Mattheo28: what are we doing/checking from line 159 to here? I'm confused why we need this and not just the last 3 lines

zpear: I was following the example of this test to alter the token metadata with a bootstrap instead of solely changing the operation mode: https://github.com/palantir/cassandra/blob/8a1cebc8c581d6935d7f00cee2b171259954c224/test/unit/org/apache/cassandra/service/LeaveAndBootstrapTest.java#L449

Do you think the test would be fine without this/only changing operation mode?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2812,7 +2818,7 @@ public void takeSnapshot(String tag, String... keyspaceNames) throws IOException
      */
     private void takeSnapshot(String tag, boolean ephemeral, String... keyspaceNames) throws IOException {
         if (operationMode == Mode.JOINING)
-            throw new IOException("Cannot snapshot until bootstrap completes");
+            logger.info("Taking snapshot of joining node");
```
leonz: Make these messages a warn, with a notice that these snapshots are not complete / valid for a live cluster.

test/unit/org/apache/cassandra/service/StorageServiceServerTest.java
Mattheo28: If you try just keeping the last 3 lines, and reverting your changes in the code back to throwing an error if the node is bootstrapping, does this test fail? If so I think we can remove this stuff. If it passes then I guess we need this stuff to get to that part of the code. Does that make sense?

Mattheo28:
lgtm, will let @leonz approve

leonz (approved):
LGTM.

PR #225, opened 2020-11-05
Log high memory usage on cell collection

leonz:
While iterating through a collection of cells, count the accumulated amount of memory consumed by the cells, and log if it breaches a threshold.  The goal is to log collections that may be contributing to heavy heap usage and garbage collection problems.

### DIFF
--- a/src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
+++ b/src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
@@ -280,6 +280,7 @@ public class SliceQueryFilter implements IDiskAtomFilter
 
         boolean hasBreachedCollectionThreshold = false;
         long dataSizeCollected = 0;
+        long metadataSizeCollected = 0;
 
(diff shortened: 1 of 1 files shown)


Sam-Kramer (approved):
LGTM, and would've been awesome for PDS-144078! 

A future goal would be try to figure out if we could figure out the number of objects as well

leonz:
Just pushed up a commit to actually parse the threshold the right way.  Sample output when the threshold is 1 MB:


> WARN  [2020-11-05T15:28:29.734Z] org.apache.cassandra.db.filter.SliceQueryFilter: Breached memory threshold while collecting cells for keyspace/cf (system,schema_columnfamilies) and key DecoratedKey(6d616772697474655f636f6f72645f316738746932735f616c7461, 6d616772697474655f636f6f72645f316738746932735f616c7461); data size: 205925; metadata size: 842976 (0: (system,schema_columnfamilies), 1: DecoratedKey(6d616772697474655f636f6f72645f316738746932735f616c7461, 6d616772697474655f636f6f72645f316738746932735f616c7461), 2: 205925, 3: 842976)

PR #228, opened 2020-11-12
Add consistency level write metrics

zpear:
Extending ClientRequestMetrics to recording metrics per consistency level can help diagnose issues between normal writes and deletes


leonz (approved):
lgtm!

PR #229, opened 2020-11-12
Thread watcher

Sam-Kramer:
Watch LocalRunnables, and forcefully interrupt them if they're taking too long.


Sam-Kramer:
While an interesting idea, C* does not like threads being interrupted randomly, and caused memory leaks.

PR #230, opened 2020-11-17
Monitor locally ran reads to properly report timeouts

Sam-Kramer:
Local reads do not timeout, and so in case of catastrophic disk failure (i.e. every read hangs on IO access), the threads never report their latency, despite technically breaching the timeout threshold. 

This change has an independent thread which monitors all reads, and will count the timeout latency if it has not completed running within the configured timeout.

### DIFF
--- a/src/java/com/palantir/cassandra/concurrent/LocalReadRunnableTimeoutWatcher.java
+++ b/src/java/com/palantir/cassandra/concurrent/LocalReadRunnableTimeoutWatcher.java
@@ -43,7 +43,7 @@ public class LocalReadRunnableTimeoutWatcher implements Runnable
         readCommandStartTimes.put(readCommand, System.currentTimeMillis());
     }
 
-    public long getTimeout() {
+    private long getTimeout() {
         return DatabaseDescriptor.getReadRpcTimeout();
     }
 
--- a/src/java/com/palantir/cassandra/concurrent/LocalReadRunnableTimeoutWatcher.java
+++ b/src/java/com/palantir/cassandra/concurrent/LocalReadRunnableTimeoutWatcher.java
@@ -21,9 +21,7 @@ package com.palantir.cassandra.concurrent;
 import java.util.ArrayList;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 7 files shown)


src/java/org/apache/cassandra/db/Keyspace.java
```diff
@@ -373,6 +375,11 @@ public Row getRow(QueryFilter filter)
     {
         ColumnFamilyStore cfStore = getColumnFamilyStore(filter.getColumnFamilyName());
         ColumnFamily columnFamily = cfStore.getColumnFamily(filter);
+        int readDelay = DatabaseDescriptor.getReadDelay();
```
Sam-Kramer: Might make sense to leave this from this PR; basically the read delay being set is only network level, and is not storage level. Thus, dynamic snitch can never function properly...

Sam-Kramer: This fixes that by making the pause in the read code

Sam-Kramer: I'm moving this in https://github.com/palantir/cassandra/pull/232

src/java/com/palantir/cassandra/concurrent/LocalReadRunnableTimeoutWatcher.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(LocalReadRunnableTimeoutWatcher.class);
+
+    private LocalReadRunnableTimeoutWatcher() { }
+
+    public void watch(ReadCommand readCommand) {
+        logger.trace("Watching read command {} for timeout {}", readCommand, getTimeout());
+        readCommandStartTimes.put(readCommand, System.currentTimeMillis());
+    }
+
+    public long getTimeout() {
```
Mattheo28: this can be private

Mattheo28 (commented):
this makes sense, just a nit! Have you tested on IL?

Sam-Kramer:
> this makes sense, just a nit! Have you tested on IL?

Yup, tested on IL, and confirmed it works as expected when experiencing SND.

Mattheo28 (approved):

leonz (approved):

PR #231, opened 2020-11-19
Count ranged tombstones

Sam-Kramer:


src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -273,6 +275,7 @@ public void collectReducedColumns(ColumnFamily container, Iterator<Cell> reduced
     {
         reducedCells = CountingCellIterator.wrapIterator(reducedColumns, now, gcBefore);
         columnCounter = columnCounter(container.getComparator(), now);
+        rangeTombstoneCounter = RangeTombstoneCounter.create(gcBefore, container.deletionInfo());
```
Sam-Kramer, line 278: Note -- we cannot use cell counting iterator as ranged tombstones are not treated as cells (and our filtered out prematurely). I thought this would be the least invasive with minimal perf degradation that we could do.

src/java/com/palantir/cassandra/utils/RangeTombstoneCounter.java
```diff
@@ -0,0 +1,62 @@
...
+    private final int rangeTombstones;
+
+    public static final RangeTombstoneCounter EMPTY = new RangeTombstoneCounter(0, 0);
+
+    public RangeTombstoneCounter(int droppableRangeTombstones, int rangeTombstones) {
+        this.droppableRangeTombstones = droppableRangeTombstones;
+        this.rangeTombstones = rangeTombstones;
+    }
+
+    public static RangeTombstoneCounter create(int gcBefore, DeletionInfo deletionInfo) {
```
leonz, line 38: Are we positive that this works?  I would expect that after exhausting the `container.deletionInfo()` iterator any further iterations would have no results.

I'm also generally nervous about this approach.  Other iterators "wrap" because it's very expensive - O(n) - to iterate the cells more than once.

I think you'll need to follow the model of the `CountingCellIterator` (where you use a delegator but also add additional functionality), and then wrap the location where the tombstone iterator is _actually_ called.

Sam-Kramer: Yup, iterator() is immutable as seen here: https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/db/RangeTombstoneList.java#L367

We do already perform an O(N) (worst-case) here: https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/db/RangeTombstoneList.java#L902

However, I think it sounds like we'd rather have something "more invasive" than computational costs, so I'll get back to the drawing board.

leonz: How many times are we doing N though?  I'm unconfident about wanting to make it 2 or 3 x N

Sam-Kramer: Yeah, it's at least 2N :(

Sam-Kramer: The other PR is here: https://github.com/palantir/cassandra/pull/233

PR #233, opened 2020-11-25
(Cheap) Count ranged tombstones 

Sam-Kramer:


src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -466,7 +470,7 @@ public boolean hitTombstoneWarnThreshold() {
 
     public int lastTombstones()
     {
-        return columnCounter == null ? 0 : columnCounter.tombstones();
+        return columnCounter == null ? 0 : columnCounter.tombstones() + rangeTombstoneCounter.getCount();
```
leonz: I would prefer to separate this into its own variable / metric, because it allows us to:
1) clearly see if there are errors in our methodology
2) easily differentiate between range and non-range tombstones in analysis
3) compare future tombstone metrics to past

src/java/com/palantir/cassandra/utils/RangeTombstoneCountingIterator.java
```diff
@@ -0,0 +1,67 @@
...
+
+package com.palantir.cassandra.utils;
+
+import java.util.Iterator;
+
+import org.apache.cassandra.db.ColumnFamily;
+import org.apache.cassandra.db.OnDiskAtom;
+import org.apache.cassandra.db.RangeTombstone;
+
+public class RangeTombstoneCountingIterator implements Iterator<OnDiskAtom>
```
leonz, line 32: Overall looks much better.  Is it reasonable to add a test?

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
Sam-Kramer: 👍

src/java/com/palantir/cassandra/utils/RangeTombstoneCountingIterator.java
Sam-Kramer: 👍

Sam-Kramer:
Explaining the metrics for future observers: 

In Cassandra 2.x, range tombstones are treated differently than cells, where as in C* 3.x, range tombstones are treated as cells. This is important detail, as the iterators used when counting/viewing cells, do not contain range tombstones, as they're not cells. Thus, we've to wrap the iterator which reads from disk, but then we're only getting tombstones from disk -- leaving any tombstones in a memtable ignored. 

So we end up with three different metrics:
(1) RangeTombstonesReadHistogram: Total number of range tombstones read from disk
(2) DroppableRangeTombstonesReadHistogram: Total number of _droppable_ range tombstones read from disk
(3) RangeTombstonesHistogram: Total number of range tombstones, both memory and disk. 

Which gives us the following: RangeTombstonesHistogram >= RangeTombstonesReadHistogram >=  DroppableRangeTombstonesReadHistogram

We cannot figure out the number of droppable tombstones for both memory/disk as we'd need to do an O(N) scan (N being number of range tombstones) to determine which are droppable -- there's also no convenient place to put this in the code where we are already performing the scan.

leonz (approved):
Looks reasonable! Would like to see this deployed to our internal development stack as an RC to confirm that it doesn't cause significant performance issues + gives accurate results before releasing generally.

PR #235, opened 2020-11-30
Shutdown node after disk usage exceeds threshold

zpear:
Previously if a node was using too much disk space it would have to be manually monitored and shut down. Replacing this by triggering a shutdown with a configurable `max_disk_utilization` threshold increases safety.

### DIFF
--- a/src/java/org/apache/cassandra/db/Directories.java
+++ b/src/java/org/apache/cassandra/db/Directories.java
@@ -62,6 +62,8 @@ import org.apache.cassandra.io.sstable.Descriptor;
 import org.apache.cassandra.io.sstable.SSTable;
 import org.apache.cassandra.io.sstable.SnapshotDeletingTask;
 import org.apache.cassandra.io.util.FileUtils;
+import org.apache.cassandra.service.StorageService;
+import org.apache.cassandra.service.StorageServiceMBean;
 import org.apache.cassandra.utils.ByteBufferUtil;
 import org.apache.cassandra.utils.FBUtilities;
 import org.apache.cassandra.utils.Pair;
@@ -318,11 +320,31 @@ public class Directories
         logger.debug(String.format("Highest data directory disk utilization on path %s (%f)",
                                    maxPathToUtilization.getKey().location, maxPathToUtilization.getValue()));
         double threshold = DatabaseDescriptor.getMaxDiskUtilizationThreshold();
-        if (maxPathToUtilization.getValue() > threshold)
+        double currUtilization = maxPathToUtilization.getValue();
+        if (currUtilization >= threshold)
         {
             ExceededDiskThresholdException cause = new ExceededDiskThresholdException(
-                                    maxPathToUtilization.getKey().location, maxPathToUtilization.getValue(), threshold);
+                                    maxPathToUtilization.getKey().location, currUtilization, threshold);
             throw new FSWriteError(cause, maxPathToUtilization.getKey().location);
+        }
+
+        if (currUtilization < threshold && StorageService.instance.isNodeDisabled() && StorageService.instance.isSetupCompleted())
+        {
+            Set<Map<String, String>> nonTransientErrors = StorageService.instance.getNonTransientErrors();
+            boolean isOnlyExceededDisk = true;
+            for (Map<String, String> nonTransientError : nonTransientErrors)
+            {
+                String errorType = nonTransientError.get(StorageService.NON_TRANSIENT_ERROR_TYPE_KEY);
+                isOnlyExceededDisk &= errorType.equals(StorageServiceMBean.NonTransientError.EXCEEDED_DISK_THRESHOLD.toString());
+            }
+            if (isOnlyExceededDisk)
+            {
+                logger.info(String.format(
+                    "Only non transient errors were of type ExceededDiskThreshold. Current disk use of %f is under threshold " +
+                    "of %f. Clearing non transient errors and enabling node", currUtilization, threshold));
+                StorageService.instance.clearNonTransientErrors();
+                StorageService.instance.enableNode();
+            }
         }
     }
 
--- a/src/java/org/apache/cassandra/db/Directories.java
+++ b/src/java/org/apache/cassandra/db/Directories.java
@@ -67,6 +67,8 @@ import org.apache.cassandra.utils.FBUtilities;
 import org.apache.cassandra.utils.Pair;
 
 import static com.google.common.collect.Sets.newHashSet;
+import static org.mockito.Mockito.spy;
+
 /**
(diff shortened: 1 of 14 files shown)


src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -274,6 +294,76 @@ public boolean accept(File file)
...
+            }
+        };
+    }
+
+    public static void verifyDiskHasEnoughUsableSpace()
+    {
+        Map<DataDirectory, Double> dirsToUtilization = Arrays.stream(dataDirectories)
+                                                             .collect(Collectors.toMap(
+                                                             Function.identity(),
+                                                             dir -> {
```
leonz, line 352: make this its own function

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -274,6 +294,76 @@ public boolean accept(File file)
...
+                                                                "Failed to filter most full data directory"));
+        logger.debug(String.format("Highest data directory disk utilization on path %s (%f)",
+                                   maxPathToUtilization.getKey().location, maxPathToUtilization.getValue()));
+        double threshold = DatabaseDescriptor.getMaxDiskUtilizationThreshold();
+        double currUtilization = maxPathToUtilization.getValue();
+        if (currUtilization >= threshold)
+        {
+            ExceededDiskThresholdException cause = new ExceededDiskThresholdException(
+                                    maxPathToUtilization.getKey().location, currUtilization, threshold);
+            throw new FSWriteError(cause, maxPathToUtilization.getKey().location);
```
leonz: Why did we decide to wrap this in an FSWriteError?  We aren't actually attempting to write anything at the time of the failure.

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -274,6 +294,76 @@ public boolean accept(File file)
...
+        }
+
+        if (currUtilization < threshold && StorageService.instance.isNodeDisabled() && StorageService.instance.isSetupCompleted())
```
leonz: don't need `currUtilization < threshold`, since it logically follows from the previous

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -46,6 +46,7 @@
      */
     public enum NonTransientError {
         COMMIT_LOG_CORRUPTION,
+        EXCEEDED_DISK_THRESHOLD,
```
leonz: So this might be slightly annoying, but technically, this _isn't_ a non-transient error because if disk usage goes back down then Cassandra will recover.

Rather, this is the first "transient" error that we're introducing.

The benefit of separating these is that it becomes clear to the operator that we don't need to worry about trying to fix an exceeded disk threshold error (whereas we do need to fix the corruptions).  However, I think we'll need to expose transient errors separately in the API so that we can pull these in the consuming products.

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -103,7 +99,8 @@ public static CassandraDaemon getInstanceForTesting()
         return instance;
     }
 
-    static {
+    static
```
leonz: While I'm not opposed to improving code style (I assume that this is automated via your editor), I'm worried that this is actually going to introduce complexity when it comes to merging upstream changes in the future.

So if possible, we should be touching as little code as possible when editing this repository.

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -274,6 +294,76 @@ public boolean accept(File file)
...
+        {
+            Set<Map<String, String>> nonTransientErrors = StorageService.instance.getNonTransientErrors();
+            boolean isOnlyExceededDisk = true;
+            for (Map<String, String> nonTransientError : nonTransientErrors)
```
leonz: see my comment below, but if we make exceeded disk a transient error, then this block simplifies to `if (nonTransientErrors.isEmpty())`

leonz (changes requested):
Good job on the tests!

See comments - my biggest concern is the transient error thing.

src/java/org/apache/cassandra/db/Directories.java
zpear: Hm, I think my logic at the start was since it will be handled by the `DefaultFSErrorHandler` it should have been wrapped with a subclass of `FSError`, so either `FSWriteError` or `FSReadError`. Looking back at it now though I agree it makes sense for it to remain independent.

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -28,30 +26,51 @@
...
 import org.apache.cassandra.io.util.FileUtils;
-import org.apache.cassandra.io.sstable.*;
+import org.apache.cassandra.service.StorageService;
+import org.apache.cassandra.service.StorageServiceMBean;
 import org.apache.cassandra.utils.ByteBufferUtil;
 import org.apache.cassandra.utils.FBUtilities;
 import org.apache.cassandra.utils.Pair;
 
+import static com.google.common.collect.Sets.newHashSet;
+import static org.mockito.Mockito.spy;
```
leonz: Can we do this in a way where we don't import Mockito into production code?

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -274,6 +292,77 @@ public boolean accept(File file)
...
+    {
+        Map.Entry<DataDirectory, Double> maxPathToUtilization = getMaxPathToUtilization();
+        double threshold = DatabaseDescriptor.getMaxDiskUtilizationThreshold();
+        double currUtilization = maxPathToUtilization.getValue();
+        if (currUtilization >= threshold)
+        {
+            throw new ExceededDiskThresholdException(maxPathToUtilization.getKey().location, currUtilization, threshold);
+        }
+
+        if (StorageService.instance.isNodeDisabled() && StorageService.instance.isSetupCompleted())
```
leonz: proposal: we should maybe add some padding here (like, maybe require that disk drops 2+% below threshold), otherwise we might get into a state where cassandra is constantly turning itself on and off right around the 99% mark.

leonz (commented):
One last suggestion, current code looks good.

src/java/org/apache/cassandra/db/Directories.java
zpear: Do you think this margin should be configurable as well or just hardcoded at 2%?

leonz: Hardcoded is probably fine.  Would do 2%

leonz (approved):

PR #236, opened 2020-12-01
Move log line to log less aggressively 

Sam-Kramer:
Move log line to log less aggressively


Mattheo28:
doesn't this log just as much, once for each timed out command?

Sam-Kramer:
Before we logged for every command, now we only log for timed out commands. I think we want to keep the logs for timed out commands, as it's pretty useful to know.

Mattheo28 (approved):

PR #240, opened 2020-12-16
Fix snapshot race condition

Mattheo28:
context: https://github.palantir.build/foundry/sls-cassandra-sidecar/issues/1098
The race condition seems to occur when these two methods are called at the same time. If a thread calls getSnapshotDetails, then another thread calls clearSnapshot, the original thread will then fail to get the size of the snapshot on disk because it has been cleared by the other thread. Making these synchronized will fix this issue, as the 2nd thread won't be able to execute until the 1st is done.

### DIFF
--- a/src/java/org/apache/cassandra/db/Directories.java
+++ b/src/java/org/apache/cassandra/db/Directories.java
@@ -38,6 +38,7 @@ import java.util.Set;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -799,7 +799,7 @@ public boolean accept(File file)
      * @return  Return a map of all snapshots to space being used
      * The pair for a snapshot has size on disk and true size.
      */
-    public Map<String, Pair<Long, Long>> getSnapshotDetails()
+    public synchronized Map<String, Pair<Long, Long>> getSnapshotDetails()
```
Sam-Kramer: How do you feel about creating a separate object for this, as we've done for compactions?

I.e., private static final Object SNAPSHOT_LOCK = new Object(); 

Then doing 

```
public Map<String, Pair<Long, Long>> getSnapshotDetails() {
   synchronized(SNAPSHOT_LOCK) { ... }
}
```


IIRC, the object that's used for locking for `synchronized` is itself, but if it's static (clearSnapshots), then it defaults to the class. I'd also imagine we want this to lock across threads?

Sam-Kramer: May also looking into if we can do a read-write style lock; calling listSnapshots from multiple threads is fine, it's only when we clear it becomes an issue.

Mattheo28: ah yeah you're right, i'll look into that

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -830,27 +833,35 @@ public boolean accept(File file)
 
     private List<File> listSnapshots()
     {
-        final List<File> snapshots = new LinkedList<>();
-        for (final File dir : dataPaths)
+        SNAPSHOT_LOCK.readLock().lock();
```
Sam-Kramer: Does this have a timeout, or is forever blocking? If not, may make sense to fail after >60 seconds, so we are aware of a deadlock scenario?

Sam-Kramer (approved):

leonz (approved):

PR #241, opened 2020-12-16
Add predicted speculative retry metrics

zpear:
Metrics for what the perceived gain during SND will be from switching to speculative retry will be useful without needing to manually switch and monitor every table. The predicted speculative retry metrics can be separated by threshold and read executor (writing metrics for executors other than `NeverSpeculatingReadExecutor` may be beneficial for fine-tuning thresholds later on). 

Any given read will trigger write predicted metrics for all thresholds faster than that read's latency using the current sampled p99 latency for the endpoint that would have been chosen for speculative retry. Tested on IL and everything looks as I would expect.


src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1573,8 +1573,14 @@ else if (logger.isDebugEnabled())
                     Keyspace.open(command.ksName).getColumnFamilyStore(command.cfName).metric.blockingReadRepairLatency.addNano(latency);
                 }
             }
+            for (AbstractReadExecutor exc : readExecutors) {
```
Sam-Kramer: We should probably do this after each time we receive a row, i.e. in a finally on line 1505 (after we go through and #get each row)

src/java/org/apache/cassandra/locator/AbstractEndpointSnitch.java
```diff
@@ -55,6 +55,11 @@ public int compare(InetAddress a1, InetAddress a2)
         });
     }
 
+    public long getP99Latency(InetAddress endpoint)
```
Sam-Kramer: Can we make this snapshot instead?

src/java/org/apache/cassandra/service/AbstractReadExecutor.java
```diff
@@ -157,6 +181,68 @@ public Row get() throws ReadFailureException, ReadTimeoutException, DigestMismat
...
+            case SECONDS_1:
+                thresholdTime = 1;
+                unit = TimeUnit.SECONDS;
+                break;
+            case MILLISECONDS_100:
+                thresholdTime = 100;
+                unit = TimeUnit.MILLISECONDS;
+                break;
+            case P50:
+                thresholdTime = (long) (cfs.metric.coordinatorReadLatency.getSnapshot().getMedian());
```
Sam-Kramer: we should use the dynamic snitch snapshot here instead!

Sam-Kramer: nevermind ignore me

src/java/org/apache/cassandra/service/AbstractReadExecutor.java
```diff
@@ -157,6 +181,68 @@ public Row get() throws ReadFailureException, ReadTimeoutException, DigestMismat
...
+     * exceeded, the current p99 latency of the retry endpoint is added to the threshold as the "predicted performance"
+     */
+    public void writePredictedSpeculativeRetryPerformanceMetrics(long timestamp) {
+        InetAddress extraReplica = Iterables.getLast(targetReplicas);
+        for (Threshold threshold : Threshold.values()) {
+            writePredSpecRetryPerformanceMetricByThreshold(threshold, timestamp, extraReplica);
+        }
+    }
+
+    private void writePredSpecRetryPerformanceMetricByThreshold(Threshold threshold, long timestamp, InetAddress extraReplica) {
```
Sam-Kramer: Can we throw all of this into a separate class? Will be easier to manage future commits/patches.

src/java/org/apache/cassandra/service/AbstractReadExecutor.java
```diff
@@ -157,6 +181,68 @@ public Row get() throws ReadFailureException, ReadTimeoutException, DigestMismat
...
+        }
+        long extraReplicaP99Latency;
+        try {
+            extraReplicaP99Latency = DatabaseDescriptor.getEndpointSnitch().getP99Latency(extraReplica);
+        } catch (RuntimeException e) {
+            logger.error("Failed to get p99 latency from endpoint snitch to record predicted speculative retry metrics", e);
+            return;
+        }
+        thresholdTime = TimeUnit.NANOSECONDS.convert(thresholdTime, unit);
+        if (timestamp - start > thresholdTime && extraReplicaP99Latency > 0) {
```
Sam-Kramer: Another thing to consider here: if we can get the contact times for each replica, and we they both exceed the threshold, then retry would've been useless here.

src/java/org/apache/cassandra/service/AbstractReadExecutor.java
```diff
@@ -110,16 +116,27 @@ private void makeRequests(ReadCommand readCommand, Iterable<InetAddress> endpoin
...
         }
 
         // We delay the local (potentially blocking) read till the end to avoid stalling remote requests.
         if (hasLocalEndpoint)
         {
+            long localStart = System.nanoTime();
             logger.trace("reading {} locally", readCommand.isDigestQuery() ? "digest" : "data");
             KeyspaceAwareSepQueue.setCurrentKeyspace(command.ksName);
             StageManager.getStage(stage(command)).maybeExecuteImmediately(new LocalReadRunnable(command, handler));
+            synchronized (latencies) {
```
Sam-Kramer: To avoid having to manually synchronize access to this list, let's use https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/ConcurrentLinkedQueue.html

Can just use Collection as the top-level type.

src/java/org/apache/cassandra/service/ReadCallback.java
```diff
@@ -85,22 +88,48 @@
...
 
-    public ReadCallback(IResponseResolver<TMessage, TResolved> resolver, ConsistencyLevel consistencyLevel, int blockfor, IReadCommand command, Keyspace keyspace, List<InetAddress> endpoints)
-    {
+    public ReadCallback(IResponseResolver<TMessage, TResolved> resolver,
+                        ConsistencyLevel consistencyLevel,
+                        int blockfor,
+                        IReadCommand command,
+                        Keyspace keyspace,
+                        List<InetAddress> endpoints) {
+        this(resolver, consistencyLevel, blockfor, command, keyspace, endpoints, null);
```
Sam-Kramer: Can we use Optional instead of null here?

src/java/org/apache/cassandra/locator/AbstractEndpointSnitch.java
```diff
@@ -55,6 +56,11 @@ public int compare(InetAddress a1, InetAddress a2)
         });
     }
 
+    public Snapshot getSnapshot(InetAddress endpoint)
```
Sam-Kramer: What are your thoughts on making this optional as well? I usually don't like controlling program flow via exceptions, but also see that this makes sense in isolation

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -668,7 +670,9 @@ public int sendRRWithFailure(MessageOut message, InetAddress to, IAsyncCallbackW
     public int sendRR(MessageOut message, InetAddress to, IAsyncCallback cb, long timeout, boolean failureCallback)
     {
         int id = addCallback(cb, message, to, timeout, failureCallback);
-        sendOneWay(failureCallback ? message.withParameter(FAILURE_CALLBACK_PARAM, ONE_BYTE) : message, id, to);
+        if (!testOnly) {
```
Sam-Kramer: why is this needed?

src/java/org/apache/cassandra/service/ReadCallback.java
```diff
@@ -171,6 +207,11 @@ public void response(MessageIn<TMessage> message)
         }
     }
 
+    @VisibleForTesting
+    public void responseForTest() {
```
Sam-Kramer: Couldn't we just pass in dummy messages to `#response` to achieve the same result?

src/java/org/apache/cassandra/net/MessagingService.java
zpear: Wasn't able to mock a remote address for testing and saw this pattern already used for a test instance of MessagingService. I just found that I can use 127.0.0.2 to model a remote endpoint so I'll remove all this

src/java/org/apache/cassandra/service/ReadCallback.java
zpear: yep!

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1860,7 +1866,11 @@ private static float calculateResultRowsUsingEstimatedKeys(ColumnFamilyStore cfs
                     // collect replies and resolve according to consistency level
                     RangeSliceResponseResolver resolver = new RangeSliceResponseResolver(nodeCmd.keyspace, command.timestamp);
                     List<InetAddress> minimalEndpoints = filteredEndpoints.subList(0, Math.min(filteredEndpoints.size(), consistency_level.blockFor(keyspace)));
-                    ReadCallback<RangeSliceReply, Iterable<Row>> handler = new ReadCallback<>(resolver, consistency_level, nodeCmd, minimalEndpoints);
+                    ReadCallback<RangeSliceReply, Iterable<Row>> handler = new ReadCallback<>(resolver,
```
Sam-Kramer, line 1869: nit (can ignore): any reason why we don't use the other constructor

Sam-Kramer (approved):
LGTM! One minor nit, not worthy of blocking for merge.

src/java/org/apache/cassandra/service/StorageProxy.java
zpear: The other one uses a couple different parameters (blockFor and keyspace) so I figured passing Optional.empty() here was cleaner than overloading the constructor again

PR #243, opened 2021-01-26
Add enable/disable node jmx endpoints

Sam-Kramer:


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4750,19 +4750,30 @@ public static void enableAutoCompaction() {
...
+    }
+
+    @Override
+    public void disableNode() {
+        setMode(Mode.DISABLED, "Node has been disabled", true);
+        internalDisableNode();
+    }
+
+    @Override
     public void enableNode() {
```
leonz: can you add a check to make sure that the node is already in disabled mode?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4750,19 +4750,30 @@ public static void enableAutoCompaction() {
...
 
+    public void internalDisableNode() {
+        logger.info("Stopping transports and disabling auto compaction");
+        instance.stopTransports();
+        disableAutoCompaction();
+    }
+
+    @Override
+    public void disableNode() {
+        setMode(Mode.DISABLED, "Node has been disabled", true);
```
leonz: i think we should only be allowed to disable a node that is either already disabled or normal.  otherwise we might accidentally wipe important information about the status

src/java/org/apache/cassandra/service/StorageService.java
Sam-Kramer: I'll separate this out into `internalEnableNode` and `enableNode`. `enableNode` is also used by the `Directories` stuff. Alternatively, we could simply perform a check here to ensure that transient errors is not empty.

leonz: I think checking for transient errors isn’t enough.

For example, if the node is joining or leaving or zombie node, if we “disable” then “enable” we will set it back to normal mode.  And I don’t know that it’s better to track the original status because, for example, disabling will interrupt joining.

Sam-Kramer: gotcha makes sense -- my latest push should account for this!

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4750,17 +4750,42 @@ public static void enableAutoCompaction() {
...
+    }
+
+    @Override
+    public void disableNode() {
+        String operationMode = getOperationMode();
+        if (Mode.NORMAL.toString().equals(operationMode)) {
+            setMode(Mode.DISABLED, "Node has been disabled", true);
+            internalDisableNode();
+        } else {
+            logger.warn("Not disabling node as in mode {} instead of NORMAL", operationMode);
```
leonz: nit:

`Not disabling node in mode {}; should be NORMAL`

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4750,17 +4750,42 @@ public static void enableAutoCompaction() {
         }
     }
 
-    public void enableNode() {
+    public void internalDisableNode() {
```
leonz: can this be made package-protected?

leonz (approved):
2 nits, lgtm!

src/java/org/apache/cassandra/service/StorageService.java
Sam-Kramer: Sadly the `Directories` class is in a separate package :(

Sam-Kramer: I wil rename to `unsafeDisableNode`

PR #258, opened 2021-03-31
publish sources/javadoc jars for c-thrift and c-all

tpetracca:
Publishing sources and javadoc are required when publishing jars to maven central. So this is a pre-req of #257


leonz:
Is the artifacts failure a flake or a real falure?

leonz (approved):

PR #260, opened 2021-04-07
Prevent bootstraping if previous bootstrap failure is detected

lyubent:
Prevents node from starting up if the most filled `data_file_directories` directory’s utilization exceeds the `palantir_cassandra.bootstrap_disk_usage_threshold_percentage` system var. If the threshold is exceeded the node will be disabled and placed into `NON_TRANSIENT_ERROR` mode stating that an unrecoverable error was hit. The node will stay up with JMX being available but will not respond to any client requests (as it is disabled).  

The sequence of expected node states would be:
STARTING -> WAITING_TO_BOOTSTRAP -> NON_TRANSIENT_ERROR

Important notes:
- the path at which bootstrapping terminates is before metrics get started, thus metrics will be unavailable
- seed nodes bypass this path

Sample log:

```
INFO  13:57:39 NON_TRANSIENT_ERROR: Detected previous bootstrap failure, data dir too full to proceed.
ERROR 13:57:39 Preventing node from continuing after failed bootstrap as data_file_dirs are too full (7.534545527230836%) and exceed palantir_cassandra.bootstrap_disk_usage_threshold_percentage (0%) 
INFO  13:57:39 Stopping transports and disabling auto compaction
ERROR 13:57:39 Stopping gossiper
WARN  13:57:39 Stopping gossip by operator request
WARN  13:57:39 No local state, state is in silent shutdown, or node hasn't joined, not announcing shutdown
Preventing node from continuing after failed bootstrap as most utilized data_file_dir (<redacted path>) is too full (7.534546) and exceeds palantir_cassandra.bootstrap_disk_usage_threshold_percentage (5.000000).  Node will stay in NON_TRANSIENT_ERROR mode.
Non-fatal bootstrap error. Server will continue but is disabled and without metrics.
```

### DIFF
--- a/src/java/org/apache/cassandra/db/Directories.java
+++ b/src/java/org/apache/cassandra/db/Directories.java
@@ -349,17 +349,7 @@ public class Directories
         return underThreshold && nodeCanBeReEnabled && zeroNonTransientErrors && onlyExceededDiskThresholdTransientError;
     }
 
-    /**
-     * Finds busiest data directory and returns its utilization as a percentage.
-     *
-     * @return Map.Entry<DataDirectory, Double> where the key represents the directory
-     *         and the value represents percentage utilization of the directory.
-     */
-    public static Map.Entry<DataDirectory, Double> getMostUtilizedDataDir() {
-        return getMaxPathToUtilization();
-    }
-
-    private static Map.Entry<DataDirectory, Double> getMaxPathToUtilization()
+    public static Map.Entry<DataDirectory, Double> getMaxPathToUtilization()
     {
         Map<DataDirectory, Double> dirsToUtilization = Arrays.stream(dataDirectories)
                                                              .collect(Collectors.toMap(
--- a/src/java/org/apache/cassandra/db/Directories.java
+++ b/src/java/org/apache/cassandra/db/Directories.java
@@ -349,6 +349,16 @@ public class Directories
         return underThreshold && nodeCanBeReEnabled && zeroNonTransientErrors && onlyExceededDiskThresholdTransientError;
     }
 
+    /**
+     * Finds busiest data directory and returns its utilization as a percentage.
+     *
+     * @return Map.Entry<DataDirectory, Double> where the key represents the directory
+     *         and the value represents percentage utilization of the directory.
+     */
+    public static Map.Entry<DataDirectory, Double> getMostUtilizedDataDir() {
+        return getMaxPathToUtilization();
+    }
+
     private static Map.Entry<DataDirectory, Double> getMaxPathToUtilization()
     {
         Map<DataDirectory, Double> dirsToUtilization = Arrays.stream(dataDirectories)
--- a/src/java/org/apache/cassandra/service/CassandraDaemon.java
+++ b/src/java/org/apache/cassandra/service/CassandraDaemon.java
@@ -59,6 +59,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.Uninterruptibles;
 
 import com.palantir.cassandra.concurrent.LocalReadRunnableTimeoutWatcher;
+import com.palantir.cassandra.db.BootstrappingSafetyException;
 import org.apache.cassandra.config.Config;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -315,6 +316,11 @@ public class CassandraDaemon
         try
         {
             StorageService.instance.initServer();
+        }
+        catch (BootstrappingSafetyException e)
+        {
+            System.out.println(e.getMessage() + "\nNon-fatal bootstrap error. Server will continue but is disabled and without metrics.");
+            return;
         }
         catch (ConfigurationException e)
         {
--- a/src/java/org/apache/cassandra/service/CassandraDaemon.java
+++ b/src/java/org/apache/cassandra/service/CassandraDaemon.java
@@ -319,12 +319,12 @@ public class CassandraDaemon
         }
         catch (BootstrappingSafetyException e)
         {
-            System.out.println(e.getMessage() + "\nNon-fatal bootstrap error. Server will continue but is disabled and without metrics.");
+            logger.error("Non-fatal bootstrap error. Server will continue but is disabled and without metrics.", e);
             return;
         }
         catch (ConfigurationException e)
         {
-            System.err.println(e.getMessage() + "\nFatal configuration error; unable to start server.  See log for stacktrace.");
+            logger.error("Fatal configuration error; unable to start server.  See log for stacktrace.", e);
             exitOrFail(1, "Fatal configuration error", e);
         }
 
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -37,6 +37,7 @@ import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
 import com.google.common.collect.*;
 import com.google.common.util.concurrent.*;
+import com.palantir.cassandra.db.BootstrappingSafetyException;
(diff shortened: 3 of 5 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -868,9 +869,38 @@ private void joinTokenRing(int delay) throws ConfigurationException
...
+                // on detection of previous bootstrap failure prevent boostrap from proceeding if data directories are
+                // filled up beyond palantir_cassandra.bootstrap_disk_usage_threshold_percentage
+                Integer threshold = Integer.getInteger("palantir_cassandra.bootstrap_disk_usage_threshold_percentage");
+                Entry<Directories.DataDirectory, Double> mostUtilizedDataDir = Directories.getMostUtilizedDataDir();
+                Double percentageDataDirUtilization = mostUtilizedDataDir.getValue() * 100;
+
+                if (threshold != null && percentageDataDirUtilization > Double.valueOf(threshold))
+                {
+                    // disable node preventing bootstrap continuation.
+                    setMode(Mode.NON_TRANSIENT_ERROR, "Detected previous bootstrap failure, data dir too full to proceed.", true);
```
leonz: Rather than setting the mode directly, let's call `recordNonTransientError` with appropriate args for this type of failure.  That way, we will be able to expose the non transient error via JMX for our health check infrastructure.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -868,9 +869,38 @@ private void joinTokenRing(int delay) throws ConfigurationException
             setMode(Mode.WAITING_TO_BOOTSTRAP, "Awaiting start bootstrap call", true);
             bootstrapManager.awaitBootstrappable();
             if (SystemKeyspace.bootstrapInProgress())
-                logger.warn("Detected previous bootstrap failure; retrying");
+            {
+                // on detection of previous bootstrap failure prevent boostrap from proceeding if data directories are
+                // filled up beyond palantir_cassandra.bootstrap_disk_usage_threshold_percentage
+                Integer threshold = Integer.getInteger("palantir_cassandra.bootstrap_disk_usage_threshold_percentage");
```
leonz: Nit: not a huge deal in this case, but our style for pulling env variables is to do it as a static field so that it only happens one time, and not in the critical path for the function.

leonz (changes requested):

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -349,6 +349,16 @@ private static boolean shouldEnableNode(double threshold, double currUtilization
...
+     *
+     * @return Map.Entry<DataDirectory, Double> where the key represents the directory
+     *         and the value represents percentage utilization of the directory.
+     */
+    public static Map.Entry<DataDirectory, Double> getMostUtilizedDataDir() {
+        return getMaxPathToUtilization();
+    }
+
     private static Map.Entry<DataDirectory, Double> getMaxPathToUtilization()
     {
```
zpear: nit: Is there a reason not to rename `getMaxPathToUtilization()` to `getMostUtilizedDataDir` and make it public?

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -316,6 +317,11 @@ private void recoverCommitlogAndCompleteSetup() {
         {
             StorageService.instance.initServer();
         }
+        catch (BootstrappingSafetyException e)
+        {
+            System.out.println(e.getMessage() + "\nNon-fatal bootstrap error. Server will continue but is disabled and without metrics.");
```
zpear: Why are we using System.out instead of the `logger` (here as well as in the pre-existing catch-es)? @leonz will we get these in Aries? I don't know myself

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -868,9 +869,38 @@ private void joinTokenRing(int delay) throws ConfigurationException
             setMode(Mode.WAITING_TO_BOOTSTRAP, "Awaiting start bootstrap call", true);
             bootstrapManager.awaitBootstrappable();
             if (SystemKeyspace.bootstrapInProgress())
-                logger.warn("Detected previous bootstrap failure; retrying");
+            {
```
zpear, line 873: Could you write some kind of unit test for this? Even if it's (understandably!) hard to mock things end-to-end given the structure of the codebase you could still pull the new logic into another helper function and use the @VisibleForTesting annotation on it to get at this specifically

src/java/org/apache/cassandra/service/CassandraDaemon.java
leonz: No, we shouldn't use System.out

src/java/org/apache/cassandra/db/Directories.java
lyubent: It will make merging new releases of oss C* harder, but seen as 2.2.x is EOL it makes sense to do this. I'll keep the function name as is (getMaxPathToUtilization) but will directly make it public rather than creating a public version of the function with a different name.

src/java/org/apache/cassandra/service/CassandraDaemon.java
lyubent: I'll refactor this and the chained catches within the function. There are a few points in the o.a.c.service.CassandraDaemon that use System.out instead of the logger but these should be refactored separately.

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -316,9 +317,14 @@ private void recoverCommitlogAndCompleteSetup() {
         {
             StorageService.instance.initServer();
         }
+        catch (BootstrappingSafetyException e)
+        {
+            logger.error(e.getMessage() + "\nNon-fatal bootstrap error. Server will continue but is disabled and without metrics.");
```
zpear: nit: These could probably be replaced with:
```
logger.error("Non-fatal bootstrap error. Server will continue but is disabled and without metrics.", e);
```

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -868,9 +870,40 @@ private void joinTokenRing(int delay) throws ConfigurationException
...
+                                            ImmutableMap.of("directory", mostUtilizedDataDir.getKey().location.toString(),
+                                                            "percentageDataDirUtilization", percentageDataDirUtilization.toString(),
+                                                            "threshold", BOOTSTRAP_DISK_USAGE_THRESHOLD.toString()));
+                    logger.error("Preventing node from continuing after failed bootstrap as data_file_dirs are too full ({}%) " +
+                                 "and exceed palantir_cassandra.bootstrap_disk_usage_threshold_percentage ({}%) ",
+                                 percentageDataDirUtilization,
+                                 BOOTSTRAP_DISK_USAGE_THRESHOLD);
+                    unsafeDisableNode();
+                    // leave node in non-transient error state and prevent it from bootstrapping into the cluster
+                    // throw new IllegalStateException("Preventing node from continuing after failed bootstrap as data_file_dirs are too full.");
```
zpear: nit: left-over commented out exception

zpear (commented):
LGTM other than the 2 nits! I'll leave the final approval for @leonz

src/java/org/apache/cassandra/service/CassandraDaemon.java
lyubent: Good catch, refactored.

src/java/org/apache/cassandra/service/StorageService.java
lyubent: Comment removed, good catch!

leonz (approved):
LGTM

PR #262, opened 2021-04-08
Join ring refactor

Sam-Kramer:
If a node is in zombie mode, allow us to join the ring with user-provided tokens. This should fail if not in zombie mode (for extra safety) or if the tokens had already been set on the node (otherwise corruption will occur).

### DIFF
--- a/src/java/org/apache/cassandra/dht/BootStrapper.java
+++ b/src/java/org/apache/cassandra/dht/BootStrapper.java
@@ -154,9 +154,8 @@ public class BootStrapper extends ProgressEventNotifierSupport
      * otherwise, if num_tokens == 1, pick a token to assume half the load of the most-loaded node.
      * else choose num_tokens tokens at random
      */
-    public static Collection<Token> getBootstrapTokens(final TokenMetadata metadata) throws ConfigurationException
+    public static Collection<Token> getBootstrapTokens(final TokenMetadata metadata, Collection<String> initialTokens) throws ConfigurationException
     {
-        Collection<String> initialTokens = DatabaseDescriptor.getInitialTokens();
         // if user specified tokens, use those
         if (initialTokens.size() > 0)
         {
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1049,14 +1059,20 @@ public void gossipSnitchInfo()
...
         {
             logger.info("Joining ring by operator request");
             try
             {
-                joinTokenRing(0);
+                if(initalTokens.isEmpty()) {
+                    joinTokenRing(0);
+                } else {
+                    initalTokens.stream().forEach(getPartitioner().getTokenFactory()::validate);
+                    Preconditions.checkState(operationMode.equals(Mode.ZOMBIE), "Cannot join ring without being in Zombie mode.");
```
Sam-Kramer, line 1072: It's not possible AFAIK for this to ever occur, but keeping it in here just in-case

Sam-Kramer:
Added `do not merge` until I've tested JOINING a node.

src/java/org/apache/cassandra/dht/BootStrapper.java
```diff
@@ -154,9 +154,8 @@ public void onFailure(Throwable throwable)
      * otherwise, if num_tokens == 1, pick a token to assume half the load of the most-loaded node.
      * else choose num_tokens tokens at random
      */
-    public static Collection<Token> getBootstrapTokens(final TokenMetadata metadata) throws ConfigurationException
+    public static Collection<Token> getBootstrapTokens(final TokenMetadata metadata, Collection<String> initialTokens) throws ConfigurationException
```
leonz, line 157: I think we should still keep one function with the signature `public static Collection<Token> getBootstrapTokens(final TokenMetadata metadata) throws ConfigurationException` that calls `getBootstrapTokens(metadata, DatabaseDescriptor.getInitialTokens())` so that we can still use config if needed.

Sam-Kramer: I _think_ we're ok here, as this method is only used in `joinTokenRing`, and we pass in the config values for initial tokens: https://github.com/palantir/cassandra/blob/a925e76c9aab62bf0f7c641e18881bcc622af234/src/java/org/apache/cassandra/service/StorageService.java#L1039

leonz (approved):
lgtm

Sam-Kramer:
Tested ring movements by joining/leaving, appeared to have work correctly. Also confirmed setting tokens via config still works as expected.

PR #268, opened 2021-04-23
2.2.18 java11 compatability

lyubent:
The patch aims to make Cassandra compatible with running on JDK 11. 
- adds needed JPMS to make internal JVM 11 apis accessible
- updates GC to G1GC
- updates GC logging params to be compatible with Java 11
- Replaces usage of `sun.misc.Usafe#monitorEnter / monitorExit` in `o.a.c.concurrent.utils.Locks` with a ReentrantLock local to `o.a.c.db.partitions.AtomicBTreePartition`.
- Replaces usage of `sun.nio.ch.DirectBuffer` with `java.nio.DirectByteBuffer` and `java.nio.MappedByteBuffer` where possible.
- adds dependency for @Generated annotation avoiding the need to update jmh-core, ohc and asm dependencies. 

Cassandra dtests were run to verify the changes - identical outcome was achieved when running the branch under JDK11 and JDK8. Also identical outcome was achieved compared to palantir-cassandra-2.2.18 (tag 2.2.18-1.54.0) aka before this pr.


lyubent:
There's a problem with running nodetool, and essentially all other JVM tools that rely on java 11. 

Running JVMs will throw an `IllegalAccessException` because the JPMS access flags weren't added to their lauch scripts. 

```
ava.lang.IllegalAccessException: access to public member failed: jdk.internal.ref.Cleaner.clean[Ljava.lang.Object;@29e495ff/invokeVirtual, from org.apache.cassandra.io.util.FileUtils (unnamed module @2f8dad04)
```

src/java/org/apache/cassandra/io/util/FileUtils.java
```diff
@@ -68,19 +71,29 @@
...
+    private static Class clsDirectBuffer;
+    private static MethodHandle mhDirectBufferCleaner;
+    private static MethodHandle mhCleanerClean;
+
     static
     {
         boolean canClean = false;
         try
         {
+            clsDirectBuffer = Class.forName("sun.nio.ch.DirectBuffer");
```
Sam-Kramer, line 83: To clarify: on quip, you mention that direct memory allocation is only possible in JDK8, and not JDK 11. Did we confirm that is the case? If so, I think it may be worth exploring using sun.misc.Unsafe.invokeCleaner(ByteBuffer) below instead then (and purposefully disabling direct memory support for jdk8), as I think we'll see a major perf hit if we can no longer perform direct memory allocations.

src/java/org/apache/cassandra/io/util/FileUtils.java
```diff
@@ -68,19 +71,29 @@
...
+            Method mDirectBufferCleaner = clsDirectBuffer.getMethod("cleaner");
+            mhDirectBufferCleaner = MethodHandles.lookup().unreflect(mDirectBufferCleaner);
+            Method mCleanerClean = mDirectBufferCleaner.getReturnType().getMethod("clean");
+            mhCleanerClean = MethodHandles.lookup().unreflect(mCleanerClean);
+
             ByteBuffer buf = ByteBuffer.allocateDirect(1);
-            ((DirectBuffer) buf).cleaner().clean();
```
Sam-Kramer, line 77: If we keep the `canClean` code, we should make sure we try and cleanup the above `buf`, and then also remove re-throwing the runtime exception below. As otherwise, we will always think we can cleanup memory mapped files, which may not be true (and I think we will fail randomly during execution). But, I could totally be reading this wrong :D

src/java/org/apache/cassandra/io/util/FileUtils.java
lyubent: The doc needs an update. TL;DR; is that cleanup is best effort and depends on JPMS flags giving Cassandra's code access to `sun.nio.ch.DirectBuffer`. Opening up access to the DirectBuffer is done via `JVM_OPTS="$JVM_OPTS --add-exports java.base/sun.nio.ch=ALL-UNNAMED"` so this now does work.

src/java/org/apache/cassandra/io/util/FileUtils.java
lyubent: The tests in linux show that the cleaner does indeed work with offheap-buffers.

src/java/org/apache/cassandra/io/util/FileUtils.java
Sam-Kramer: Awesome. The only other thing then I'll flag (which is continuing on my comment below), is that we should probably ensure we never do direct memory allocation if we cannot cleanup. My understanding is that direct memory allocations will never be cleaned up otherwise, and we will eventually hit out-of-memory.

lyubent: The runtime-exception is propagated to the main CassandraDaemon thread if either reflection on internal apis isn't enable or if the `sun.nio.ch.DirectBuffer` is unavailable for some reason, this would prevent startup before any memory access takes place (at the static initialization stage). This is about as safe as it gets. 

FWIW, I though about capturing the exception and placing the node in disabled mode with a logging of non-transient error but this just doesn't make sense to do. If the node is configured to use direct memory buffers it should just be allowed to crash before it can create / acess new memtables as that would just simply fail. We should never end up in a scenario where the node is configured for direct buffers but isn't allowed to use reflection to verify that `sun.nio.ch.DirectBuffer` is available.

Sam-Kramer: I would also suggest we add `clean(buf)`, to ensure our actual invoke works as well, to be extra safe. Agreed on not making this a non-transient error.  

So:
```
            clsDirectBuffer = Class.forName("sun.nio.ch.DirectBuffer");
            Method mDirectBufferCleaner = clsDirectBuffer.getMethod("cleaner");
            mhDirectBufferCleaner = MethodHandles.lookup().unreflect(mDirectBufferCleaner);
            Method mCleanerClean = mDirectBufferCleaner.getReturnType().getMethod("clean");
            mhCleanerClean = MethodHandles.lookup().unreflect(mCleanerClean);

            ByteBuffer buf = ByteBuffer.allocateDirect(1);
            clean(buf);
            ...
```

lyubent: This is a good idea, pushed a commit to address forcing a call to FileUtils#clean during the initialization of the cleaner.

src/java/org/apache/cassandra/io/util/FileUtils.java
```diff
@@ -68,19 +71,30 @@
...
+            clean(buf);
             canClean = true;
         }
         catch (Throwable t)
         {
+            logger.error("FATAL: Cannot initialize optimized memory deallocator. Some data, both in-memory and on-disk, may live longer due to garbage collection.");
             JVMStabilityInspector.inspectThrowable(t);
-            logger.info("Cannot initialize un-mmaper.  (Are you using a non-Oracle JVM?)  Compacted data files will not be removed promptly.  Consider using an Oracle JVM or using standard disk access mode");
+            throw new RuntimeException(t);
```
leonz, line 97: why do we choose to throw here?

leonz:
This looks good to me generally.

src/java/org/apache/cassandra/io/util/FileUtils.java
lyubent: TL;DR; if we don't throw an exception, C* will start up and might keep hold of memory it might not be using.

If the cleaner's initialization fails it means that cleaning memory wont be done as intended and can lead to Cassandra retaining locks on memory that it isn't using but that it cannot GC. The exception is a fail-fast way of preventing this scenario. 

My justification for crashing here is that it really shouldn't happen if everything is configured correctly. The `sun.nio.ch.DirectBuffer` class should be available in jdk11 and we should be able to check for it assuming deep reflection was allowed using the correct `--add-opens` flags. If we don't have access to the class we're either using the wrong version of java or are missing JPMS flags.

leonz:
LGTM.  I think in terms of releasing - we can rebase this branch off the main branch and then cut an RC directly, rather than merging.

lyubent:
Rebased on [origin](https://github.com/palantir/cassandra.git)/palantir-cassandra-2.2.18

leonz (approved):

PR #269, opened 2021-04-30
Stop current compactions disk threshold

zpear:
Previously if we exceeded our set disk thresholds we would stop scheduling new compactions / executing pending compactions. This is still dangerous as large in-progress compactions were allowed to continue beyond the thresholds. Now, after passing the configured max disk utilization we stop all in-progress compactions to prevent this dangerous behavior.

Throttling compaction throughput and checking compactionstats in a loop yields the following upon exceeding the disk threshold:

With compaction shutdown:

image
The in-progress compaction is stopped when the shutdown is announced. Once disk usage returns underneath the threshold compactions may be scheduled again.

Without compaction shutdown:
image
The in-progress compaction continues until completion after the shutdown is announced.

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
@@ -1671,6 +1671,14 @@ public class CompactionManager implements CompactionManagerMBean
         }
     }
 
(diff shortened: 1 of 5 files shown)


leonz (approved):
Nice!

PR #272, opened 2021-06-15
add name query metrics

Sam-Kramer:


src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2053,31 +2054,28 @@ public ColumnFamily getColumnFamily(QueryFilter filter)
 
             removeDroppedColumns(result);
 
+            DeletionInfo deletionInfo = result.deletionInfo();
```
Sam-Kramer, line 2057: I was worried previously that the DeletionInfo would be recycled. However, upon further investigation, the deletion info should remain the same from when the query filter runs, and being referenced here.

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -65,15 +67,11 @@
...
     private boolean hitTombstoneWarnThreshold = false;
 
-    private DeletionInfo deletionInfo;
-
     // Not serialized, just a ack for range slices to find the number of live column counted, even when we group
     private ColumnCounter columnCounter;
 
     private CountingCellIterator reducedCells;
 
-    private RangeTombstoneCounter rangeTombstoneCounter = new RangeTombstoneCounter();
```
Sam-Kramer, line 75: Unused

PR #273, opened 2021-06-17
Record cell metrics for range scans 

Sam-Kramer:


publishToMavenLocal.sh
```diff
@@ -8,4 +8,6 @@ ant maven-ant-tasks-retrieve-build
 ./gradlew --no-daemon
 ant build build-test
 ./gradlew --no-daemon ant-artifacts
+./gradlew ant-javadoc-jar
+./gradlew ant-sources-jar
```
Sam-Kramer, line 12: these are needed now, ant-artifacts are not publishing them. figure it has to do with j11 change?

leonz (approved):

PR #280, opened 2021-08-06
Add paxos threadpool

Sam-Kramer:
This would be useful for SND, as the paxos threadpool shares the mutation threadpool, and performs both a read and the write. 

In the case that the data drive is slow, this will block all writes from proceeding, even though there may not be any degradation on the commitlog.


Sam-Kramer:
A couple of potential downsides from this:
1) More writes can occur as we no longer share a common threadpool with mutations 
2) Thread contention simply due to there being more threads.

PR #286, opened 2021-09-09
Add cross-VPC handshake

zpear:
Adds most of the boilerplate needed for a new cross-VPC IP mapping handshake. Currently the handshake defaults to disabled. Once a cluster has upgraded, `cross_vpc_internode_communication_enabled` can be set to true, after which handshakes will commence.

This will be followed up with PRs to 
* add the logic to create + update an in-memory IP mapping, (also IP -> hostname) as well as swap out IPs when needed.
* maintain metrics related to the mapping
* Refactor the scheduled handshake to instead trigger on internode connection failures

See https://github.com/palantir/cassandra/pull/291 for current state of logic


src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
```diff
@@ -0,0 +1,114 @@
...
+ * mapping for cross-vpc connections.
+ */
+public class CrossVpcIpMappingHandshaker
+{
+    private static final Logger logger = LoggerFactory.getLogger(CrossVpcIpMappingHandshaker.class);
+    public static final CrossVpcIpMappingHandshaker instance = new CrossVpcIpMappingHandshaker();
+    private static final DebuggableScheduledThreadPoolExecutor executor = new DebuggableScheduledThreadPoolExecutor(
+    "CrossVpcIpMappingTasks");
+    private volatile ScheduledFuture<?> scheduledCVIMTask;
+    public final static int intervalInMillis = 30000;
```
leonz: mini style nit: use _ or an operation for large numbers, i.e., `30_000` or `30 * 1000`

src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private CrossVpcIpMappingHandshaker() {}
+
+    public static void triggerHandshakeFromSelf(Set<InetAddress> targets)
+    {
+        InetAddressHostname selfName = new InetAddressHostname(FBUtilities.getLocalAddress().getHostName());
+        InetAddressIp selfIp = new InetAddressIp(FBUtilities.getBroadcastAddress().getHostAddress());
+        logger.trace("Triggering handshakes from {}/{} to {}", selfName, selfIp, targets);
+        targets.forEach(target -> triggerHandshake(selfName, selfIp, target));
```
leonz: is there any chance that triggerHandshake throws a runtime exception? should we be ensuring that the rest of the handshakes still get triggered?

src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public void start()
+    {
+        if (!DatabaseDescriptor.isCrossVpcIpSwappingEnabled()) {
+            logger.info("Cross VPC IP Swapping is disabled. Not scheduling handshake task.");
+            return;
+        }
+        if (isEnabled())
+        {
+            logger.info("Cross VPC IP Swapping already enabled and scheduled. Ignoring extra start() call.");
```
leonz, line 249: did you mean to return here?  why would we ever hit this block?

src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+        logger.info("Started running CrossVpcIpMappingTask at interval of {} ms",
+                     CrossVpcIpMappingHandshaker.intervalInMillis);
+        scheduledCVIMTask = executor.scheduleWithFixedDelay(() -> {
+            // seeds should be provided via config by hostname with our setup. Could probably get fancier with deciding
+            // who we currently are going to handshake with like the Gossiper does
+            Set<InetAddress> seeds = DatabaseDescriptor.getSeeds()
```
leonz: I think this is fine for now - if the need arises, we can likely supply this as fresh configuration.

src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
```diff
@@ -0,0 +1,114 @@
...
+            logger.info("Cross VPC IP Swapping is disabled. Not scheduling handshake task.");
+            return;
+        }
+        if (isEnabled())
+        {
+            logger.info("Cross VPC IP Swapping already enabled and scheduled. Ignoring extra start() call.");
+        }
+        logger.info("Started running CrossVpcIpMappingTask at interval of {} ms",
+                     CrossVpcIpMappingHandshaker.intervalInMillis);
+        scheduledCVIMTask = executor.scheduleWithFixedDelay(() -> {
```
leonz, line 254: You should _always_ wrap functions submitted to Java executors with a global try/catch.  If this function throws any kind of exception, then the thread will die silently and never re-run.  We would prefer to log the error loudly and try again.

src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
leonz: one other clean possibility is to use Duration instead: `Duration.ofSeconds(30)`

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -401,6 +402,11 @@ public void startNativeTransport()
             throw new IllegalStateException("thrift transport should be set up before it can be started");
         thriftServer.start();
         logger.info("Thrift server running on {}", new InetSocketAddress(DatabaseDescriptor.getRpcAddress(), DatabaseDescriptor.getRpcPort()));
+
+        if (!CrossVpcIpMappingHandshaker.instance.isEnabled())
```
leonz: Why is this the right place for this call? Why not when we start gossip, for instance?

leonz: Also we should double check that this is properly started if we disable/re-enable the appropriate protocols, i.e., when we call `StorageService#startTransports`

leonz (changes requested):

src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
zpear: Good catch -- did mean to return. I don't think we should ever hit it but just in case wanted to log to be defensive

src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
zpear: good catch, will do

src/java/org/apache/cassandra/service/CassandraDaemon.java
zpear: So you suggest adding it here? https://github.com/palantir/cassandra/blob/bdcda1712fd76196f793fe737f9e683826511905/src/java/org/apache/cassandra/service/StorageService.java#L829

leonz: my hunch is that you might need to add it everywhere that Gossiper#start is called

zpear: Added explicitly to Gossiper#start, Gossiper#stop, as well as shadow round

leonz: why stop?

zpear: Conceptually it made sense to me to stop sending these handshakes when we stop gossip + want the node to be isolated. Are you suggesting keeping it on all the time even when gossip is down? Either way I don't feel strongly.

leonz: Ah nvm, I thought you were going to start handshaking on "stop".

leonz (approved):
LGTM! Definitely needs lots of testing in a dev environment before we roll out to production :)

PR #287, opened 2021-09-10
Restrict Max SSTable Size due to STCS in L0 when using LCS

lyubent:
Restricts size of individual SSTables when selecting candidates for L0 STCS compactions. 

- The setting is off by default (`-Dpalantir_cassandra.limit_max_sstable_size_in_l0` - false by default)
- When its turned on, it will default to 100G as max size for an SSTable to be selected into a compaction
- Additional setting allows to tune max size of individual sstables (`-Dpalantir_cassandra.max_sstable_size_in_l0_in_gb` - 100G by default)

Implementation works by checking size of "pairs" (created by createSSTableAndLengthPairs) inside of `getSSTablesForSTCS` which is in-turn called by `getSTCSInL0CompactionCandidate`. The pair represents a sstable reader and the size of the sstable that is about to be compacted. If the size of the sstable exceeds `palantir_cassandra.max_sstable_size_in_l0_in_gb` it will be discarded from the compaction.

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/LeveledManifest.java
+++ b/src/java/org/apache/cassandra/db/compaction/LeveledManifest.java
@@ -19,6 +19,7 @@ package org.apache.cassandra.db.compaction;
 
 import java.io.IOException;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Predicate;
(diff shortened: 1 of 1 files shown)


lyubent:
I cut an RC for the above change - [Tag 1.58.0-rc8](https://github.com/palantir/cassandra/releases/tag/1.58.0-rc8)

leonz (approved):

PR #289, opened 2021-09-13
Restructure change to be pre-emptive - skip sstables that would push us over the limit

lyubent:
Restructures the change to:
- prevent tables being added to the candidate list if they're going to push the target over the `MAX_COMPACTING_SIZE_IN_L0` size. 
- Prevents an NPE due to `ageSortedSSTables(candidates).subList(0, MAX_COMPACTING_L0)` accessing an index that doesn't exist - this can happen due to individual giant SSTables triggering a compaction where there are less than 32 tables (the 32 is dictated by the hardcoded `MAX_COMPACTING_L0`).

The previous implementation relies on sstables being selected via `ageSortedSSTables` which can still cause problems in a scenario where there is a single (or anything less than 33) giant sstable.(s) `ageSortedSSTables` would still fetch the first 32 oldest sstables - if these 32 are over `MAX_COMPACTING_SIZE_IN_L0` then we still end up with a massive compaction. 

There is also some human-readable logging added to simplify debugging large compactions.

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/LeveledManifest.java
+++ b/src/java/org/apache/cassandra/db/compaction/LeveledManifest.java
@@ -30,6 +30,7 @@ import com.google.common.collect.Sets;
 import com.google.common.primitives.Ints;
 
 import org.apache.cassandra.io.sstable.format.SSTableReader;
+import org.apache.cassandra.io.util.FileUtils;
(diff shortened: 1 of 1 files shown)


leonz (approved):
This makes more sense than what I had!

PR #291, opened 2021-09-15
Add IP + Hostname Mapping Logic

zpear:


src/java/org/apache/cassandra/net/OutboundTcpConnection.java
```diff
@@ -507,7 +508,7 @@ private boolean connect()
                 // SSL errors won't be recoverable within timeout period so we'll just abort
                 return false;
             }
-            catch (IOException e)
+            catch (Exception e)
```
leonz, line 511: what other exceptions should we expect here?

src/java/org/apache/cassandra/security/SSLFactory.java
```diff
@@ -80,20 +94,26 @@ public static SSLServerSocket getServerSocket(EncryptionOptions options, InetAdd
...
 
     /** Create a socket and connect */
-    public static SSLSocket getSocket(EncryptionOptions options, InetAddress address, int port, InetAddress localAddress, int localPort) throws IOException
+    public static SSLSocket getSocket(EncryptionOptions options, InetAddress address, int port, InetAddress localAddress, int localPort) throws Exception
     {
         SSLContext ctx = createSSLContext(options, true);
-        SSLSocket socket = (SSLSocket) ctx.getSocketFactory().createSocket(address, port, localAddress, localPort);
-        prepareSocket(socket, options, address);
+        InetAddress mappedAddress = CrossVpcIpMappingHandshaker.instance.maybeSwapAddress(address);
+        SSLSocket socket;
```
leonz: why is socket defined separately from its type declaration? (and in below functions)

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -1010,7 +1010,21 @@ public void run()
...
                     // determine the connection type to decide whether to buffer
                     DataInputStream in = new DataInputStream(socket.getInputStream());
-                    MessagingService.validateMagic(in.readInt());
+                    try
+                    {
+                        MessagingService.validateMagic(in.readInt());
+                    }
+                    catch (Exception e)
+                    {
+                        if (e.getCause() instanceof EOFException)
```
leonz, line 1019: 👍

src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
```diff
@@ -48,14 +54,136 @@
...
+    }
+
+    /**
+     * Depending on which cross-vpc flags are enabled/disabled, will check the cross-vpc mappings and either swap the
+     * given endpoint with an endpoint derived from a public IP, or DNS using a hostname. If no mappings are found will
+     * return the original endpoint.
+     */
+    public InetAddress maybeSwapAddress(InetAddress endpoint)
+    {
+        if (!DatabaseDescriptor.isCrossVpcInternodeCommunicationEnabled()) {
```
leonz, line 103: can we add `!DatabaseDescriptor.isCrossVpcHostnameSwappingEnabled()` up here and remove it below? seems like if its false then we always return "endpoint"

leonz: oh nevermind, i'm reading this wrong.

src/java/org/apache/cassandra/security/SSLFactory.java
zpear: Oversight from previous commit

src/java/org/apache/cassandra/net/OutboundTcpConnection.java
zpear: Replacing IOException w/ Exception is stemming from `SSLFactory.maybeConnectWithTimeout()`. The call to `Future.get()` throws InterruptedException, ExecutionException, and TimeoutException. Then the call to `Callable.call` throws `Exception`. This is what we discussed offline in wrapping w/ a "mega exception" rather than propagating each of these up the stack.

leonz (approved):

PR #292, opened 2021-09-20
Count amount of copying done on range tombstone lists

j-baker:
Range tombstone lists are a weird structure which seemingly are designed
to not regularly be copied; unlike the normal memtable structures
(stored as a b-tree) the range tombstone list is stored as four arrays.

When the partition is modified, the entire array is copied - to me, the
author assumed that range tombstones were rare. For us, range tombstones
are as common as normal cells.

Put simply, I believe that we could be suffering from (with dynamic
columns tables) some unnecessary O(n^2) behaviour which will make
performance worse and worse unnecessarily.

If this is proven, we have the tactic of modifying the
RangeTombstoneList to actually use the same b-tree logic (which is
likely difficult, but possible). If unproven, we can avoid this.

### DIFF
--- a/src/java/org/apache/cassandra/db/RangeTombstoneList.java
+++ b/src/java/org/apache/cassandra/db/RangeTombstoneList.java
@@ -30,12 +30,17 @@ import com.google.common.collect.Iterators;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


lyubent (approved):
LGTM.

leonz (approved):

PR #299, opened 2021-11-15
Rebuild by keyspace

zpear:
Ported from:

https://issues.apache.org/jira/browse/CASSANDRA-10406

https://github.com/yukim/cassandra/commits/10406-trunk

https://github.com/apache/cassandra-dtest/blob/trunk/rebuild_test.py#L245

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -1221,10 +1221,10 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
 
     public void rebuild(String sourceDc)
     {
-        rebuild(sourceDc, null, null);
+        rebuild(sourceDc, null);
     }
 
-    public void rebuild(String sourceDc, String keyspace, String tokens)
+    public void rebuild(String sourceDc, String keyspace)
     {
         // check on going rebuild
         if (!isRebuilding.compareAndSet(false, true))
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1240,8 +1255,37 @@ public void rebuild(String sourceDc)
...
+            if (keyspace == null)
+            {
+                for (String keyspaceName : Schema.instance.getNonSystemKeyspaces())
+                    streamer.addRanges(keyspaceName, getLocalRanges(keyspaceName));
+            }
+            else if (tokens == null)
+            {
+                streamer.addRanges(keyspace, getLocalRanges(keyspace));
+            }
+            else
```
leonz, line 1255: I suggest we remove the ability to rebuild by token range - this is the most complicated part of this PR and we do not plan to use it.

zpear: removed!

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -27,6 +27,8 @@
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.regex.MatchResult;
```
leonz: nit: remove

leonz (approved):

PR #300, opened 2021-11-16
Add endpoint to enable/disable keyspace creation

Sam-Kramer:


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -852,6 +852,11 @@ public static void createAllDirectories()
                 throw new ConfigurationException("saved_caches_directory must be specified", false);
 
             FileUtils.createDirectory(conf.saved_caches_directory);
+
+            if (conf.lock_keyspaces_directory == null)
```
leonz: should we make this more generic? do we anticipate need to store any other files here?

Sam-Kramer: yeah I was also thinking this, probably makes sense to make this a bit more generic, just in-case we need to use this for something else in the future. maybe we call it 'persistent_settings_directory'?

leonz: 👍

leonz (approved):

PR #301, opened 2021-11-17
Add tracking for keyspaces that have completed streaming ranges

zpear:
Adds a JMX endpoint and nodetool command to see the set of keyspaces that do not have any token ranges that need to be streamed to the given node. This can be used to check the status of a rebuild programmatically rather than looking through logs.

```
Datacenter: DC11
================
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address      Load       Tokens       Owns    Host ID                               Rack
UN  10.0.20.213  11.4 GB    512          ?       c58ec2b2-1729-4944-af37-968c03698090  RAC1
UN  10.0.23.42   11.4 GB    512          ?       12e52712-6072-47e2-bcb2-db954dc4b10e  RAC1
UN  10.0.23.45   11.38 GB   512          ?       88c2f28d-a24d-448e-a7dd-d94e45df3826  RAC1
Datacenter: DC12
================
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address      Load       Tokens       Owns    Host ID                               Rack
UN  10.0.20.157  9.88 MB    512          ?       d8cc048c-193d-4408-9164-f31c8189ba7a  RAC1


$ ./service/bin/nodetool getkeyspaceswithallrangesavailable | grep foundry_email
$ ./service/bin/nodetool rebuild -ks foundry_email -- DC11
$ ./service/bin/nodetool getkeyspaceswithallrangesavailable | grep foundry_email
[...foundry_email...]
```

### DIFF
--- a/src/java/org/apache/cassandra/dht/RangeStreamer.java
+++ b/src/java/org/apache/cassandra/dht/RangeStreamer.java
@@ -358,12 +358,7 @@ public class RangeStreamer
             InetAddress preferred = SystemKeyspace.getPreferredIP(source);
             Collection<Range<Token>> ranges = entry.getValue().getValue();
 
-            // filter out already streamed ranges
-            Set<Range<Token>> availableRanges = stateStore.getAvailableRanges(keyspace, StorageService.getPartitioner());
-            if (ranges.removeAll(availableRanges))
-            {
-                logger.info("Some ranges of {} are already available. Skipping streaming those ranges.", availableRanges);
-            }
+            removeAvailableRanges(entry, true);
 
             if (logger.isTraceEnabled())
                 logger.trace("{}ing from {} ranges {}", description, source, StringUtils.join(ranges, ", "));
@@ -372,5 +367,31 @@ public class RangeStreamer
         }
 
         return streamPlan.execute();
+    }
+
+    public boolean areAllRangesPresent()
+    {
+        boolean allPresent = true;
+        for (Map.Entry<String, Map.Entry<InetAddress, Collection<Range<Token>>>> entry : toFetch.entries())
+        {
+            removeAvailableRanges(entry, false);
+            Collection<Range<Token>> ranges = entry.getValue().getValue();
+            allPresent &= ranges.isEmpty();
+        }
+        return allPresent;
+    }
+
+    @VisibleForTesting
+    void removeAvailableRanges(Map.Entry<String, Map.Entry<InetAddress, Collection<Range<Token>>>> entry, boolean log)
+    {
+        String keyspace = entry.getKey();
+        Collection<Range<Token>> ranges = entry.getValue().getValue();
+
+        // filter out already streamed ranges
+        Set<Range<Token>> availableRanges = stateStore.getAvailableRanges(keyspace, StorageService.getPartitioner());
+        if (ranges.removeAll(availableRanges) && log)
+        {
+            logger.info("Some ranges of {} are already available. Skipping streaming those ranges.", availableRanges);
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1278,6 +1257,67 @@ public void rebuild(String sourceDc, String keyspace)
         }
     }
 
+    public Set<String> getKeyspacesWithAllRangesAvailable(String sourceDc)
```
zpear, line 1260: This is a bit verbose so I'm open to suggestions. Figured I'd start out with the most literal name first though

src/java/org/apache/cassandra/dht/RangeStreamer.java
```diff
@@ -358,12 +358,7 @@ public StreamResultFuture fetchAsync()
             InetAddress preferred = SystemKeyspace.getPreferredIP(source);
             Collection<Range<Token>> ranges = entry.getValue().getValue();
 
-            // filter out already streamed ranges
-            Set<Range<Token>> availableRanges = stateStore.getAvailableRanges(keyspace, StorageService.getPartitioner());
-            if (ranges.removeAll(availableRanges))
-            {
-                logger.info("Some ranges of {} are already available. Skipping streaming those ranges.", availableRanges);
-            }
+            removeAvailableRanges(entry, true);
```
leonz: Doesn't this mean that we aren't actually removing these ranges from `ranges` anymore?

src/java/org/apache/cassandra/dht/RangeStreamer.java
```diff
@@ -373,4 +368,30 @@ public StreamResultFuture fetchAsync()
...
+            allPresent &= ranges.isEmpty();
+        }
+        return allPresent;
+    }
+
+    @VisibleForTesting
+    void removeAvailableRanges(Map.Entry<String, Map.Entry<InetAddress, Collection<Range<Token>>>> entry, boolean log)
+    {
+        String keyspace = entry.getKey();
+        Collection<Range<Token>> ranges = entry.getValue().getValue();
```
leonz: From reading this I'm not 100% sure if this is a reference or a copy.  I would prefer that we be more explicit and return the final set of ranges from this function.

src/java/org/apache/cassandra/service/StorageService.java
leonz: Probably fine.

src/java/org/apache/cassandra/dht/RangeStreamer.java
zpear: It's a reference since one of the unit tests would fail if just a copy. I'll refactor this to return the set!

zpear: pushed

src/java/org/apache/cassandra/dht/RangeStreamer.java
```diff
@@ -373,4 +369,34 @@ public StreamResultFuture fetchAsync()
...
+    {
+        String keyspace = entry.getKey();
+        Collection<Range<Token>> ranges = ImmutableSet.copyOf(entry.getValue().getValue());
+
+        // filter out already streamed ranges
+        Set<Range<Token>> availableRanges = stateStore.getAvailableRanges(keyspace, StorageService.getPartitioner());
+        Set<Range<Token>> unavailableRanges = ranges.stream()
+                                                    .filter(range -> !availableRanges.contains(range))
+                                                    .collect(Collectors.toSet());
+        entry.getValue().getValue().removeAll(availableRanges);
```
leonz, line 395: Mostly a nit, but I think it's slightly more idiomatic to just do:

HashSet<> modifiableSet = new HashSet<>(entry.getValue().getValue());
modifiableSet.removeAll(availableRanges);
return ImmutableSet.copyOf(modifiableSet);

leonz: or with streams:

entry.getValue().getValue().stream()
.filter(x -> !availableRanges.contains(x))
.toSet();

leonz (approved):

PR #307, opened 2021-11-25
stream rebuild for range from paired rack

Sam-Kramer:
Currently, rebuilds are not strictly consistent, as the "closest" rack in the neighboring datacenter is used for streaming, which is more/less arbitrary. 

To resolve, we deterministically map racks across datacenters if their topologies match, and then stream only from replicas from that datacenter/rack. 

To deterministically map, we simply lexicographically sort the racks within each datacenter, and the map them directly by their ordinal index. 

I.e.:
DC1 [A,B,C] - DC2 [F,E,D]

Would map to:
DC1 A -> DC2 D
DC1 B -> DC2 E
DC1 C -> DC2 F

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -49,6 +49,7 @@ import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.Appender;
 import com.palantir.cassandra.cvim.CrossVpcIpMappingAckVerbHandler;
 import com.palantir.cassandra.cvim.CrossVpcIpMappingSynVerbHandler;
+import com.palantir.cassandra.dht.SingleRackFilter;
 import com.palantir.cassandra.utils.LockKeyspaceUtils;
 import org.apache.cassandra.auth.AuthKeyspace;
 import org.apache.cassandra.auth.AuthMigrationListener;
@@ -1310,7 +1311,17 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                                                    streamStateStore);
         streamer.addSourceFilter(new RangeStreamer.FailureDetectorSourceFilter(FailureDetector.instance));
         if (sourceDc != null)
+        {
             streamer.addSourceFilter(new RangeStreamer.SingleDatacenterFilter(DatabaseDescriptor.getEndpointSnitch(), sourceDc));
+            HashMultimap<String, String> topology = HashMultimap.create();
+            Gossiper.instance.getEndpointStates().stream()
+                             .map(Entry::getKey)
+                             .forEach(address -> topology.put(DatabaseDescriptor.getEndpointSnitch().getDatacenter(address),
+                                                              DatabaseDescriptor.getEndpointSnitch().getRack(address)));
+            String localDc = DatabaseDescriptor.getEndpointSnitch().getDatacenter(FBUtilities.getBroadcastAddress());
+            String localRack = DatabaseDescriptor.getEndpointSnitch().getRack(FBUtilities.getBroadcastAddress());
+            streamer.addSourceFilter(SingleRackFilter.create(topology, sourceDc, localDc, localRack));
+        }
 
         if (keyspace == null)
         {
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -1313,6 +1313,12 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
         if (sourceDc != null)
         {
             streamer.addSourceFilter(new RangeStreamer.SingleDatacenterFilter(DatabaseDescriptor.getEndpointSnitch(), sourceDc));
+            /**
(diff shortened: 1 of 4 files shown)


Sam-Kramer:
"Proof of correctness" for historical data:

Simply put, as we're mirroring our topology across datacenters, the only way for us to be _incorrect_, is for our source datacenter to be incorrect. 

### Details: 
RF = 3
Nodes A, B, C that reside in DC1.
Nodes D, E, F that reside in DC2, which we will be 'rebuilding' to. 

### Requirements: 

Read (quorum) at DC2 = Read (quorum) at DC1. 

Cases: (Format: Node: [K, V]) 

- Scenario 1: DC1 A: [1, 5], B: [1, 5], C: [1, 5]
- Scenario 2: DC1 A: [2, 5], B: [2, 5], C: [2, null]
- Scenario 3: DC1 A: [2, 5], B: [2, 1], C: [2, 5]
- Scenario 4: DC1 A: [1, null], B: [1, 5], C: [1, 5]

### Proofs:
Following scenarios rebuild as: DC1 rebuilds to DC2, nodes stream data from: [A -> D], [B -> E], [C -> F] (proposed case). 
- Scenario 1
  - Read(1, D/E, DC2): 5 = Read(1, A/B, DC1) : 5
- Scenario 2
  - Read(2, E/F, DC2): 5 = Read(2, B/C, DC1): 5
- Scenario 3
  - Read(2, E/F, DC2): 5 = Read(2, B/C, DC1): 5
- Scenario 4
  - Read(1, D/F, DC2): 5 = Read(1, A/C, DC1): 5

Following scenarios rebuild as: DC1 rebuilds to DC2, nodes stream data from: [A -> D], [A -> E], [A -> F] (present case). 
- Scenario 1
  - Read(1, D/E, DC2): 5 = Read(1, A/B, DC1) : 5
- Scenario 4
  - Read(1, E/F, DC2): null != Read(1, B/C, DC1): 5 **(failed case)** 

Following scenarios rebuild as: DC1 rebuilds to DC2, nodes stream data from: [A -> D], [A -> E], [B -> F] (present case). 
- Scenario 1
  - Read(1, D/E, DC2): 5 = Read(1, A/B, DC1) : 5
- Scenario 4
  - Read(1, E/F, DC2): null != Read(1, B/C, DC1): 5 **(failed case)** 
  
Edge cases to consider:
- count(racks of dc1) = 1, this will fail. This is due to the fact we stream based on rack, rather than replica. As a result, we will potentially stream from the same replica twice, which violates our quorum requirements, as the replica could be missing or have an old value, thus serving an incorrect read. 
- racks change during rebuilding. this will affect how we lexicographically sort racks potentially, and could result in the same replica being streamed twice. 

Additional assumptions made:
- We start receiving writes at _quorum_ in the new datacenter _before_ we rebuild. This is important, as we could miss a future value say for key '3', which does not exist at the time of the first two rebuilds (D, E), but does after the third (F). As a read(3, D,E) would result in NULL.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1326,7 +1326,8 @@ private RangeStreamer getRebuildStreamer(String sourceDc, String keyspace)
                                                               DatabaseDescriptor.getEndpointSnitch().getRack(address)));
             String localDc = DatabaseDescriptor.getEndpointSnitch().getDatacenter(FBUtilities.getBroadcastAddress());
             String localRack = DatabaseDescriptor.getEndpointSnitch().getRack(FBUtilities.getBroadcastAddress());
-            streamer.addSourceFilter(SingleRackFilter.create(topology, sourceDc, localDc, localRack));
+            AbstractReplicationStrategy replicationStrategy = Keyspace.open(keyspace).getReplicationStrategy();
```
zpear: what happens if a rebuild is triggered without specifying any keyspace? will this break here

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1313,21 +1313,23 @@ private RangeStreamer getRebuildStreamer(String sourceDc, String keyspace)
...
+                HashMultimap<String, String> topology = HashMultimap.create();
+                Gossiper.instance.getEndpointStates().stream()
+                                 .map(Entry::getKey)
+                                 .forEach(address -> topology.put(DatabaseDescriptor.getEndpointSnitch().getDatacenter(address),
+                                                                  DatabaseDescriptor.getEndpointSnitch().getRack(address)));
+                String localDc = DatabaseDescriptor.getEndpointSnitch().getDatacenter(FBUtilities.getBroadcastAddress());
+                String localRack = DatabaseDescriptor.getEndpointSnitch().getRack(FBUtilities.getBroadcastAddress());
+                AbstractReplicationStrategy replicationStrategy = Keyspace.open(keyspace).getReplicationStrategy();
+                streamer.addSourceFilter(SingleRackFilter.create(topology, sourceDc, localDc, localRack, replicationStrategy));
+            }
```
zpear, line 1332: can we add a log saying whether we've added the filter or not and expect the rebuild to be consistent?

zpear: nvm

zpear (approved):

PR #311, opened 2022-01-04
Persistent setting to prevent client interfaces on startup

leonz:
This PR adds:
- A refactor to rely less on static functions and reuse across multiple classes.
- a persistent setting which will prevent client interfaces to startup if true.

For functionality around actually enabling and disabling:
- Disabling the native transports can happen on startup (via a system flag), or via a JMX call.
- Enabling the native transports can only happen via a JMX call


src/java/com/palantir/cassandra/settings/LockKeyspaceCreationSetting.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.palantir.cassandra.settings;
+
+public class LockKeyspaceCreationSetting extends PersistentSetting
+{
+    private static final String SETTING_NAME = "lock_keyspace_creation";
+
+    public static final LockKeyspaceCreationSetting instance = new LockKeyspaceCreationSetting();
+
+    public LockKeyspaceCreationSetting() {
```
Sam-Kramer: This looks great, the only nit I have is that maybe we make these constructors private, to force the usage of the singleton?

leonz: done!

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4933,17 +4934,23 @@ public void enableNode() {
...
     @Override
     public boolean isKeyspaceCreationEnabled() {
-        return !LockKeyspaceUtils.isKeyspaceCreationLocked();
+        return !LockKeyspaceCreationSetting.instance.isTrue();
+    }
+
+    @Override
+    public void persistentEnableClientInterfaces() {
+        DisableClientInterfaceSetting.instance.setFalse();
+        instance.startTransports();
```
zpear, line 4953: If we can only persist the disabled setting on startup, how will we disable one of the DCs during the migration in case of:
* forward: turning off the pcloud interfaces so all requests solely hit rubix
* rollback: turning off the rubix interfaces so all requests solely hit pcloud

Currently we hit the sidecar's `disableClientInterfaces` endpoint which simply calls `stopRpcServer()` and `stopNativeTransport`

leonz: ah good point. okay!

leonz: adding a call to disable via endpoint

zpear (approved):
LGTM! (resolves #310)

Sam-Kramer (approved):

PR #314, opened 2022-01-12
Upgrade Cassandra to Java 11

Sam-Kramer:
Upgrade Cassandra to Java 11


src/java/org/apache/cassandra/security/SSLFactory.java
```diff
@@ -215,9 +205,18 @@ static SSLSocket maybeConnectWithTimeout(Callable<SSLSocket> callable, long time
...
-    private static void maybeAddSni(InetAddress addr, SSLParameters sslParameters)
+    private static void maybeAddSni(InetAddress addr, SSLParameters sslParameters, EncryptionOptions options)
     {
-        String providedName = addr.getHostName();
+        Set<String> validServerNames = getValidServerNames(options);
+        String hostname = addr.getHostName();
+
+        if(!validServerNames.isEmpty() && !validServerNames.contains(hostname)) {
+            hostname = addr.getCanonicalHostName();
+            logger.info("Using canoical hostname {} instead of localhost {} as no entry for hostname found in keystore {}", addr.getHostName(), hostname, validServerNames);
```
Sam-Kramer: typo

src/java/org/apache/cassandra/utils/FBUtilities.java
```diff
@@ -133,7 +133,7 @@ public static InetAddress getLocalAddress()
             try
             {
                 localInetAddress = DatabaseDescriptor.getListenAddress() == null
-                                    ? InetAddress.getLocalHost()
+                                    ? InetAddress.getByName(InetAddress.getLocalHost().getCanonicalHostName())
```
leonz, line 136: This generally seems reasonable.  How does this resolve in Kubernetes?

Sam-Kramer: That's a good flag. I tested this myself, and we should be ok here.

```
ptdocker@cassandra-0:/mnt/secrets/certs$ hostname -f 
cassandra-0.cassandra.<ns>.svc.cluster.local
```
Interesting, `hostname` returns just `cassandra-0`, but the certs are much more resilient in k8s, where they do all permutations of the FQDN.

Sam-Kramer: Actually that's a bit of a lie -- in fact, this fix is also required for K8s, as there is no entry for just the hostname!

leonz (approved):

PR #317, opened 2022-02-02
Cross-VPC handshake all nodes

zpear:
Only handshaking seeds is problematic when we have more than 3 nodes in a DC, so we want to separately provide all hostnames and handshake each one of them.


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1311,6 +1312,32 @@ public static String getPersistentSettingsLocation() {
         return ImmutableSet.<InetAddress>builder().addAll(seedProvider.getSeeds()).build();
     }
 
+    public static Set<InetAddress> getAllHosts()
```
leonz, line 1315: nit: hosts -> peers

leonz (approved):

PR #318, opened 2022-02-04
Handle SSL NTE

Sam-Kramer:
Currently, if SSL handshake failures occur, we operate as "normal", despite the fact that the node is effectively down, as the node may not be able to participate in requests, thus violating quorum requirements.

### DIFF
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -264,6 +264,8 @@ public class Config
 
     public boolean inter_dc_tcp_nodelay = true;
 
+    public boolean enable_ssl_nte = true;
+
     public MemtableAllocationType memtable_allocation_type = MemtableAllocationType.heap_buffers;
 
     /**
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -264,8 +264,6 @@ public class Config
 
     public boolean inter_dc_tcp_nodelay = true;
 
-    public boolean enable_ssl_nte = true;
-
     public MemtableAllocationType memtable_allocation_type = MemtableAllocationType.heap_buffers;
 
     /**
--- a/src/java/org/apache/cassandra/net/OutboundTcpConnection.java
+++ b/src/java/org/apache/cassandra/net/OutboundTcpConnection.java
@@ -49,6 +49,8 @@ import net.jpountz.xxhash.XXHashFactory;
 import org.apache.cassandra.io.util.DataOutputStreamPlus;
 import org.apache.cassandra.io.util.BufferedDataOutputStreamPlus;
 import org.apache.cassandra.io.util.WrappedDataOutputStreamPlus;
+import org.apache.cassandra.service.StorageService;
+import org.apache.cassandra.service.StorageServiceMBean;
 import org.apache.cassandra.tracing.TraceState;
 import org.apache.cassandra.tracing.Tracing;
 import org.apache.cassandra.utils.CoalescingStrategies;
@@ -62,6 +64,7 @@ import org.xerial.snappy.SnappyOutputStream;
 import org.apache.cassandra.config.Config;
 import org.apache.cassandra.config.DatabaseDescriptor;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.util.concurrent.Uninterruptibles;
 
 public class OutboundTcpConnection extends Thread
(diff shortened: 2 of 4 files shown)


src/java/org/apache/cassandra/net/OutboundTcpConnection.java
```diff
@@ -505,6 +508,10 @@ private boolean connect()
             {
                 logger.error("SSL handshake error for outbound connection to " + socket, e);
                 socket = null;
+                if (DatabaseDescriptor.isSslNteEnabled()) {
```
Sam-Kramer: I wonder if this will throw even for Connection timeouts/etc? Probably should test a rolling restart with this.

lyubent: We should try to force timeout by issuing a range scan or making the query timeout window unreasonably small to make sure we treat timeouts as intended.

Sam-Kramer: Yup confirmed on IL that this is ok, and I investigated what could cause SSLHandshakeErrors, and this should be safe.

leonz: Want to confirm - this will only catch SSL exceptions that originate on _this_ node right? Is there any chance that these exceptions might cause other nodes/the whole cluster to shut down too?

Sam-Kramer: Shit, yeah you're right... this could throw if another server's certs are bad, and would cause a cluster wide outage :(

Sam-Kramer: I'll need to rethink this approach

leonz: maybe this is a use case for a _transient error_ where we don't kill the server on failure? our only goal is to surface and prevent the cluster from rolling (change the to "not ready")

src/java/org/apache/cassandra/net/OutboundTcpConnection.java
```diff
@@ -510,7 +510,6 @@ private boolean connect()
                 socket = null;
                 if (DatabaseDescriptor.isSslNteEnabled()) {
                     StorageService.instance.recordNonTransientError(StorageServiceMBean.NonTransientError.SSL_ERROR, ImmutableMap.of());
```
leonz, line 513: recordTransientError?

Sam-Kramer: lol so I'm torn on this, just b/c an ssl error _could_ be transient, or non-transient. I'd argue it's more so the latter, as the only `transient` form of this is a certificate issued too early (which will keep occurring, until the issuer manually fixes this).

leonz: this is mostly up to us to define, cause we don't actually have transient anywhere else defined.  i think "could be resolved with a bounce" probably means it's transient, because the contrapositive "non-transient" explicitly means _requires manual action to resolve_.  agree that sometimes a bounce won't help though, but then the question is, is there an automated action we could take to resolve?

ideally we would have the node that is the _source_ be non transient and the rest of the nodes are transient but that might be hard

Sam-Kramer: I'd push for keeping this non-transient then, as I for-see the majority of our SSL issues requiring manual action, as it indicates that certificates are wrong in some sort of sense.

 I.e., we're issuing certificates too early --> need to manually fix what's generating the certs.
Certs are missing hostname entries --> won't be fixed until a code fix or change how certificates are generated. 
Certs are expired --> need to re-generate new certs. 

I don't think we can realistically differentiate source vs remote in terms the source of issue, as we could be reaching out to ourselves, which is how one of the issues we hits manifests.

leonz: sounds good then! we should make sure that all of our upstream code doesn't barf on a new non transient error type

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -264,6 +264,8 @@
 
     public boolean inter_dc_tcp_nodelay = true;
 
+    public boolean enable_ssl_nte = true;
```
leonz: prefer making changes in cassandra_env.sh rather than Config.java.

Reason being that Config.java changes are breaking changes that make rolling versions harder, whereas cassandra_env.sh will just no-op if the variable doesn't exist.

src/java/org/apache/cassandra/net/OutboundTcpConnection.java
Sam-Kramer: yeah we fail hard on a new non-transient error type (if thrown), so we will have to handle that :( I'm thinking we will cut a release adding this transient type, updating said dependencies, then releasing a new version of C*

Sam-Kramer: although this may not be as bad as it seems, as we also do not handle `bootstrap_error`. https://github.com/palantir/cassandra/pull/319/files

leonz (approved):

PR #323, opened 2022-03-07
Add cleanup persistent state per node

inespot:
Adds cleanup state to Cassandra node.


src/java/org/apache/cassandra/service/CleanupStateTracker.java
```diff
@@ -0,0 +1,171 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+
+public class CleanupStateTracker
+{
+    private static final Logger log = LoggerFactory.getLogger(CleanupStateTracker.class);
+
+    @VisibleForTesting
+    static final String CLEANUP_STATE_FILE_NAME = "node_cleanup_state.json";
+    private static final Instant MIN_TS = Instant.parse("2022-01-01T00:00:00Z");
```
leonz: this probably works, but the standard is to use `Instant.EPOCH` as a "minimum" value

src/java/org/apache/cassandra/service/CleanupStateTracker.java
```diff
@@ -0,0 +1,171 @@
...
+        this.cleanupStateCache = new CleanupState(createOrRetrieveCleanupStateFile());
+    }
+
+    @VisibleForTesting
+    File createOrRetrieveCleanupStateFile() throws IOException
+    {
+        File cleanupStateFile = new File(cleanupStateFileLocation());
+        try
+        {
+            cleanupStateFile.createNewFile();
```
leonz: do we want to call this if it already exists?

src/java/org/apache/cassandra/service/CleanupStateTracker.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @VisibleForTesting
+    static class CleanupState
+    {
+        private final ConcurrentMap<String, Long> tableEntries;
+        private final File persistentCleanupStateFile;
+        private final ObjectMapper objectMapper = new ObjectMapper();
+
+        public CleanupState(File cleanupStateFile) throws IOException
```
leonz: My recommendation is that you do any file-based operations (file creation, mapping, persistence) in one class.  Right now you have some operations in this inner class, and others in the class above.

A more logical separation would be for `CleanupState` to be only responsible for holding the in-memory representation of the data (i.e., the tableEntries), and then maybe a `CleanupStatePersister` which is responsible for writing or reading from disk.

src/java/org/apache/cassandra/service/CleanupStateTracker.java
inespot: According to the [doc](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#createNewFile--), this function only creates the file if it does not already exist

src/java/org/apache/cassandra/service/CleanupStateTracker.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+    @VisibleForTesting
+    String cleanupStateFileLocation(){
+        return DatabaseDescriptor.getPersistentSettingsLocation() + CLEANUP_STATE_FILE_NAME;
+    }
+
+    /** Creates table entry if it does not already exist */
+    public synchronized void createCleanupEntryForTableIfNotExists(String keyspaceName, String columnFamily)
```
leonz: I think it's a bug for this and the next function to need to be synchronized.  Ideally you can contain the need for synchronization inside of the CleanupState class.

src/java/org/apache/cassandra/service/CleanupStateTracker.java
inespot: That makes sense, I'll make those changes

leonz:
So high level, maybe the way you want to organize this is:

```
public CleanupStateTracker () { // the class that a user would interact with
   this.persister = new CleanupStatePersister(); // responsible for reading/writing to disk
   this.state = new CleanupState(); // responsible for managing current state
}

// the minimal api that a user would need to interact with this
```

src/java/org/apache/cassandra/service/cleanupstate/CleanupStateTracker.java
```diff
@@ -0,0 +1,93 @@
...
+    static final Instant MIN_TS = Instant.EPOCH;
+    private static final String CLEANUP_STATE_FILE_NAME = "node_cleanup_state.json";
+
+    private final CleanupState state;
+    private final CleanupStatePersister persister;
+
+    public CleanupStateTracker() throws IOException
+    {
+        this.persister = new CleanupStatePersister(cleanupStateFileLocation());
+        this.state = new CleanupState(persister.readCleanupStateFromFile());
```
zpear: What do you think about making CleanupStatePersister a property of CleanupState? So when you update CleanupState the file is updated in the same call

src/java/org/apache/cassandra/service/cleanupstate/CleanupStatePersister.java
```diff
@@ -0,0 +1,88 @@
...
+            this.persistentFile = cleanupStateFile;
+        }
+        catch (IOException e)
+        {
+            log.error("Cannot retrieve or create node cleanup state file.", cleanupStateFile.getAbsolutePath(), e);
+            throw e;
+        }
+    }
+
+    public Map<String, Long> readCleanupStateFromFile() throws IOException
```
zpear: Can we create some typed wrapper class instead of using String? Something like:
```
public class SomeKeyspaceTableWrapper extends Wrapper<String>
{

    private final String keyspace;
    private final String columnFamily;

    public SomeKeyspaceTableWrapper(String keyspace, String columnFamily)
    {
        super(keyspace + ":" + columnFamily);
        this.keyspace = keyspace;
        this.columnFamily = columnFamily;
    }
```
https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/com/palantir/cassandra/objects/Wrapper.java

src/java/org/apache/cassandra/service/cleanupstate/CleanupStatePersister.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.codehaus.jackson.map.ObjectMapper;
+
+public class CleanupStatePersister
+{
```
zpear: In the future we might do a similar refactor to track last repair times by keyspace/table. Could we make this class more general use?

src/java/org/apache/cassandra/service/cleanupstate/CleanupState.java
```diff
@@ -0,0 +1,66 @@
...
+        if (tableEntries.containsKey(entryKey) && tableEntries.get(entryKey).compareTo(value) > 0)
+            throw new IllegalArgumentException("Can only update cleanup state entry with increasing timestamp");
+
+        tableEntries.put(entryKey, value);
+        return tableEntries;
+    }
+
+    public Long getMinimumTsOfAllEntries()
+    {
+        if (tableEntries.isEmpty())
```
zpear: How hard would it be to use `Instant` everywhere instead of converting between `Instant` and `Long`? If it's to make serialization easier, could we contain the conversion within the persister class?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2912,6 +2925,14 @@ public int getCurrentGenerationNumber()
         return Gossiper.instance.getCurrentGenerationNumber(FBUtilities.getBroadcastAddress());
     }
 
+    public Instant getLastSuccessfulCleanupTsForNode()
+    {
+        if (cleanupState == null)
+            throw new IllegalStateException("Tracker is null, cleanup state info unretrievable.");
```
zpear: If we fail to create the cleanup state tracker will all the cleanup endpoints that require it fail too until we restart c* and try again? If the initial creation fails, could we retry creating it on these endpoints to avoid the restart?

src/java/org/apache/cassandra/service/cleanupstate/CleanupStatePersister.java
inespot: Meaning adding an interface?

src/java/org/apache/cassandra/service/cleanupstate/CleanupState.java
inespot: Yep, I'll make some changes to contain it into the persister class

src/java/org/apache/cassandra/service/StorageService.java
inespot: So try again and then throw an error if it does not work on retry?

src/java/org/apache/cassandra/service/cleanupstate/CleanupStateTracker.java
inespot: I am not sure CleanupState would really need to exist in this case. CleanupStateTracker could just contain the cache. The logical use I see in CleanupState atm is that it clarifies that we use both a cache and a persister, which are two separate entities. What do you think?

zpear: discussed offline: not a strong preference so OK to stay as is

src/java/org/apache/cassandra/service/cleanupstate/CleanupStatePersister.java
zpear: Discussed offline: basically just a rename so it makes sense to reuse for repairs if/when we do that

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
```diff
@@ -0,0 +1,116 @@
...
+            return new HashMap<>();
+
+        try
+        {
+            return convertMapTypeToKeyspaceTableKeyInstant(OBJECT_MAPPER.readValue(persistentStateFile, Map.class));
+        }
+        catch (Exception e)
+        {
+            log.warn("Failed to retrieve state from file.", persistentStateFile.getAbsolutePath(), e);
+            return new HashMap<>();
```
zpear: If we have a file on disk that we fail to read initially I think it's possible we could:
* populate the cache with an empty map
* the next cleanup update will have the cache keep a map with a single entry
* try to persist the cache to disk
* persister successfully reads the file this time
* overwrite the entire map with the single entry known by the cache

I think we'd have to reconcile the cache with what is on disk when we are eventually able to read the file successfully

test/unit/org/apache/cassandra/service/opstate/CleanupStateTrackerTest.java
```diff
@@ -0,0 +1,115 @@
...
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+public class CleanupStateTrackerTest
+{
+
+    @After
+    public void afterEach(){
+        new File(OpStateTestConstants.TEST_STATE_FILE_LOCATION).delete();
```
zpear: nit: you can use `Files.createTempFile()` which should do some clean up for you

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStateCache.java
```diff
@@ -0,0 +1,67 @@
...
+            throw new IllegalArgumentException("Can only update cache entry with increasing timestamp");
+
+        tableEntries.put(entryKey, value);
+        return tableEntries;
+    }
+
+    public Instant getMinimumTsOfAllEntries()
+    {
+        if (tableEntries.isEmpty())
+            return null;
```
zpear: nit: instead of returning null here and when the persister fails to create the file could we return `Optional.empty()`?

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
inespot: Theoretically, I wonder if the above will be a truly problematic failure mode (for cleanup at least) since the cache should always be more updated than the file. As a result, if my understanding is correct, you are only losing info (instead of having incorrect one). The worse thing that could happen in this case is the subsequent cleanup being delayed. 
Nevertheless, I think your approach is more correct, I can probably read before write and insert updated cache entries instead of overwriting.

test/unit/org/apache/cassandra/service/opstate/CleanupStateTrackerTest.java
inespot: That creates the file in a temp directory correct? Trick is the StatePersister uses the normal `new File` method. I could probably provide a tmp directory path instead of the current `TEST_STATE_FILE_LOCATION` but then I am unclear about when the File would  actually get deleted and if different Test would erroneously interact with the same File.

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
```diff
@@ -0,0 +1,147 @@
...
+    public boolean updateStateInPersistentLocation(Map<KeyspaceTableKey, Instant> updatedEntries)
+    {
+        File persistentStateFile = getOrMaybeCreateStateFile();
+        if (persistentStateFile == null)
+            return false;
+        try
+        {
+            Map<KeyspaceTableKey, Instant> persistentEntries = readStateFromFile(persistentStateFile);
+
+            // We assume inputted entries will always be more up to date compared to the ones from the file
```
inespot, line 74: Do you think that is an acceptable assumption? Or should we explicitly check (and potentially throw)?

zpear: I think the issue is that if something goes wrong with reading the file we'll return an empty map (line 59) instead of an `Optional.empty()` or throwing an exception, which opens us up to edge cases (admittedly not likely and not really a critical error, as you mentioned in a previous comment).

I think if we have some indication that reading the file failed (whichever way you think is cleanest) this method should just return false instead of trying to write afterwards

test/unit/org/apache/cassandra/service/opstate/CleanupStateTrackerTest.java
zpear: I suppose you could do something like:
```
    private static Path directory;
    private static Path path;
    
    @BeforeClass
    public static void before() throws IOException
    {
         directory = Files.createTempDirectory("test-dir");
         path = directory.resolve("actual-file");
    }

    @After
    public void afterEach()
    {
        new File(path.toString()).delete();
    }
```
Which would provide the actual file path like you have above (side note can the KeyspaceTableOpStatePersister take a `Path` instead of a `String` in the constructor?)

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
inespot: Unless I am mistaken somehow, this is what it is currently doing by using the `readStateFromFile` function which throws if it cannot read?

zpear: ooooo I misread! my bad yep this is all good! (thought we were calling readStateFromPersistentLocation)

zpear (approved):

PR #324, opened 2022-03-11
Collapse and track repairs

zpear:
For large tables, tracking repair progress via JMX notifications is problematic. JMX notifications can be lost (which seems to happen relatively frequently on some of our deployments with 300G+ tables) which we assume means the repair failed, even though in many cases it continues progressing. (https://support.datastax.com/s/article/Repair-service-may-fail-due-to-lost-JMX-notifications-from-Cassandra)

Instead of retrying and accumulating more and more repair work, we want to collapse redundant repair requests if the first is still in progress. Additionally, we want to provide a more failsafe way to track repair progress than relying on an uninterrupted stream of JMX notifications.


src/java/org/apache/cassandra/repair/RepairTracker.java
```diff
@@ -0,0 +1,56 @@
...
+package org.apache.cassandra.repair;
+
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.apache.cassandra.service.StorageServiceMBean;
+
+public class RepairTracker
+{
+    private final ConcurrentHashMap<RepairArguments, RepairRunnable> argsToMostRecentRepair;
```
Sam-Kramer: I think it may be worth making this regular hashmaps, and then having all the methods be `synchronized`.  Reason being, I see `argsToMostRecentRepair` as an index, and all changes that happen to `argsToMostRecentRepair` must happen concurrently to both maps. As it stands, I can call `track`, which could put a new value into `argsToMostRecentRepair`, but it doesn't have the lock on `commandToRepairs`, so we could perform a read on `commandToRepairs`, and not see the repair we expect

Sam-Kramer: I also think we need to consider how we clean this up afterwards; I'm not sure what `RepairRunnable` holds references to, but this is technically a reference leak.

test/unit/org/apache/cassandra/repair/RepairTrackerTest.java
```diff
@@ -0,0 +1,128 @@
...
+        RepairRunnable task2 = mock(RepairRunnable.class);
+        doReturn(StorageServiceMBean.ProgressState.IN_PROGRESS).when(task2).getCurrentState();
+        doReturn(2).when(task2).getCommand();
+        tracker.track(2, args, task2);
+
+        assertThat(tracker.getInProgressRepair(args)).contains(2);
+    }
+
+    @Test
+    public void getInProgressRepair_matchesSameArgs()
```
Sam-Kramer, line 163: Could we also test that it doe not return when the repair args are different? I think it'd also be worth while trying a couple of different permutations of the repair args, to ensure your hashcode/equals works as expected!

src/java/org/apache/cassandra/repair/RepairTracker.java
zpear: Good catch! Will do for the first point. For the second, we do want to keep returning the final state a repair hit after it's completed -- in that case we could remove the repair from our 2 current maps and add a third that's just Integer->ProgressState?

Sam-Kramer: Up to you! How is the command ID generated? Is there any chance that we could re-use an older command id for a newer request?

zpear: Repair command is a simple counter starting at 1. The only realistic way I think we'd reuse a command ID would be if c* restarted (which would have all this tracking state wiped anyway).

Sam-Kramer: Then yeah I think that's fine.

zpear: Pushed. Though currently we'll only clean up when the tracker gets invoked so it's possible we maintain the reference to 1 RepairRunnable without cleaning it up:
* sidecar triggers repair
* sidecar bounces so it no longer listens for that repairs progress + does not hit our new progress endpoint
* reference will be maintained until c* bounces or sidecar issues another repair, at which point we'll clean up

Do you think it's worth scheduling the cleanup on an executor to keep us 100% clean?

zpear: discussed offline, having the tracker implement `ProgressListener`

src/java/org/apache/cassandra/repair/RepairTracker.java
```diff
@@ -0,0 +1,159 @@
...
+        maybeUpdateProgressState(event, repairCommand.get());
+    }
+
+    private void updateProgressState(StorageServiceMBean.ProgressState state, int command)
+    {
+        logger.info("Updating state for repair command {} to {}", command, state);
+        commandToProgressState.put(command, state);
+        if (isCompleteState(state))
+        {
+            cleanCompletedRepairs();
```
Sam-Kramer: I think we can just clean a single repair at a time now

Sam-Kramer (approved):

PR #326, opened 2022-03-22
Expose isCleanupRunning endpoint on storage service

inespot:
Creates a StorageService endpoint that returns true if at least one cleanup operation is currently running on node


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2916,6 +2919,13 @@ public int getCurrentGenerationNumber()
         return Gossiper.instance.getCurrentGenerationNumber(FBUtilities.getBroadcastAddress());
     }
 
+    public boolean isCleanupRunning()
```
leonz, line 2920: discussed in person but i don't think this accounts for pending tasks

leonz (approved):

PR #328, opened 2022-03-25
Ignore EOF exceptions for SSL NTE

Sam-Kramer:


src/java/org/apache/cassandra/net/OutboundTcpConnection.java
```diff
@@ -510,7 +511,10 @@ private boolean connect()
                 logger.error("SSL handshake error for outbound connection to " + socket, e);
                 socket = null;
                 if (ENABLE_SSL_NTE) {
-                    StorageService.instance.recordNonTransientError(StorageServiceMBean.NonTransientError.SSL_ERROR, ImmutableMap.of());
+                    if (!(e.getCause() instanceof EOFException)) {
```
leonz, line 516: can you add a comment why?

Sam-Kramer: should be gtg!

leonz (approved):

PR #330, opened 2022-03-28
Do not add left endpoint to list of unreachable nodes

Sam-Kramer:
Do not add left endpoint to list of unreachable nodes

### DIFF
--- a/src/java/org/apache/cassandra/gms/Gossiper.java
+++ b/src/java/org/apache/cassandra/gms/Gossiper.java
@@ -998,7 +998,10 @@ public class Gossiper implements IFailureDetectionEventListener, GossiperMBean
             logger.trace("marking as down {}", addr);
         localState.markDead();
         liveEndpoints.remove(addr);
-        unreachableEndpoints.put(addr, System.nanoTime());
+        if (!justRemovedEndpoints.containsKey(addr))
+        {
+            unreachableEndpoints.put(addr, System.nanoTime());
+        }
         logger.info("InetAddress {} is now DOWN", addr);
         for (IEndpointStateChangeSubscriber subscriber : subscribers)
             subscriber.onDead(addr, localState);


src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -998,7 +998,10 @@ public void markDead(InetAddress addr, EndpointState localState)
             logger.trace("marking as down {}", addr);
         localState.markDead();
         liveEndpoints.remove(addr);
-        unreachableEndpoints.put(addr, System.nanoTime());
+        if (!justRemovedEndpoints.containsKey(addr))
```
Sam-Kramer, line 1001: I _think_ this is fairly safe to do -- unreachable endpoints are mostly used for reporting purposes, and if we still see the node as "left", we shouldn't really care that it's not participating in the cluster.

leonz: Confirming my understanding:
- justRemovedEndpoints is a map of endpoints which have been _quarantined_ as part of being removed from Gossip
- the quarantine lasts for a short amount of time, and then ends (eps are removed from this map)
- if an endpoint is in `justRemovedEndpoints`, we know it has been evicted and so it cannot exist in `liveEndpoints` or `unreachableEndpoints`

This PR prevents us from re-adding nodes to the unreachable endpoint map if they were just removed?  In which situations might this happen otherwise?

Sam-Kramer: (1) & (2) are correct, although to clarify (2), nodes that have left are persisted for 72 hours (or until cluster roll)

`if an endpoint is in justRemovedEndpoints, we know it has been evicted and so it cannot exist in liveEndpoints or unreachableEndpoints` It could end up living in `liveEndpoints`, if the node comes back online for whatever reason. Otherwise, we do expect otherwise that a recently remove endpoint should not be in live/unreachable endpoints, as it's really not a member of gossip anymore. 

`This PR prevents us from re-adding nodes to the unreachable endpoint map if they were just removed? In which situations might this happen otherwise?`

Correct, as I think the race condition that's happening here: node announces it has left, and shuts down. Node has left, so it's quarantined, removed from all the necessary lists. However, there could be the case that, the node going down is seen by the failure detector, and failure detector still has it in it's list of active nodes, therefore marks it as dead -> adds it to the list of unreachable nodes. Or some other mechanism, which influences gossip state.

PR #331, opened 2022-04-11
If node bootstraps to join the cluster, truncate all old data before it does

Sam-Kramer:
If node bootstraps to join the cluster, truncate all old data before it does. 

I think this is relatively safe to do, seeing as if a node that already has tons of data attempts to join the cluster, it will cause corruption (thus effectively data loss).


Sam-Kramer:
For now, let's just set state to be NTE, and later on we can revisit this once we get signal how frequently this is an issue.

PR #332, opened 2022-04-11
Do not bootstrap if previous data is found for keyspace

Sam-Kramer:


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -888,36 +889,36 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
...
+                    {
+                        logger.error("Found previous SSTables {} for keyspace {} and cf {}.", tables, keyspaceName, store.name);
+                        previousDataFound = true;
+                    }
                 }
             }
+
+            if (previousDataFound)
+            {
+                recordNonTransientError(NonTransientError.BOOTSTRAP_ERROR,
```
Sam-Kramer, line 910: Was thinking of sprucing this up a bit, and adding the keyspaces/CF, but that list may be be too long to display for a NTE?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -888,36 +889,35 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
...
             bootstrapManager.awaitBootstrappable();
-            if (SystemKeyspace.bootstrapInProgress())
-            {
-                // on detection of previous bootstrap failure prevent boostrap from proceeding if data directories are
-                // filled up beyond palantir_cassandra.bootstrap_disk_usage_threshold_percentage
-                Entry<Directories.DataDirectory, Double> mostUtilizedDataDir = Directories.getMaxPathToUtilization();
-                Double percentageDataDirUtilization = mostUtilizedDataDir.getValue() * 100;
+            boolean previousDataFound = false;
 
-                if (BOOTSTRAP_DISK_USAGE_THRESHOLD != null && percentageDataDirUtilization > Double.valueOf(BOOTSTRAP_DISK_USAGE_THRESHOLD))
```
Sam-Kramer, line 898: Need to also remove the constants everywhere if people agree this is ok to remove.

src/java/org/apache/cassandra/service/StorageService.java
leonz: Yea that's gonna be way too big.  Probably just knowing that data exists is enough.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -888,36 +889,35 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
...
+            if (previousDataFound)
+            {
+                recordNonTransientError(NonTransientError.BOOTSTRAP_ERROR,
+                                        ImmutableMap.of("previousDataFound", "true"));
+                unsafeDisableNode();
+                // leave node in non-transient error state and prevent it from bootstrapping into the cluster
+                throw new BootstrappingSafetyException("Detected data from previous bootstrap, failling.");
+            }
+
+            if (SystemKeyspace.bootstrapInProgress())
```
leonz, line 917: How likely is this to happen? I would imagine that we'd almost immediately create the keyspaces.

Sam-Kramer: I just re-added this to match trunk; I suspect this code path will only be reachable after we've removed any lone sstables.

leonz (approved):

PR #336, opened 2022-04-21
Lock tokens when removing endpoints to prevent race condition

Sam-Kramer:
Locks all changes to token metadata when an old endpoint is removed from the storage ring. 

Only affects clusters where nodes > RF. 

This is to prevent a race condition where: 
1- A node changes IP addresses, thus the endpoint is removed from the storage ring. 
2- A new in-bound request comes in after this, and when calculating which endpoints to contact, it chooses the next closest (via token value) node in the rack, as we deleted the previous endpoint, but did not add it's new endpoint. 
3- We re-add the _new_ endpoint to the storage ring. However, the cached token map is not invalidated at this juncture.
4- Only after the next gossip round (from a topology change) will we detect the new endpoint. At this time, we will start to read/write from the correct node. 

This change effectively makings the deletion/addition into a single atomic statement.

### DIFF
--- a/src/java/org/apache/cassandra/locator/TokenMetadata.java
+++ b/src/java/org/apache/cassandra/locator/TokenMetadata.java
@@ -141,6 +141,7 @@ public class TokenMetadata
     {
         int n = 0;
         Collection<Range<Token>> sourceRanges = getPrimaryRangesFor(getTokens(source));
+        publicLock.readLock().lock();
         lock.readLock().lock();
         try
         {
@@ -152,6 +153,7 @@ public class TokenMetadata
         finally
         {
             lock.readLock().unlock();
+            publicLock.readLock().unlock();
         }
         return n;
     }
@@ -232,8 +234,8 @@ public class TokenMetadata
         assert hostId != null;
         assert endpoint != null;
 
-        lock.writeLock().lock();
         publicLock.readLock().lock();
+        lock.writeLock().lock();
         try
         {
             InetAddress storedEp = endpointToHostIdMap.inverse().get(hostId);
@@ -265,8 +267,8 @@ public class TokenMetadata
     /** Return the unique host ID for an end-point. */
     public UUID getHostId(InetAddress endpoint)
     {
-        lock.readLock().lock();
         publicLock.readLock().lock();
+        lock.readLock().lock();
         try
         {
             return endpointToHostIdMap.get(endpoint);
@@ -281,8 +283,8 @@ public class TokenMetadata
     /** Return the end-point for a unique host ID */
     public InetAddress getEndpointForHostId(UUID hostId)
     {
-        lock.readLock().lock();
         publicLock.readLock().lock();
+        lock.readLock().lock();
         try
         {
             return endpointToHostIdMap.inverse().get(hostId);
@@ -297,8 +299,8 @@ public class TokenMetadata
     /** @return a copy of the endpoint-to-id map for read-only operations */
     public Map<InetAddress, UUID> getEndpointToHostIdMapForReading()
     {
-        lock.readLock().lock();
         publicLock.readLock().lock();
+        lock.readLock().lock();
         try
         {
             Map<InetAddress, UUID> readMap = new HashMap<>();
@@ -328,8 +330,8 @@ public class TokenMetadata
         assert tokens != null && !tokens.isEmpty();
         assert endpoint != null;
 
-        lock.writeLock().lock();
         publicLock.readLock().lock();
+        lock.writeLock().lock();
         try
         {
             InetAddress oldEndpoint;
@@ -362,8 +364,8 @@ public class TokenMetadata
         assert replacingTokens != null && !replacingTokens.isEmpty();
         assert newNode != null && oldNode != null;
 
-        lock.writeLock().lock();
         publicLock.readLock().lock();
+        lock.writeLock().lock();
         try
         {
             Collection<Token> oldNodeTokens = tokenToEndpointMap.inverse().get(oldNode);
@@ -400,8 +402,8 @@ public class TokenMetadata
     {
         assert tokens != null && !tokens.isEmpty();
 
-        lock.writeLock().lock();
         publicLock.readLock().lock();
+        lock.writeLock().lock();
         try
         {
             for (Token token : tokens)
@@ -418,8 +420,8 @@ public class TokenMetadata
     {
         assert endpoint != null;
 
-        lock.writeLock().lock();
         publicLock.readLock().lock();
+        lock.writeLock().lock();
         try
         {
             leavingEndpoints.add(endpoint);
@@ -440,8 +442,8 @@ public class TokenMetadata
     {
         assert endpoint != null;
 
-        lock.writeLock().lock();
         publicLock.readLock().lock();
+        lock.writeLock().lock();
 
         try
         {
@@ -458,8 +460,9 @@ public class TokenMetadata
     {
         assert endpoint != null;
 
-        lock.writeLock().lock();
         publicLock.readLock().lock();
+        lock.writeLock().lock();
+
         try
         {
             bootstrapTokens.removeValue(endpoint);
@@ -488,8 +491,9 @@ public class TokenMetadata
     {
         assert endpoint != null;
 
-        lock.writeLock().lock();
         publicLock.readLock().lock();
+        lock.writeLock().lock();
+
         try
         {
             logger.info("Updating topology for {}", endpoint);
@@ -509,8 +513,8 @@ public class TokenMetadata
      */
     public void updateTopology()
     {
-        lock.writeLock().lock();
         publicLock.readLock().lock();
+        lock.writeLock().lock();
         try
         {
             logger.info("Updating topology for all endpoints that have changed");
@@ -532,8 +536,8 @@ public class TokenMetadata
     {
         assert endpoint != null;
 
-        lock.writeLock().lock();
         publicLock.readLock().lock();
+        lock.writeLock().lock();
         try
         {
             for (Pair<Token, InetAddress> pair : movingEndpoints)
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/locator/TokenMetadata.java
```diff
@@ -253,34 +266,39 @@ public void updateHostId(UUID hostId, InetAddress endpoint)
...
         lock.readLock().lock();
+        publicLock.readLock().lock();
         try
         {
             return endpointToHostIdMap.get(endpoint);
         }
         finally
         {
             lock.readLock().unlock();
+            publicLock.readLock().unlock();
```
schlosna, line 279: Should these be unlocked in reverse order of locking?

```suggestion
            publicLock.readLock().unlock();

            lock.readLock().unlock();
```

src/java/org/apache/cassandra/locator/TokenMetadata.java
```diff
@@ -174,6 +183,7 @@ public void updateNormalTokens(Multimap<InetAddress, Token> endpointTokens)
         if (endpointTokens.isEmpty())
             return;
 
+        publicLock.readLock().lock();
         lock.writeLock().lock();
```
schlosna, line 177: Should we always acquire `lock` before `publicLock` to ensure consistent ordering and no deadlocks? Do we have concurrency tests to hammer these and verify correctness and liveness guarantees?

src/java/org/apache/cassandra/locator/TokenMetadata.java
```diff
@@ -672,7 +726,7 @@ private static TokenMetadata removeEndpoints(TokenMetadata allLeftMetadata, Set<
      */
     public TokenMetadata cloneAfterAllSettled()
     {
-        lock.readLock().lock();
+        lock.readLock().lock(); publicLock.readLock().lock();
```
schlosna:
```suggestion
        lock.readLock().lock();
        publicLock.readLock().lock();
```

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2296,105 +2296,117 @@ private void handleStateNormal(final InetAddress endpoint, final String status)
...
+                && (hostId.equals(Gossiper.instance.getHostId(DatabaseDescriptor.getReplaceAddress()))))
+                logger.warn("Not updating token metadata for {} because I am replacing it", endpoint);
+            else
+            {
+                if (existing != null && !existing.equals(endpoint))
+                {
+                    // It's safe for us to say we acquired the lock before we've, as this is simply used for calling unlock.
+                    // Unlock will throw if a separate thread that is not holding it tries to unlock it.
+                    acquiredTokenLock = true;
+                    tokenMetadata.lock();
```
schlosna, line 2311: Seems odd to flip the flag before we actually lock
```suggestion

                    tokenMetadata.lock();
                    acquiredTokenLock = true;
```

Sam-Kramer: Yup so there is a trade-off here: If we acquire _after_ we lock, there is a chance that we do not reach the `acquiredTokenLock` statement (such as an interrupt), which places us into a state of deadlock. Whereas if we assign the lock _before_ we acquire it, the worst that can happen is we throw an exception when we try to unlock the lock as our current thread doesn't own it. And seeing as an exception has already been thrown in this case (i.e. lock() failed), it should be relatively harmless to throw yet _another_ exception.

src/java/org/apache/cassandra/locator/TokenMetadata.java
Sam-Kramer: Good flag -- yes, my plan is after we smoke tested that the change actually works, my next step is to create a couple of tests to ensure we cannot deadlock.

schlosna:
At first glance, I don't see anything specific upstream that would address these potential race conditions but curious if you have we compared our 2.2.x fork to upstream? For example, there have been a lot of upstream changes around Paxos, simulating and fixing correctness and linearizability. 

Some potentially relevant bits:

- [CASSANDRA-16873: Tolerate missing DNS entry when completing a host replacement](https://issues.apache.org/jira/browse/CASSANDRA-16873) - commit https://github.com/apache/cassandra/commit/f59411f1c985043850154971e9c4066013f355cb
- [CASSANDRA-14781](https://issues.apache.org/jira/browse/CASSANDRA-14781) - commit https://github.com/apache/cassandra/commit/bc18b4dd4e33020d0d58c3701077d0af5c39bce6
- [CASSANDRA-16673: Avoid race in AbstractReplicationStrategy endpoint caching](https://issues.apache.org/jira/browse/CASSANDRA-16673)  - commit https://github.com/apache/cassandra/commit/b1a8a56c563b85ab9a34d3bbf9c16278dd441157
- [CEP-10: Cluster and Code Simulations](https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-10%3A+Cluster+and+Code+Simulations)
- [CEP-14: Paxos Improvements](https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-14%3A+Paxos+Improvements)
- https://github.com/apache/cassandra/commits/trunk/src/java/org/apache/cassandra/locator/TokenMetadata.java
- https://github.com/apache/cassandra/commits/trunk/src/java/org/apache/cassandra/service/StorageService.java

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2296,105 +2296,117 @@ private void handleStateNormal(final InetAddress endpoint, final String status)
...
-            && (hostId.equals(Gossiper.instance.getHostId(DatabaseDescriptor.getReplaceAddress()))))
-            logger.warn("Not updating token metadata for {} because I am replacing it", endpoint);
-        else
-        {
-            if (existing != null && !existing.equals(endpoint))
+        boolean acquiredTokenLock = false;
+        try {
+            if (replacing && isReplacingSameAddress() && Gossiper.instance.getEndpointStateForEndpoint(DatabaseDescriptor.getReplaceAddress()) != null
+                && (hostId.equals(Gossiper.instance.getHostId(DatabaseDescriptor.getReplaceAddress()))))
+                logger.warn("Not updating token metadata for {} because I am replacing it", endpoint);
```
schlosna, line 2303: Should we pull in https://github.com/apache/cassandra/commit/f59411f1c985043850154971e9c4066013f355cb which extracts this condition out to a method and handles `UnknownHostException`?

https://github.com/apache/cassandra/commit/f59411f1c985043850154971e9c4066013f355cb#diff-9bf2c26bc294ef9085e16bf287490223665eaa2eb8ec24bcf5bd8653c713644bR2722-R2752

tpetracca:
@Sam-Kramer @leonz I will try to get a review in for this sometime this evening. Can we also open up an apache ticket. Don't need to block merging this on that, but I'd really like to get the community's eyes on it and see if there isn't something that should eventually be upstreamed here. My guess is that folks are hitting this through increased k8s usage across the c* community and it isn't unique to us. We probably just notice corruption / consider things corrupted more easily because we require durable writes (whereas most folks do not).

Sam-Kramer:
Sorry for the slow reply here @schlosna 

- CEP14 is definitely on our radar, and I believe is something we're wanting to backport once it's ready. 
- CEP10 would also be ideal, although I do not think the backport here will be trivial (appears to be mostly written for C* 4.x sadly :( )
- [CASSANDRA-16873: Tolerate missing DNS entry when completing a host replacement](https://issues.apache.org/jira/browse/CASSANDRA-16873) -- this is probably worth doing in a separate PR, just to keep the git commit history clean when we commit into develop (as we squash). 
- [CASSANDRA-14781](https://issues.apache.org/jira/browse/CASSANDRA-14781) & https://issues.apache.org/jira/browse/CASSANDRA-16673 also peek my interest a bit -- I'll take a further glance on this Monday. Not that it solves the original issue, but curious to see if it solves another race condition we're facing.

Sam-Kramer:
Also just noticed that the PR for 14781 contained a typo by the commit author, I think it's for 14871 https://issues.apache.org/jira/browse/CASSANDRA-14871. Hilariously coincidental, ticket 14781 was filed by Tom and Leon!

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2296,105 +2296,117 @@ private void handleStateNormal(final InetAddress endpoint, final String status)
         // Order Matters, TM.updateHostID() should be called before TM.updateNormalToken(), (see CASSANDRA-4300).
         UUID hostId = Gossiper.instance.getHostId(endpoint);
         InetAddress existing = tokenMetadata.getEndpointForHostId(hostId);
-        if (replacing && isReplacingSameAddress() && Gossiper.instance.getEndpointStateForEndpoint(DatabaseDescriptor.getReplaceAddress()) != null
-            && (hostId.equals(Gossiper.instance.getHostId(DatabaseDescriptor.getReplaceAddress()))))
-            logger.warn("Not updating token metadata for {} because I am replacing it", endpoint);
-        else
-        {
-            if (existing != null && !existing.equals(endpoint))
+        boolean acquiredTokenLock = false;
```
leonz, line 2299: Should we synchronize this entire function?  Not sure if we ever expect more than one thread to ever enter here anyway, but might be safer to ensure that.

leonz, line 2299: This diff is hard to read.  Did anything significantly change besides adding the lock?

Sam-Kramer: There should only be a single thread for gossip, but I agree let's be defensive here.

Sam-Kramer: Nope, only the lock, unlock, and the `try {} finally {}`

leonz (approved):

PR #337, opened 2022-04-26
Throw when resumeBootstrap is called, and fail bootstraps if previous availableRanges exist for non-system keyspaces.

Sam-Kramer:


leonz (approved):
Seems fine to me 👍

PR #339, opened 2022-05-03
Do not cleanup if nodes are boostrapping

inespot:
Prevent StorageService from triggering cleanups if at least one node in the cluster is bootstrapping

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -2966,6 +2966,9 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
 
     public int forceKeyspaceCleanup(int jobs, String keyspaceName, String... columnFamilies) throws IOException, ExecutionException, InterruptedException
     {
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2966,6 +2966,9 @@ public int forceKeyspaceCleanup(String keyspaceName, String... columnFamilies) t
 
     public int forceKeyspaceCleanup(int jobs, String keyspaceName, String... columnFamilies) throws IOException, ExecutionException, InterruptedException
     {
+        if (getJoiningNodes().size() > 0)
+            throw new RuntimeException("Triggering a cleanup while at least one node is bootstrapping can cause corruption");
```
leonz: Let's add "Cleanup operation not permitted." to the start of that error?

leonz (approved):

inespot:
Note that this check does not account for nodes which are in the WAITING_ON_BOOTSTRAP state, as we unfortunately have not yet found a simple way to include them

PR #344, opened 2022-05-12
Probe endpoint to inspect if sstables need cleanup

leonz:
We have found evidence that running a decommission while nodes require cleanup can result in data corruption, when ghost data is pulled back into a token range.

As a shortcut, the cleanup function checks if sstables need to be cleaned before performing the compaction.  This PR leverages that codepath to create a separate endpoint which _only_ performs the check.

For the sake of keeping the code paths similar, I leverage the CompactionManager abstractions and I use the compaction executor to run this check.  It should be pretty fast, and isn't on any critical path.

The intention is that any user who wishes to run a decommission should first run this check against all nodes.  We can add this check to our internal tooling.

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
@@ -370,7 +370,7 @@ public class CompactionManager implements CompactionManagerMBean
                         return callable.execute(txn);
                     }
                 };
-                Future<T> fut = (ListenableFuture<T>) executor.submitIfRunning(callableGeneric, "paralell sstable callable operation");
+                Future<T> fut = (ListenableFuture<T>) executor.submitIfRunning(callableGeneric, "parallel sstable callable operation");
                 if (!fut.isCancelled())
                     futures.add(fut);
                 else
--- a/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
@@ -333,10 +333,77 @@ public class CompactionManager implements CompactionManagerMBean
         }
     }
 
+    /**
+     * Run an operation over all sstables using jobs threads
+     *
+     * @param cfs the column family store to run the operation on
+     * @param callable the callable operation to run
+     * @param jobs the number of threads to use - 0 means use all available. It never uses more than concurrent_compactors threads
+     * @return status of the operation
+     * @throws ExecutionException
+     * @throws InterruptedException
+     */
+    @SuppressWarnings("resource")
+    private <T> Optional<List<T>> parallelAllSSTableCallable(final ColumnFamilyStore cfs, final OneSSTableCallable<T> callable, int jobs, OperationType operationType) throws ExecutionException, InterruptedException
+    {
+        List<LifecycleTransaction> transactions = new ArrayList<>();
+        try (LifecycleTransaction compacting = cfs.markAllCompacting(operationType))
+        {
+            Iterable<SSTableReader> sstables = compacting != null ? Lists.newArrayList(callable.filterSSTables(compacting)) : Collections.<SSTableReader>emptyList();
+            if (Iterables.isEmpty(sstables))
+            {
+                logger.info("No sstables for {}.{}", cfs.keyspace.getName(), cfs.name);
+                return Optional.of(ImmutableList.of());
+            }
+
+            List<Future<T>> futures = new ArrayList<>();
+
+            for (final SSTableReader sstable : sstables)
+            {
+                final LifecycleTransaction txn = compacting.split(singleton(sstable));
+                transactions.add(txn);
+                Callable<T> callableGeneric = new Callable<T>()
+                {
+                    @Override
+                    public T call() throws Exception
+                    {
+                        return callable.execute(txn);
+                    }
+                };
+                Future<T> fut = (ListenableFuture<T>) executor.submitIfRunning(callableGeneric, "paralell sstable callable operation");
+                if (!fut.isCancelled())
+                    futures.add(fut);
+                else
+                    return Optional.empty();
+
+                if (jobs > 0 && futures.size() == jobs)
+                {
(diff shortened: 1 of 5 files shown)


src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -333,12 +333,79 @@ public Object call() throws Exception
...
+                transactions.add(txn);
+                Callable<T> callableGeneric = new Callable<T>()
+                {
+                    @Override
+                    public T call() throws Exception
+                    {
+                        return callable.execute(txn);
+                    }
+                };
+                Future<T> fut = (ListenableFuture<T>) executor.submitIfRunning(callableGeneric, "paralell sstable callable operation");
```
inespot: Typo

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -333,12 +333,79 @@ public Object call() throws Exception
...
+     * @param jobs the number of threads to use - 0 means use all available. It never uses more than concurrent_compactors threads
+     * @return status of the operation
+     * @throws ExecutionException
+     * @throws InterruptedException
+     */
+    @SuppressWarnings("resource")
+    private <T> Optional<List<T>> parallelAllSSTableCallable(final ColumnFamilyStore cfs, final OneSSTableCallable<T> callable, int jobs, OperationType operationType) throws ExecutionException, InterruptedException
+    {
+        List<LifecycleTransaction> transactions = new ArrayList<>();
+        try (LifecycleTransaction compacting = cfs.markAllCompacting(operationType))
```
inespot, line 350: To make sure I understand this correctly, this call opens a cleanup transaction on all sstable, which we are then using to understand whether a cleanup is actually necessary?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2993,6 +2993,24 @@ public int forceKeyspaceCleanup(int jobs, String keyspaceName, String... columnF
         }
     }
 
+    public boolean isKeyspaceFullyClean(int jobs, String keyspaceName, String... columnFamilies) throws IOException, ExecutionException, InterruptedException
+    {
+        if (getJoiningNodes().size() > 0)
+            throw new RuntimeException("Cleanup operation not permitted: triggering cleanups while at least one node is bootstrapping can cause corruption");
```
inespot: Should we switch this error message and the next to something like "Cannot check if a keyspace is clean while at least one node is bootstrapping"? With the current ones, it seems like we are actually trying to trigger cleanup operations on the given keyspace which might be confusing for the reader

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -845,6 +956,23 @@ public static boolean needsCleanup(SSTableReader sstable, Collection<Range<Token
...
+    {
+        SSTableReader sstable = txn.onlyOne();
+        if (!hasIndexes && !new Bounds<>(sstable.first.getToken(), sstable.last.getToken()).intersects(ranges))
+        {
+            txn.obsoleteOriginals();
+            txn.finish();
+            return false;
+        }
+        if (!needsCleanup(sstable, ranges))
+        {
```
inespot, line 969: This might be a dumb question but I don't fully understand why we don't (need to) close the transaction at this point. From my current understanding of things, it seems that the cleanupOne function does not do it as well.

leonz: So, I took a further peek and I only think you need to run `txn.finish()` after you run `txn.obsoleteOriginals()`, because you are actually changing the table in that transaction.

For all other cases, the transaction will ultimately get closed in the finally block of the `parallelSstableCallable`:

```
        finally
        {
            Throwable fail = Throwables.close(null, transactions);
            if (fail != null)
                logger.error("Failed to cleanup lifecycle transactions {}", fail);
        }

File: src/java/org/apache/cassandra/db/compaction/CompactionManager.java (hunk 2 above)
Comment (leonz) [reply src/java/org/apache/cassandra/db/compaction/CompactionManager.java:350]: yes

File: src/java/org/apache/cassandra/service/StorageService.java (hunk 3 above)
Comment (leonz) [reply src/java/org/apache/cassandra/service/StorageService.java]: fixed

File: src/java/org/apache/cassandra/db/compaction/CompactionManager.java (hunk 4 above)
Comment (leonz) [reply src/java/org/apache/cassandra/db/compaction/CompactionManager.java:969]: By the way I found this relatively easily by scanning the PR that added Lifecycle transactions (from git blame): https://github.com/apache/cassandra/commit/e5a76bdb5fc04ffa16b8becaa7877186226c3b32

**Comment by leonz** [comment]
Validated that this works as expected:

1. Bootstrapped a new node to a 3 node cluster.

2. Once it joined, checked if it had tables to clean:
```
$> checkFullyCleaned
result: true
```

3. Checked on one of the original nodes.  Saw that it was not fully cleaned.  Ran a cleanup, and then saw that it was true.
```
$ ./service/bin/nodetool status
Datacenter: DC11
================
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address      Load       Tokens       Owns    Host ID                               Rack
UN  10.2.32.178  1.48 MB    512          ?       bf4c9197-f2e1-4fa5-98ca-f11cf5e6c25a  RAC1
UN  10.2.33.104  1.34 MB    512          ?       2cdfd21e-9d20-4f83-8c0f-9d2a5f382894  RAC1
UN  10.2.32.232  14.53 KB   512          ?       ef1e89b7-3947-4618-8071-17e128eb5308  RAC1
UN  10.2.32.153  1.36 MB    512          ?       19f96aee-9971-41e2-a7a3-e695d3bdb533  RAC1
$> checkFullyCleaned
result: false

$ ./service/bin/nodetool cleanup

$> checkFullyCleaned
result: true
```

inespot (approved):

PR #348, opened 2022-05-27
verify keys on write

Sam-Kramer:


src/java/com/palantir/cassandra/utils/MutationVerificationUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class MutationVerificationUtils
+{
+    private static final boolean VERIFY_KEYS_ON_WRITE = Boolean.valueOf(System.getProperty("palantir_cassandra.verify_keys_on_write", "false"));
+    private static final Logger logger = LoggerFactory.getLogger(MutationVerificationUtils.class);
+
+    private MutationVerificationUtils() {}
+
+    public static void verifyMutation(Mutation mutation) {
+        if (VERIFY_KEYS_ON_WRITE) {
```
leonz: nit: save an indent 

```
if (!VERIFY_KEYS_ON_WRITE) {
  return;
}

File: src/java/com/palantir/cassandra/utils/MutationVerificationUtils.java [hunk 2]
```diff
@@ -0,0 +1,63 @@
...
+                Collection<InetAddress> pendingEndpoints = StorageService.instance.getTokenMetadata().pendingEndpointsFor(tk, mutation.getKeyspaceName());
+
+                if(!naturalEndpoints.contains(FBUtilities.getBroadcastAddress()) && !pendingEndpoints.contains(FBUtilities.getBroadcastAddress())) {
+                    keyspace.metric.invalidMutations.inc();
+                    logger.error("Cannot apply mutation as this host {} does not contain key {}. Only hosts {} and {} do.",
+                                 FBUtilities.getBroadcastAddress(),
+                                 mutation.key(),
+                                 naturalEndpoints,
+                                 pendingEndpoints);
+                    throw new RuntimeException("Cannot apply mutation as this host does not contain key.");
```
leonz: i assume the outcome here is that we fail the write?

Sam-Kramer: Yes! And thinking through QUORUM and ALL, QUORUM requests should still succeed, as we'd expect at worst only one node to be viewed incorrectly here, and for ALL, this would only be broken "temporarily" until gossip caught up.

Sam-Kramer: In the case that it is more than one node, the current state of the coordinator is so bad that we actually do want to fail hard

leonz (approved):

PR #352, opened 2022-06-06
Fix expansions

zpear:
Synchronizing `handleStateNormal()`  in https://github.com/palantir/cassandra/pull/336 may end up blocking gossip on bootstrapping nodes. 

main takes out a lock on the StorageService instance while initializing + bootstrapping "this" node. Then, the gossip thread will block on acquiring a lock on the StorageService instance after beginning gossip w/ another node in the cluster and attempting to notify subscribers (one of which is the StorageService) of the change. As the lock is already taken out by the main thread we block and may be prevented from contacting any seeds successfully (then crash).

image

image (4)

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -2265,7 +2265,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
      *
      * @param endpoint node
      */
-    private synchronized void handleStateNormal(final InetAddress endpoint, final String status)
+    private void handleStateNormal(final InetAddress endpoint, final String status)
     {
         Collection<Token> tokens = getTokensFor(endpoint);
         Set<Token> tokensToUpdateInMetadata = new HashSet<>();


leonz:
@schlosna fysa

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2265,7 +2265,7 @@ private void handleStateBootreplacing(InetAddress newNode, String[] pieces)
      *
      * @param endpoint node
      */
-    private synchronized void handleStateNormal(final InetAddress endpoint, final String status)
+    private void handleStateNormal(final InetAddress endpoint, final String status)
```
schlosna, line 2268: Which specific data structures was the `synchronized` intending to guard reads/writes? Are there specific JMM effects that will no longer hold with removal?

zpear: I believe the `synchronized` was added as a "why not be even safer" during code review, not during testing on RCs (see: https://github.com/palantir/cassandra/pull/336/files#r858486528). The primary intention as I understand was guarding reads/writes to TokenMetadata during node IP replacements, which are still in place (cc @Sam-Kramer).

leonz (approved):
This is sound to me.

Sam-Kramer (approved):

src/java/org/apache/cassandra/service/StorageService.java
schlosna: Got it, thanks!

PR #353, opened 2022-06-16
Make cleanup state file writes atomic

inespot:
Writes to the cleanup state persistent location have been seen failing mid-air, leaving the file in an unparsable state. This PR makes sure writes are now atomic.

### DIFF
--- a/src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
+++ b/src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
@@ -29,9 +29,11 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 
+import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
```diff
@@ -117,18 +127,39 @@ private Map<KeyspaceTableKey, Instant> readStateFromFile(File file) throws IOExc
...
+            atomicWritetoFile(file, OBJECT_MAPPER.writeValueAsString(convertMapTypeToStringLong(updatedEntries)));
+            return true;
+        }
+        catch (IOException e)
+        {
+            return false;
+        }
+    }
+
+    private void atomicWritetoFile(File file, String content) throws IOException
```
leonz: you probably want to synchronize this too right? otherwise two writes could clobber each other

inespot: Yep, totally right, especially since we now sometimes write at read time as well

leonz (approved):

PR #354, opened 2022-06-16
Ignore obsolete entries in getMinimumTs for cleanup state

inespot:
Before PR: Deleted tables are not removed from the cleanup persistent state -> when a node wide cleanup is triggered, it only considers existing tables, meaning cleanup ts min will always equal the last time a deleted table was cleaned and cleanups will continue getting triggered.

PR: Ignore obsolete entries in getMinimumTs and update the cache accordingly. This will eventually lead to updating the persistent state as well.

### DIFF
--- a/src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStateCache.java
+++ b/src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStateCache.java
@@ -23,10 +23,15 @@ import java.time.Instant;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Sets;
+
+import org.apache.cassandra.config.Schema;
 
 public class KeyspaceTableOpStateCache
 {
@@ -63,6 +68,19 @@ public class KeyspaceTableOpStateCache
         if (tableEntries.isEmpty())
             return Optional.empty();
 
+        // Remove potentially deleted keyspaces and tables from entries
+        Set<KeyspaceTableKey> invalidEntries =
+            Sets.difference(tableEntries.keySet(), Sets.intersection(tableEntries.keySet(), getValidKeyspaceTableEntries()));
+        invalidEntries.forEach(tableEntries::remove);
+
         return Optional.of(Collections.min(tableEntries.values()));
+    }
+
+    @VisibleForTesting
+    Set<KeyspaceTableKey> getValidKeyspaceTableEntries(){
+        return Schema.instance.getKeyspaces().stream()
+                              .flatMap(keyspace -> Schema.instance.getKeyspaceInstance(keyspace).getColumnFamilyStores().stream()
+                                                                  .map(cf -> KeyspaceTableKey.of(keyspace, cf.getColumnFamilyName())))
+                              .collect(Collectors.toSet());
     }
 }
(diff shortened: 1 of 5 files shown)


src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStateCache.java
```diff
@@ -63,6 +69,25 @@ public Optional<Instant> getMinimumTsOfAllEntries()
         if (tableEntries.isEmpty())
             return Optional.empty();
 
+        // Remove potentially deleted keyspaces and tables from entries
+        Set<KeyspaceTableKey> invalidEntries =
+            Sets.difference(tableEntries.keySet(), Sets.intersection(tableEntries.keySet(), getValidKeyspaceTableEntries()));
+        invalidEntries.forEach(tableEntries::remove);
+
         return Optional.of(Collections.min(tableEntries.values()));
```
zpear, line 77: Is it possible we have a couple successful cleanups in the cache but we haven't been able to read from disk yet? In that case we would return the min of what's in the cache, which could be later than some other table that's logged on disk and avoid triggering a cleanup even if it might have been needed?

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStateCache.java
```diff
@@ -63,6 +69,25 @@ public Optional<Instant> getMinimumTsOfAllEntries()
...
+    Set<KeyspaceTableKey> getValidKeyspaceTableEntries(){
+        Set<KeyspaceTableKey> validEntries = new HashSet<>();
+        Schema.instance.getKeyspaces().forEach(keyspace -> {
+            Set<KeyspaceTableKey> keyspaceEntries = Schema.instance.getKeyspaceInstance(keyspace).getColumnFamilyStores()
+                                                                   .stream()
+                                                                   .map(cf -> KeyspaceTableKey.of(keyspace, cf.getColumnFamilyName()))
+                                                                   .collect(Collectors.toSet());
+            validEntries.addAll(keyspaceEntries);
+        });
+        return validEntries;
```
zpear: nit: can 1-line this

```suggestion
return Schema.instance.getKeyspaces().stream()
                              .map(keyspace -> Schema.instance.getKeyspaceInstance(keyspace).getColumnFamilyStores()
                                                              .stream()
                                                              .map(cf -> KeyspaceTableKey.of(keyspace, cf.getColumnFamilyName()))
                                                              .collect(Collectors.toSet()))
                              .flatMap(Set::stream)
                              .collect(Collectors.toSet());
```

inespot: I tried using flatMap first but for some reason this converts the stream to a Stream<Object> type which makes the eclipse-warnings fail (see commit above)

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStateCache.java
inespot: Yes this is correct, but the assumption was that this behavior was acceptable. You would indeed miss one cleanup, but you would probably pick up on it at a later time (next time we successfully read from the file or next time a bootstrap is triggered).

zpear: gotcha, if we've decided that's fine I'm cool with it, just wanted to make sure

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStateCache.java
zpear: Hm that's odd. I see [what you're talking about](https://app.circleci.com/pipelines/github/palantir/cassandra/2004/workflows/dfe75481-f000-48d2-ad69-d609888847ad), yet I pushed another branch with your old way and my suggested way here but they both passed 🤔 
https://app.circleci.com/pipelines/github/palantir/cassandra/2012/workflows/f71fb1ed-40f9-49e6-bb6c-e00776f58581

https://app.circleci.com/pipelines/github/palantir/cassandra/2013/workflows/89778490-a4fc-4043-b25d-0c3581d1c816

inespot: Strange ... I'll push your suggestion then, hopefully the time it failed was just a fluke

zpear (approved):

PR #369, opened 2022-10-04
Add thrift method for obtaining list of host IDs that the current node currently sees

Sam-Kramer:


src/java/org/apache/cassandra/thrift/CassandraServer.java
```diff
@@ -1671,6 +1672,11 @@ public Map<String, String> describe_token_map() throws InvalidRequestException
         return StorageService.instance.getTokenToEndpointMap();
     }
 
+    public List<String> get_host_ids() throws TException
+    {
+        return StorageService.instance.getHostIdMap().keySet().stream().collect(Collectors.toList());
```
zpear: This should be the values of the map right? Right now this looks like it'd be the addresses https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/service/StorageService.java#L1945-L1953

Sam-Kramer: good catch -- my mistake!

zpear (approved):

PR #371, opened 2022-10-06
Add cleanup state at columnFamily creation time

inespot:
Adds cleanup state to <keyspace,cf> pair when a new cf is created.
To ensure that we account for keyspaces/tables without state in our LastSuccessfulCleanupTs computation, return empty ts (equivalent to EPOCH) if we find that some entries are missing from our state.


src/java/org/apache/cassandra/service/opstate/KeyspaceColumnFamilyKey.java
```diff
@@ -20,29 +20,29 @@
 
 import com.palantir.cassandra.objects.Wrapper;
 
-public class KeyspaceTableKey extends Wrapper<String>
+public class KeyspaceColumnFamilyKey extends Wrapper<String>
```
lyubent: Is there a reason for this rename? - it will make merging Cassandra's trunk more complicated and bloats the pr significantly.

src/java/org/apache/cassandra/service/opstate/KeyspaceColumnFamilyOpStateCache.java
```diff
@@ -34,56 +33,63 @@
 
 import org.apache.cassandra.config.Schema;
 
-public class KeyspaceTableOpStateCache
+public class KeyspaceColumnFamilyOpStateCache
```
lyubent: As above - do we need to rename?

src/java/org/apache/cassandra/service/opstate/KeyspaceColumnFamilyOpStatePersister.java
```diff
@@ -36,22 +36,22 @@
 import org.codehaus.jackson.JsonParseException;
 import org.codehaus.jackson.map.ObjectMapper;
 
-public class KeyspaceTableOpStatePersister
+public class KeyspaceColumnFamilyOpStatePersister
```
lyubent: As above - do we need to rename?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -3012,6 +3012,11 @@ public int forceKeyspaceCleanup(String keyspaceName, String... columnFamilies) t
         return forceKeyspaceCleanup(0, keyspaceName, columnFamilies);
     }
 
+    public void createCleanupEntryForCfIfNotExists(String keyspaceName, String columnFamily, Optional<Instant> cleanupTs)
+    {
+        cleanupState.createCleanupEntryForCfIfNotExists(keyspaceName, columnFamily, cleanupTs);
+    }
```
lyubent: What's the advantage of pulling this into its own function instead of directly modifying the `cleanupState.createCleanupEntryForTableIfNotExists(...)` call [here](https://github.com/palantir/cassandra/pull/371/files#diff-9bf2c26bc294ef9085e16bf287490223665eaa2eb8ec24bcf5bd8653c713644bL3029)?

src/java/org/apache/cassandra/service/opstate/KeyspaceColumnFamilyOpStateCache.java
```diff
@@ -34,56 +33,63 @@
...
+            Sets.difference(cfEntries.keySet(), Sets.intersection(cfEntries.keySet(), validEntries));
+        invalidEntries.forEach(cfEntries::remove);
+
+        // If not all keyspaces/tables have a tracked last cleanup ts, default to empty value
+        Set<KeyspaceColumnFamilyKey> missingEntries = Sets.difference(validEntries, cfEntries.keySet());
+        if (!missingEntries.isEmpty())
+        {
+            return Optional.empty();
+        }
+        return Optional.of(Collections.min(cfEntries.values()));
```
lyubent: Pref ternary op here and also reverse condition to return `Optional.of(Collections.min(cfEntries.values()))` if  `missingEntries` is empty to simplify reading this:

```java
return missingEntries.isEmpty() ? Optional.of(Collections.min(cfEntries.values())) : Optional.empty();
```

lyubent (changes requested):
PR looks good functionally - but I'm concerned with some aesthetics around renaming classes and the effect that might have on our ability to merge going forward. Also I have a nitpick around pulling code into its own function.

src/java/org/apache/cassandra/service/opstate/KeyspaceColumnFamilyKey.java
inespot: I found the word `Table` to be confusing and misleading since Cassandra code refers to them as `ColumnFamily`.

src/java/org/apache/cassandra/service/StorageService.java
inespot: Because this function is called from `Keyspace` and `cleanupState` is a private field

src/java/org/apache/cassandra/service/opstate/KeyspaceColumnFamilyKey.java
lyubent: This is a painful naming convention but as far as modern Cassandra is concerned Table == Column Family where the latter is the legacy name so we're going backwards with this rename - if you want more details check out [this Blog](https://medium.com/informatics/architecture-and-cassandra-data-model-9eb7a3cda969#85b6) on the internal naming conventions of the storage model. [CASSANDRA-14340](https://issues.apache.org/jira/browse/CASSANDRA-14340) exists to address the rest of the codebase talking about CFs instead of Tables. As for our fork - It's a bad idea to rename things even if their naming conventions are not the best - it means that if we want to merge a newer branch or try rebasing that process will be trickier.

src/java/org/apache/cassandra/service/StorageService.java
lyubent: Makes sense - it's less invasive than modifying the existing function, so this change fine!

lyubent (approved):

PR #377, opened 2022-12-05
Add logs when node finishes bootstrapping

inespot:
Goal: facilitate debugging when bootstrap operation fails mid-way.


leonz (dismissed):

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -906,6 +906,7 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
             }
             else
             {
+                logger.info("Setting bootstrap state to IN_PROGRESS");
```
leonz: actually maybe for both of these we should change "Setting -> Set" and put the log line _after_ the setBootstrapState call.

inespot: Discussed offline

leonz (approved):

PR #378, opened 2022-12-08
Add restricted IP property

zpear:
Primarily for use in K8s if we're replacing a given IP and want to ensure no other node mistakenly bounces and takes that IP

### DIFF
--- a/src/java/org/apache/cassandra/service/StartupChecks.java
+++ b/src/java/org/apache/cassandra/service/StartupChecks.java
@@ -406,17 +406,17 @@ public class StartupChecks
     {
         public void execute() throws StartupException
         {
-            String avoidIp = System.getProperty("palantir_cassandra.avoid_ip", null);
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/service/StartupChecks.java
```diff
@@ -400,4 +401,25 @@ public void execute() throws StartupException
             }
         }
     };
+
+    public static final StartupCheck checkIp = new StartupCheck()
+    {
+        public void execute() throws StartupException
+        {
+            String restrictedIp = System.getProperty("palantir_cassandra.restricted_ip", null);
```
inespot, line 409: Where/who sets this system property?

zpear: we do, via cassandra-env.sh

leonz (approved):

PR #386, opened 2023-03-01
Fix an import in unit test

rhuffy:
I ran into issues running the unit tests locally because a couple tests import `com.sun.tools.javac.util.List` for the `List.of()` factory. This PR replaces those calls with `Arrays.asList()` from `java.util.Arrays`, which maintains compatibility with `java8`.


lyubent (approved):
LGTM +1.

PR #388, opened 2023-03-07
C* nodes refuse to start if is_new_cluster is still set to true after 4 days

inespot:
This adds protection against a failure to remove the is_new_cluster config after the cluster creation.


rhuffy:
Can you put your { on new lines to match C* code style?

rhuffy (approved):

PR #390, opened 2023-03-07
Fixes for Java 11 and Java 17

rhuffy:
Adds version checks on JVM args that work on Java 11 but not on 17.

With this change, `conf/cassandra-env.sh` is compatible with both Java 11 and Java 17.
`bin/cassandra.in.sh` requires Java 17


lyubent (approved):
LGTM

PR #391, opened 2023-03-16
Add is_new_cluster StorageService endpoint

inespot:
Exposes a StorageServiceMBean endpoint that returns the value of `palantir_cassandra.is_new_cluster` env variable if set, or false otherwise


lyubent (approved):
LGTM

PR #394, opened 2023-03-30
Restrict some cipher suites

zpear:
While these wouldn't be used unless the socket supported them and EncryptionOptions were explicitly configured to allow them, we should not allow them

### DIFF
--- a/src/java/org/apache/cassandra/security/SSLFactory.java
+++ b/src/java/org/apache/cassandra/security/SSLFactory.java
@@ -30,13 +30,12 @@ import java.util.Enumeration;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.FutureTask;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
 
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SNIHostName;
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/security/SSLFactory.java
```diff
@@ -181,15 +178,27 @@ public static SSLContext createSSLContext(EncryptionOptions options, boolean bui
...
+
+        String[] ret = Iterables.toArray(ldesired.stream()
+            .peek(suite -> {
+                if (EncryptionOptions.restricted_cipher_suites.contains(suite)) {
+                    logger.warn("Removing restricted cipher suite {}", suite);
+                } else if (!ssupported.contains(suite)) {
+                    logger.warn("Removing suite that isn't supported by the socket");
+                }
+            })
+                .filter(ssupported::contains)
```
inespot: Nitpick: indentation is weird

src/java/org/apache/cassandra/security/SSLFactory.java
```diff
@@ -181,15 +178,27 @@ public static SSLContext createSSLContext(EncryptionOptions options, boolean bui
...
         List<String> ldesired = Arrays.asList(desired);
         ImmutableSet<String> ssupported = ImmutableSet.copyOf(supported);
-        String[] ret = Iterables.toArray(Iterables.filter(ldesired, Predicates.in(ssupported)), String.class);
+
+        String[] ret = Iterables.toArray(ldesired.stream()
+            .peek(suite -> {
+                if (EncryptionOptions.restricted_cipher_suites.contains(suite)) {
+                    logger.warn("Removing restricted cipher suite {}", suite);
+                } else if (!ssupported.contains(suite)) {
+                    logger.warn("Removing suite that isn't supported by the socket");
```
inespot: Do we wanna log the suite name here as well?

inespot (approved):

PR #395, opened 2023-03-30
respect CASSANDRA_HOME when defined externally, like by CCM

rhuffy:
CCM clusters fail to start in DTests on our fork because CCM expects a CASSANDRA_HOME and CASSANDRA_CONF that are not in SERVICE_HOME.

This change fixes test startup. ~This could break something if any production environments define a bad value for CASSANDRA_HOME, relying on it to be overridden here.~ Paths set externally by tests will not be overridden if `USE_CUSTOM_CASSANDRA_DIRS` is set.

### DIFF
--- a/bin/cassandra.in.sh
+++ b/bin/cassandra.in.sh
@@ -16,7 +16,7 @@
 
 SERVICE_HOME=$(cd "$(dirname "$0")/../../" && pwd)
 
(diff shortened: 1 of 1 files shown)


lyubent:
Would it not be better to just set the needed env var in the dtest?

PR #397, opened 2023-04-17
do not bootstrap with an empty schema

rhuffy:
Adds an additional check before bootstrapping. This ensures that a node will not try to bootstrap if its schema is empty.

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -920,6 +920,11 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                 }
                 Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
             }
+            while (!isNewCluster() && Schema.instance.getVersion().equals(Schema.emptyVersion))
+            {
+                setMode(Mode.JOINING, "our schema is empty and we are joining an existing cluster, continuing to wait for schema", true);
+                Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
+            }
             // if our schema hasn't matched yet, keep sleeping until it does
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -920,6 +920,11 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
                 }
                 Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
             }
+            while (!isNewCluster() && Schema.instance.getVersion().equals(Schema.emptyVersion))
```
tpetracca, line 923: should we do the `!isNewCluster()` in a wrapping if statement seeing as how it will never update?

tpetracca (commented):
I wonder if we should have a sort of "TERMINATED_TO_PREVENT_BAD_STATE" or something "Mode", that sidecar/manager can then observe, report, and potential monitor for?

and then in this case instead of looping in the `while` for forever we bail after like a minute or something.

could then re-use said Mode in other codepaths in StorageService that we don't think our deployment models should ever reach/hit; @zpear and I started to discuss something similar today

zpear:
> I wonder if we should have a sort of "TERMINATED_TO_PREVENT_BAD_STATE" or something "Mode", that sidecar/manager can then observe, report, and potential monitor for?
> 
> and then in this case instead of looping in the `while` for forever we bail after like a minute or something.
> 
> could then re-use said Mode in other codepaths in StorageService that we don't think our deployment models should ever reach/hit; @zpear and I started to discuss something similar today

The type of mode you're talking about we can do easily already by adding a new non transient error (we already do this for bootstraps that start with data already on disk to avoid corruption). 

Separately, I just found https://issues.apache.org/jira/browse/CASSANDRA-15158 which is worth a read

tpetracca:
Screenshot 2023-04-17 at 5 23 58 PM

PR #400, opened 2023-04-20
Backport CASSANDRA-15158: Wait for schema agreement when bootstrapping

rhuffy:
This PR includes a backport of CASSANDRA-15158, and some related subsequent patches from the 3.0 branch that change MigrationCoordinator. See PDS-352582 for context.

Specifically, we cherry-pick the following:

- Main fix | a960d87e05f01000a758d8e7e5a58be57d13eb33 | https://issues.apache.org/jira/browse/CASSANDRA-15158
- fixed SCHEMA_DELAY to use getSchemaDelay and no longer convert it from seconds to millis (since its already millis) | 17ebee3186d1bfdee9a2b355cb8f139492d144e8 | https://issues.apache.org/jira/browse/CASSANDRA-15158
- Don't run schema pulls during decommission | 3edacd632c2f2c360ca1d931c6b4f6fd7326511f | https://issues.apache.org/jira/browse/CASSANDRA-16495
- Don't wait for migrations from removed nodes, mention flag to skip | 4164313f926b605a24395a8ca920e8bea0691204 | https://issues.apache.org/jira/browse/CASSANDRA-16577
- Logging improvement | f55b2fb1b3a8af03a3424abfe7afa2ac60ac9dff | https://issues.apache.org/jira/browse/CASSANDRA-18096


rhuffy:
This backport overrode two previous changes we made on this fork

- https://github.com/palantir/cassandra/pull/36
- https://github.com/palantir/cassandra/pull/95

We should evaluate if we need to re-implement these as well. I think #95 more likely to be something we still want.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -857,6 +875,36 @@ else if (shouldBootstrap())
         }
     }
 
+    public void waitForSchema(int delay)
+    {
+        // first sleep the delay to make sure we see all our peers
+        for (long i = 0; i < delay; i += 1000)
+        {
+            // if we see schema, we can proceed to the next check directly
+            if (!Schema.instance.isEmpty())
```
inespot, line 889: Thrift and native connections are only enabled on a node once the node finishes joining, correct? Or more generally, do we know for sure that in all cases, when we start a new node, the only existing keyspace will be the system keyspace (and therefore schemaVersion will match empty schema)?

src/java/org/apache/cassandra/service/MigrationCoordinator.java
```diff
@@ -0,0 +1,512 @@
...
+        if (!FBUtilities.getBroadcastAddress().equals(InetAddress.getLoopbackAddress()))
+            CassandraDaemon.waitForGossipToSettle();
+
+        WaitQueue.Signal signal = null;
+        try
+        {
+            synchronized (this)
+            {
+                List<WaitQueue.Signal> signalList = new ArrayList<>(versionInfo.size());
+                for (VersionInfo version : versionInfo.values())
```
inespot, line 484: Could there be a scenario where other nodes can't reach us within the schema delay wait time and we never populate the `versionInfo` map?

src/java/org/apache/cassandra/service/MigrationCoordinator.java
```diff
@@ -0,0 +1,512 @@
...
+    {
+        return FailureDetector.instance.isAlive(endpoint);
+    }
+
+    @VisibleForTesting
+    protected boolean shouldPullSchema(UUID version)
+    {
+        if (Schema.instance.getVersion() == null)
+        {
+            logger.debug("Not pulling schema for version {}, because local schama version is not known yet", version);
```
inespot, line 204: Typo, but not sure if we wanna fix that ...

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -764,6 +764,9 @@ public enum TransientError {
      */
     public boolean resumeBootstrap();
 
+    /** Returns a map of schema version -> list of endpoints reporting that version that we need schema updates for */
+    public Map<String, Set<InetAddress>> getOutstandingSchemaVersions();
```
inespot, line 768: Do we know what the usage is for this function?

src/java/org/apache/cassandra/service/MigrationManager.java
```diff
@@ -77,121 +69,6 @@ public void unregister(MigrationListener listener)
...
-                    EndpointState epState = Gossiper.instance.getEndpointStateForEndpoint(endpoint);
-                    if (epState == null)
-                    {
-                        logger.debug("epState vanished for {}, not submitting migration task", endpoint);
-                        return;
-                    }
-                    VersionedValue value = epState.getApplicationState(ApplicationState.SCHEMA);
-                    UUID currentVersion = UUID.fromString(value.value);
-
-                    if (!currentVersion.equals(theirVersion))
```
inespot, line 143: I _think_ that this check, which was introduced in this PR: https://github.com/palantir/cassandra/pull/95 is still relevant and should still be included after the refactor if we wanna keep the optimization it provided. Even though `MigrationCoordinator` functions are synchronized and the class only has one instance, you could still receive Gossip schema changes and make changes to the endpoint state map in parallel. Indeed, the update to the endpoint map happens before the MigrationCoordinator one. Code ref [here](https://github.com/palantir/cassandra/blob/164482859c6e6b370ad6c14974612c51142f9e3b/src/java/org/apache/cassandra/service/StorageService.java#L2134)

rhuffy: I agree that this is probably a behavior we still want to have, but since the whole codepath is refactored, the check should probably be implemented differently. Optimization is: 

- We've scheduled a schema pull in 60s for a particular remote schema version
- During the wait, the schema has changed, and we schedule another schema pull for a different remote schema version
- We now have two schema pull requests for the same endpoint, so we should only execute one.

src/java/org/apache/cassandra/service/MigrationCoordinator.java
```diff
@@ -0,0 +1,512 @@
...
+
+        // There is a chance that quite some time could have passed between now and the MM#maybeScheduleSchemaPull(),
+        // potentially enough for the endpoint node to restart - which is an issue if it does restart upgraded, with
+        // a higher major.
+        if (!shouldPullFromEndpoint(callback.endpoint))
+        {
+            logger.info("Skipped sending a migration request: node {} has a higher major version now.", callback.endpoint);
+            callback.fail();
+            return;
+        }
```
rhuffy, line 445: This check is similar to #95. If we want to include #95, the check should probably go here.

src/java/org/apache/cassandra/service/MigrationCoordinator.java
```diff
@@ -0,0 +1,512 @@
...
+        for (int i=0, isize=info.requestQueue.size(); i<isize; i++)
+        {
+            InetAddress endpoint = info.requestQueue.remove();
+            if (!info.endpoints.contains(endpoint))
+                continue;
+
+            if (shouldPullFromEndpoint(endpoint) && info.outstandingRequests.add(endpoint))
+            {
+                return scheduleSchemaPull(endpoint, info);
+            }
```
rhuffy, line 160: here, `scheduleSchemaPull` is only called if the endpoint is not already in `outstandingRequests`, so I believe this effectively does the same thing as #95

zpear:
Skimming through related issues, we probably want to think about these before we release:
* https://issues.apache.org/jira/browse/CASSANDRA-17883 -- this seems scary given K8s nodes switch + replace IPs
* https://issues.apache.org/jira/browse/CASSANDRA-17972

zpear:
Also https://issues.apache.org/jira/browse/CASSANDRA-16856

Looking through the history for MigrationCoordinator on 3.11 are there any changes we wouldn't want to take?
https://github.com/apache/cassandra/commits/cassandra-3.11/src/java/org/apache/cassandra/service/MigrationCoordinator.java

Some additional changes in MigrationManager:
https://github.com/apache/cassandra/commits/cassandra-3.11/src/java/org/apache/cassandra/service/MigrationManager.java

rhuffy:
> Also https://issues.apache.org/jira/browse/CASSANDRA-16856

This seems like a reasonable change to take

rhuffy:
> Skimming through related issues, we probably want to think about these before we release:
> 
> * https://issues.apache.org/jira/browse/CASSANDRA-17883 -- this seems scary given K8s nodes switch + replace IPs
> * https://issues.apache.org/jira/browse/CASSANDRA-17972

These issues specifically are caused by https://issues.apache.org/jira/browse/CASSANDRA-16692, which we are _not_ backporting at this time.

PR #403, opened 2023-05-04
Fix docker image for Cassandra 

Sam-Kramer:


docker/Dockerfile
```diff
@@ -8,10 +8,19 @@ ENV PATH $CASSANDRA_DIR/bin:$PATH
 
 # Untar our distribution
 ADD palantir-cassandra-*tgz /
+RUN mkdir $CASSANDRA_DIR/logs; \
+  mkdir -p $CASSANDRA_DIR/data/saved_caches; \
+  mkdir -p /var/lib/cassandra/persistent_settings; \
```
Sam-Kramer, line 13: This is super lame, not sure why we cannot just use the default directory path (which should be `$CASSANDRA_HOME/data`, and the folders being owned by the `cassandra` user). For whatever reason, on the new version, we're failing to create these directories, so I do wonder if this never worked. Opt'd instead just to create directories manually, which isn't the craziest idea (the C* 3.X image does something similar)

docker/Dockerfile
```diff
@@ -8,10 +8,19 @@ ENV PATH $CASSANDRA_DIR/bin:$PATH
...
+  chown -R cassandra:cassandra /var/lib/cassandra /var/data/tmp $CASSANDRA_DIR
 
-# We give the cassandra user permissions to create things in the Cassandra dir
-RUN ["sh", "-c", "chmod +rwx ${CASSANDRA_DIR}"]
-RUN ["sh", "-c", "chown cassandra:cassandra ${CASSANDRA_DIR}"]
+
+# Install java 17
+RUN apt-get update && apt-get -y install openjdk-17-jdk
+ENV JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64/
+ENV PATH=$JAVA_HOME:$PATH
```
rhuffy: Should we use corretto here? https://docs.aws.amazon.com/corretto/latest/corretto-17-ug/generic-linux-install.html

Sam-Kramer: Good question! Originally I tried to use Corretto, but it's not possible as they use alpine linux , which uses `musl` rather than `glibc` which is needed for some debugging purposes.

docker/logback.xml
```diff
@@ -0,0 +1,101 @@
...
+    <discardingThreshold>0</discardingThreshold>
+    <includeCallerData>true</includeCallerData>
+    <appender-ref ref="DEBUGLOG" />
+  </appender>
+
+  <!-- STDOUT console appender to stdout (INFO level) -->
+
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+      <level>DEBUG</level>
```
Sam-Kramer, line 79: this is the only diff vs the default. we enable debug logging simply to give us more logging, which probably makes sense as this image will strictly be used by tests.

docker/cassandra.yaml
```diff
@@ -1,916 +1,92 @@
-# Cassandra storage config YAML
-
-# NOTE:
-#   See http://wiki.apache.org/cassandra/StorageConfiguration for
-#   full explanations of configuration directives
-# /NOTE
-
```
Sam-Kramer, line 7: Removed a ton of cruft, this shouldn't be apart of the config on the docker container...

docker/Dockerfile
```diff
@@ -8,19 +8,29 @@ ENV PATH $CASSANDRA_DIR/bin:$PATH
...
 # Configs set up
 ENV CASSANDRA_CONFIG $CASSANDRA_DIR/conf/
 ENV CASSANDRA_YAML $CASSANDRA_CONFIG/cassandra.yaml
 ENV CASSANDRA_ENV $CASSANDRA_CONFIG/cassandra-env.sh
 COPY cassandra.yaml $CASSANDRA_YAML
 COPY cassandra-env.sh $CASSANDRA_ENV
+COPY logback.xml $CASSANDRA_CONFIG/logback.xml
 
-ENV _JAVA_OPTIONS=-Dcassandra.skip_wait_for_gossip_to_settle=0
+ENV _JAVA_OPTIONS="-Dcassandra.skip_wait_for_gossip_to_settle=0 -Dpalantir_cassandra.is_new_cluster=true"
```
rhuffy: why do we need to set is_new_cluster here if we weren't previously?

Sam-Kramer: The image before was completely broken :) That said, I forgot that we enable this behavior within our seed providers internally, but the OSS version if largely unaffected by this change. Happy to remove.

rhuffy (approved):

PR #407, opened 2023-05-24
Publish cassandra-clientutil-jar

rhuffy:
`org.apache.cassandra.tools` contains some useful classes for cassandra sidecar. This PR adds this package to clientutil and publishes clientutil separate jar.


rhuffy:
I've tested that the jar I expect is built with `publishToMavenLocal`, and that it contains the classes of `org.apache.cassandra.tools`

leonz (approved):

rhuffy:
I'm waiting to confirm that additional classes don't need to be added before merging.

PR #409, opened 2023-05-31
Add single_node_cluster flag seed check

inespot:
Node will refuse to start if `palantir_cassandra.single_node_cluster` is set to true and the seed list is not a singleton with only localhost in it.


lyubent (approved):
LGTM

lyubent (approved):
LGTM

PR #411, opened 2023-06-15
Avoid ByteBuffer copy when computing serialized size for rangeScanBytesRead metric

schlosna:
While reviewing some JFR profiles, I noticed that as part of computing the serialized size of composites for `rangeScanBytesRead` metrics, it is allocating new `ByteBuffer`. This is very expensive unnecessary allocation and causes significant GC pressure. We should consider allocation-free serialized size computation. 

https://github.com/palantir/cassandra/blob/aad2368b01e601c87020092d7cda87720dd4cd34/src/java/org/apache/cassandra/db/ColumnFamilyStore.java#L2569

https://github.com/palantir/cassandra/blob/aad2368b01e601c87020092d7cda87720dd4cd34/src/java/org/apache/cassandra/db/composites/AbstractCType.java#L387

image

### DIFF
--- a/src/java/org/apache/cassandra/db/composites/AbstractComposite.java
+++ b/src/java/org/apache/cassandra/db/composites/AbstractComposite.java
@@ -88,6 +88,21 @@ public abstract class AbstractComposite implements Composite
         return result;
     }
 
+    @Override
+    public int serializedSize()
+    {
+        int serializedSize = 0;
+        if (isStatic())
+        {
+            serializedSize += Short.BYTES;
+        }
+        for (int i = 0; i < size(); i++)
+        {
+            serializedSize += Short.BYTES + get(i).remaining() + 1;
+        }
+        return serializedSize;
+    }
+
     public int dataSize()
     {
         int size = 0;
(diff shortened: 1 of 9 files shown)


src/java/org/apache/cassandra/db/composites/AbstractComposite.java
```diff
@@ -88,6 +88,21 @@ public ByteBuffer toByteBuffer()
         return result;
     }
 
+    @Override
+    public int serializedSize()
```
carterkozak, line 93: Perhaps we could deduplicate this with the ByteBuffer.allocate pre-sizing above? I think the two are equivalent.
https://github.com/palantir/cassandra/blob/2c0e4564f8b533ae57fce879486e6614f8d8b725/src/java/org/apache/cassandra/db/composites/AbstractComposite.java#L72-L76

src/java/org/apache/cassandra/db/composites/Composite.java
```diff
@@ -73,6 +73,8 @@ public static EOC from(int eoc)
 
     public ByteBuffer toByteBuffer();
 
+    public int serializedSize();
```
carterkozak: I don't know the backcompat constraints -- if there may be arbitrary implementations of `Composite` that we're unaware of, we could provide an expensive default implementation which delegates to `toByteBuffer().remaining()`

src/java/org/apache/cassandra/db/composites/AbstractComposite.java
schlosna: good call

src/java/org/apache/cassandra/db/composites/Composite.java
schlosna: done

PR #412, opened 2023-06-15
Expose getRangesOwnedByEndpoint via StorageServiceMBean

inespot:


src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -128,6 +130,14 @@ public enum ProgressState
      */
     public List<String> getTokens(String endpoint) throws UnknownHostException;
 
+    /**
+     * Get all ranges an endpoint is responsible for (by keyspace)
+     * @param ep endpoint we are interested in.
+     * @param hostId host id used for endpoint <-> node mapping verification, if non-null.
+     * @return ranges for the specified endpoint, with format (startToken,endToken]
+     */
+    public List<String> getRangesOwnedByEndpoint(String keyspaceName, InetAddress ep, UUID hostId);
```
zpear, line 150: can we just overload this method instead of using a nullable parameter?

zpear (approved):

PR #414, opened 2023-06-29
Add WAITING_TO_FINISH_BOOTSTRAP OperationMode

rhuffy:
As a followup to #402, report WAITING_TO_FINISH_BOOTSTRAP as our operation mode while waiting for a finishBootstrap JMX.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204936313517379


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -190,7 +190,7 @@ public Collection<Range<Token>> getPrimaryRangesWithinDC(String keyspace)
     private double traceProbability = 0.0;
 
     @VisibleForTesting
-    static enum Mode { STARTING, NORMAL, JOINING, LEAVING, DECOMMISSIONED, MOVING, DRAINING, DRAINED, ZOMBIE, NON_TRANSIENT_ERROR, TRANSIENT_ERROR, WAITING_TO_BOOTSTRAP, DISABLED }
+    static enum Mode { STARTING, NORMAL, JOINING, LEAVING, DECOMMISSIONED, MOVING, DRAINING, DRAINED, ZOMBIE, NON_TRANSIENT_ERROR, TRANSIENT_ERROR, WAITING_TO_BOOTSTRAP, WAITING_TO_FINISH_BOOTSTRAP, DISABLED }
```
zpear, line 195: we should include this in all the places that have logic based on `JOINING`

also, can you validate that writes will continue to be forwarded to this node in the new operation mode?

rhuffy: confirmed that write behavior is not related to OperationMode

inespot: Also confirmed that all other `JONING` related logic is only for logging.

inespot (approved):

PR #418, opened 2023-07-13
Nagarwal/tracing v1

nimai-agarwal:


src/java/org/apache/cassandra/service/TraceLogSpanObserver.java
```diff
@@ -0,0 +1,82 @@
...
+    private String convert(TraceLogV1 trace) {
+        com.palantir.sls.spec.logging.Span span = trace.getSpan();
+       String annotations = span.getAnnotations().stream()
+                                 .map(this::convertAnnotation).collect(Collectors.joining(","));
+
+       return String.format("{\"type\":\"trace.1\",\"time\":\"%s\",\"unsafeParams\":{}," +
+       "\"span\":{\"traceId\":\"%s\",\"id\":\"%s\",\"name\":\"%s\"," +
+       "\"parentId\":\"%s\",\"timestamp\":%s,\"duration\":%s,\"annotations\":[%s]," +
+       "\"tags\":{}}}", trace.getTime(), span.getTraceId(), span.getId(), span.getName(), span.getParentId().orElse(""),
+                            span.getTimestamp(), span.getDuration(), annotations);
```
schlosna, line 73: Do we need to ensure any of these are encoded appropriately for json?

`String.format` can be slow as each invocation has to parse the format string, so we might want to consider just string concatenation, see https://www.javaspecialists.eu/archive/Issue294-String.format-3x-faster-in-Java-17.html


```suggestion
       return "{\"type\":\"trace.1\",\"time\":\"" + trace.getTime() + "\",\"unsafeParams\":{}," +
       "\"span\":{\"traceId\":\"" + span.getTraceId() + "\",\"id\":\"" + span.getId() + "\",\"name\":\"" + span.getName() + "\"," +
       "\"parentId\":\"" + span.getParentId().orElse("") + "\",\"timestamp\": " + span.getTimestamp() + ",\"duration\": " + span.getDuration() + ",\"annotations\":[" + annotations + "]," +
       "\"tags\":{}}}";
```

nimai-agarwal: I got ObjectMapper working so this is no longer needed

PR #420, opened 2023-07-14
Add diskused metric representing most full data_file_dir

lyubent:
Adds DiskUsed to the StorageMetrics metric family. This metric shows the busiest data directory listed in `data_file_directories` as a percentage ranging from 0 to 100.

Storage#DiskUsed

### DIFF
--- a/src/java/org/apache/cassandra/metrics/StorageMetrics.java
+++ b/src/java/org/apache/cassandra/metrics/StorageMetrics.java
@@ -30,17 +30,20 @@ public class StorageMetrics
 {
     private static final MetricNameFactory factory = new DefaultNameFactory("Storage");
 
+    /* Tracks the most filled data_file_directory */
+    static
+    {
+        Metrics.register(factory.createMetricName("DiskUsed"), new Gauge<Double>()
+        {
+            public Double getValue()
+            {
+                return Directories.getMaxPathToUtilization().getValue();
+            }
+        });
+    }
+
     public static final Counter load = Metrics.counter(factory.createMetricName("Load"));
     public static final Counter exceptions = Metrics.counter(factory.createMetricName("Exceptions"));
     public static final Counter totalHintsInProgress  = Metrics.counter(factory.createMetricName("TotalHintsInProgress"));
     public static final Counter totalHints = Metrics.counter(factory.createMetricName("TotalHints"));
-
-    /* Tracks the most filled data_file_directory */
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/metrics/StorageMetrics.java
```diff
@@ -32,4 +34,13 @@ public class StorageMetrics
     public static final Counter exceptions = Metrics.counter(factory.createMetricName("Exceptions"));
     public static final Counter totalHintsInProgress  = Metrics.counter(factory.createMetricName("TotalHintsInProgress"));
     public static final Counter totalHints = Metrics.counter(factory.createMetricName("TotalHints"));
+
+    /* Tracks the most filled data_file_directory */
+    public static final Gauge<Double> diskUsed = Metrics.register(factory.createMetricName("DiskUsed"), new Gauge<Double>()
```
inespot: This variable is not used anywhere, correct? If so, it might be cleaner to just register the metric inside a static block:
```
static {
     Metrics.register(...)
}
```

src/java/org/apache/cassandra/metrics/StorageMetrics.java
```diff
@@ -32,4 +34,13 @@ public class StorageMetrics
...
+    {
+        public Double getValue()
+        {
+            return Directories.getMaxPathToUtilization().getValue();
```
inespot: Do we need to register the metric after a specific time laps? Or do we expect this value to be available as soon as C* starts up

lyubent: Doesn't really matter either way. The underlying method checks disk usage using the Java API for doing so, which should be something that can happen instantly.

src/java/org/apache/cassandra/metrics/StorageMetrics.java
lyubent: Really hate the idea of putting this in a static block because its not used. There are plenty of places where it will improve logging insight so I'll add it there.

src/java/org/apache/cassandra/metrics/StorageMetrics.java
inespot: You're right, it should be fine. The `Directories` class populates the `dataDirectories` variable through a static block, and this variable is the one used to compute load

inespot (approved):

PR #421, opened 2023-07-27
Backport idle timeouts

wi11dey:
Backport https://github.com/apache/cassandra/commit/0240a46 from Cassandra 3 for added security: clusters can now optionally set `native_transport_idle_timeout_in_ms` to specify a timeout in ms after which idle `cqlsh` sessions should be logged out automatically

The only differences from the Cassandra 3 implementation is that it has been moved later in the pipeline due to netty changes (and as a result, the first `IdleStateEvent` fired counts from the beginning of the connection and does not factor in later read/writes, which is also handled by this implementation), and another test case has been added


wi11dey:
superseded by https://github.com/palantir/cassandra/pull/476

PR #422, opened 2023-07-31
Add config flag that would allow skip blocking for read repairs

Sam-Kramer:
We do not rely on monotonic reads, thus we should be ok with 'skipping' read repairs (i.e. not waiting for writes to acknowledged) during the read path.

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageProxy.java
+++ b/src/java/org/apache/cassandra/service/StorageProxy.java
@@ -99,6 +99,8 @@ public class StorageProxy implements StorageProxyMBean
 
     private static final double CONCURRENT_SUBREQUESTS_MARGIN = 0.10;
 
+    private static final boolean DISABLE_BLOCKING_READ_REPAIRS = Boolean.getBoolean("palantir_cassandra.disable_blocking_read_repairs");
+
     private StorageProxy() {}
 
     static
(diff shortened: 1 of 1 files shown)


lyubent:
-1 on this. Regular repairs would be needed if we're evicting components of Cassandra's anti-entropy.

Sam-Kramer:
> -1 on this. Regular repairs would be needed if we're evicting components of Cassandra's anti-entropy.

To clarify, we still perform the read repair, it just isn't blocking the request from returning. Range scans in Cassandra 2.x also do not perform read repairs, and this is simply extending that to the get API, although we de use `get` more than `getRanges`.

tpetracca:
Generally we use cassandra in a way that does not require repairs to execute to maintain correctness. I view read repairs as an optimization for our flows, not a driver of consistency (we rely on immediate consistency via atlasdb, not some form of eventual consistency in the database).

I think ideally for heavily overwritten cells, we wouldn't read repair at all. For instance take the `sequential.offsets` table. This thing is literally just a counter that gets updated wildly quickly. I would hate to see us waste compute & IO processing repairs against a cell that are immediately invalidated via another update.

@Sam-Kramer think we should consider the impact this change would have on *total work done by the cluster* as opposed to just "minimizing latency of a given read". Not saying we should or shouldn't go through with this, but its worth assessing.

PR #427, opened 2023-08-11
Match the libthrift version in lib to build.xml

carterkozak:
Note that there's a newer 0.18.1 release as well.
This libthrift release provides both `TFramedTransport` and `TFastFramedTransport`, I opted for the most similar option to the existing implementation. The latter option appears to reuse buffers which may be helpful, but I don't want to add risk unnecessarily.

cc @Sam-Kramer due to #381


carterkozak:
Closing in favor of #428

PR #428, opened 2023-08-11
Align the libthrift version in build.xml with `lib/`

carterkozak:
This is a safer alternative to #427

When I attempt to resolve published jars and run cassandra from the published classpath, I get errors due to abi breaks in newer libthrift versions.

cc @Sam-Kramer due to https://github.com/palantir/cassandra/pull/381

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -393,7 +393,7 @@
             <exclusion groupId="org.slf4j" artifactId="slf4j-log4j12"/>
           </dependency>
           <dependency groupId="org.yaml" artifactId="snakeyaml" version="2.1"/>
-          <dependency groupId="org.apache.thrift" artifactId="libthrift" version="0.17.0">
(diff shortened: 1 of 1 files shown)


tpetracca:
amusingly, the only impact this change had was to affect atlasdb's transitive version recommendation for libthrift (see https://github.com/palantir/atlasdb/pull/6421), which is in practice the only thing ensuring a bunch of our atlasdb clients actually have a libthrift version on their classpath that isn't riddled with CVEs

tpetracca:
atlasdb needs to enforce the recommendation itself before we merge this

carterkozak:
@tpetracca https://github.com/palantir/cassandra/pull/432 may be a better option, but it has been a long time since I touched thrift, so I'm not sure what the level of risk is.

PR #431, opened 2023-08-15
Bumping guava from 16.0 to 32.1.2-jre

nimai-agarwal:
In prep for switching cassandra to sls-logging, we need to bump guava to the version used by sls-logging.
Also adds usages of guava-compatibility-agent to allow tests to pass and usage in runtime.


build.xml
```diff
@@ -365,7 +365,7 @@
           <dependency groupId="org.xerial.snappy" artifactId="snappy-java" version="1.1.10.1"/>
           <dependency groupId="net.jpountz.lz4" artifactId="lz4" version="1.3.0"/>
           <dependency groupId="com.ning" artifactId="compress-lzf" version="0.8.4"/>
-          <dependency groupId="com.google.guava" artifactId="guava" version="16.0"/>
+          <dependency groupId="com.google.guava" artifactId="guava" version="31.1-jre"/>
```
carterkozak: Can we use 32.1.2-jre (latest)? Idea is that we’ll resolve up to that version from all our libraries, so it’s best to verify compatibility premerge here

nimai-agarwal: sls-cassandra is currently resolving to guava 31.1-jre. But I'm ok bumping this to latest

lib/guava-32.1.2-jre.jar
carterkozak, line 1: Please rename `lib/licenses/guava-16.0.txt` to `lib/licenses/guava-32.1.2-jre.txt` and create another copy for the new third-party failureaccess jar: `lib/licenses/failureaccess-1.0.1.txt`
I don't think it's necessary for the compat-agent since it's one of ours.

wi11dey (approved):
tested on imola, looks good other than the license stuff, only substitutions with updated guava equivalents and the addition of https://github.com/palantir/guava-compatibility-agent

rhuffy (approved):

nimai-agarwal:
comment to unblock bulldozer

PR #432, opened 2023-08-18
Upgrade to libthrift 0.17.0

Sam-Kramer:
We previously tried to do this upgrade, but didn't actually change the jar checked in to the repo. Cassandra 2.2 libraries are not managed by a dependency manager, the jars are just committed to the `libs` dir.


src/java/org/apache/cassandra/thrift/CustomTThreadPoolServer.java
```diff
@@ -199,14 +204,10 @@ public void run()
...
-                // down. this is necessary for graceful shutdown.  (but not sufficient,
-                // since process() can take arbitrarily long waiting for client input.
-                // See comments at the end of serve().)
-                while (!stopped && processor.process(inputProtocol, outputProtocol))
+
+                while (!stopped)
                 {
-                    inputProtocol = inputProtocolFactory_.getProtocol(inputTransport);
-                    outputProtocol = outputProtocolFactory_.getProtocol(outputTransport);
+                    processor.process(inputProtocol, outputProtocol);
```
Sam-Kramer, line 213: https://github.com/apache/thrift/commit/e20ab3e2371803e69b321f13116e04a926ca39be

src/java/org/apache/cassandra/thrift/TCustomNonblockingServerSocket.java
```diff
@@ -44,9 +44,9 @@ public TCustomNonblockingServerSocket(InetSocketAddress bindAddr, boolean keepAl
 
     @Override
     @SuppressWarnings("resource")
-    protected TNonblockingSocket acceptImpl() throws TTransportException
+    public TNonblockingSocket accept() throws TTransportException
```
Sam-Kramer, line 47: https://github.com/apache/thrift/commit/6e023df1ded255dda00eb4c041c201e66c8d1fbc#diff-a0898a449b63e30396cba929f3cc094646073af86f64730c65b03fef6259a970

Sam-Kramer: https://github.com/apache/thrift/commit/6e023df1ded255dda00eb4c041c201e66c8d1fbc#diff-7b0ac7ee46ceeeb0e705052b66af1eae5a4ecd9571e160d72c5a004732ea5cff

tpetracca:
All the changes here are strictly oriented around generated code right? Mostly want to make sure the bump can't have actually caused the wire format of the API to have changed...

Sam-Kramer:
> All the changes here are strictly oriented around generated code right? Mostly want to make sure the bump can't have actually caused the wire format of the API to have changed...

Yes, although we should test on a development stack to be absolutely certain. I don't expect a wire-format change, as then it would make bumping libthrift near impossible, but wouldn't be surprised.

I am also slightly worried about behaviors of some methods changing here, and the old generated code not utilizing them correctly.

src/java/org/apache/cassandra/thrift/CustomTThreadPoolServer.java
tpetracca: okay so `process` no longer returns a boolean and instead relies on exceptions

however the other change here is we're not updating the input/output protocol on every process which previously we were; I'm not quite sure if it's valid we're not doing it anymore (or why it was in the first place)

src/java/org/apache/cassandra/hadoop/ConfigHelper.java
```diff
@@ -233,9 +233,9 @@ private static String thriftToString(TBase object)
     {
         assert object != null;
         // this is so awful it's kind of cool!
-        TSerializer serializer = new TSerializer(new TBinaryProtocol.Factory());
```
inespot, line 236: Nit: we probably wanna move this useful comment inside the try statement as well

src/java/org/apache/cassandra/thrift/CustomTThreadPoolServer.java
inespot: Based on the thrift PR, I think we were already relying on exceptions for the `process` function. I don't see any try/catch blocks + the only time the function returns false is for the AsyncProcessor, which I don't believe we are using.

Regarding the protocol refresh change, what is the reason we are not keeping it the way it was before?

rhuffy:
This has been soaking on a dev stack via sls-cassandra 4.534.0-rc2 and appears to be working.

rhuffy (approved):

PR #435, opened 2023-08-25
Adding tracing around keyspace creation and schema updates

nimai-agarwal:


build.xml
```diff
@@ -450,6 +450,10 @@
           <dependency groupId="io.netty" artifactId="netty-all" version="4.0.44.Final" />
           <dependency groupId="com.google.code.findbugs" artifactId="jsr305" version="2.0.2" />
           <dependency groupId="com.clearspring.analytics" artifactId="stream" version="2.5.2" />
+          <dependency groupId="com.palantir.tracing" artifactId="tracing" version="4.23.0" />
+          <dependency groupId="com.palantir.tracing" artifactId="tracing-api" version="4.23.0" />
+          <dependency groupId="com.palantir.safe-logging" artifactId="preconditions" version="1.8.2" />
+          <dependency groupId="com.palantir.safe-logging" artifactId="safe-logging" version="1.8.2" />
```
carterkozak, line 456: I'm surprised we don't need the safe-logging:logger dependencies, iirc the `tracing` library uses it

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -23,6 +23,7 @@
 import java.util.*;
 
 import org.apache.commons.lang3.StringUtils;
+import com.palantir.tracing.CloseableTracer;
```
wi11dey: unused import ?

wi11dey (commented):
overall looks good! nit: could you change the braces format to be consistent with the C* codebase ? i.e. opening brace on newline

There are some places unrelated to the added `try`s as well that the opening brace was moved to the start of the control flow structure and it would be an unnecessary annoyance in diffs when trying to reconcile with upstream

wi11dey (approved):

PR #437, opened 2023-08-30
Add logging of incoming Thrift connections

wi11dey:


src/java/org/apache/cassandra/thrift/CustomTThreadPoolServer.java
```diff
@@ -75,16 +78,20 @@ public class CustomTThreadPoolServer extends TServer
     // Server options
     private final TThreadPoolServer.Args args;
 
-    //Track and Limit the number of connected clients
+    // Track and Limit the number of connected clients
```
lyubent, line 81: This will make rebasing upstream harder. Yes it doesn't fit our code style, but synthetic changes should be avoided in my opinion.

wi11dey: agree, should have mentioned this PR is for observability during our testing only and won't be merged (will close after we've root caused the issue)

src/java/org/apache/cassandra/thrift/CustomTThreadPoolServer.java
```diff
@@ -229,6 +242,9 @@ public void run()
                 if (socket != null)
                     ThriftSessionManager.instance.connectionComplete(socket);
 
+                InetSocketAddress clientAddressAndIp = (InetSocketAddress) ((TCustomSocket) client_).getSocket().getRemoteSocketAddress();
+                InetAddress clientAddress = clientAddressAndIp.getAddress();
+                connections.computeIfPresent(clientAddress, (_ip, count) -> count - 1);
```
carterkozak, line 247: I think this will leak memory unless we remove values once they reach zero.
Also worth using a wrapper around the count value otherwise we'll constantly box and unbox integers.

wi11dey:
was just for making RCs to aid in root causing

PR #446, opened 2023-10-24
Add truncateAll nodetool command

d-guo:


d-guo:
Tested that all tables are all truncated while schemas are left intact.

src/java/org/apache/cassandra/config/Schema.java
```diff
@@ -333,6 +333,14 @@ public List<String> getUserKeyspaces()
         return ImmutableList.copyOf(Sets.difference(getNonSystemKeyspacesSet(), replicatedSystemKeyspaceNames));
     }
 
+    /**
+     * @return collection of the user defined keyspaces excluding system_palantir
+     */
+    public List<String> getUserKeyspacesPalantir()
```
zpear: is there a name we can use that doesn't reference palantir? like getNonAdminKeyspaces or something? then we can also replace the equivalent in StorageService::areKeyspacesEmptyForBootstrap

zpear (approved):

PR #450, opened 2023-11-01
Refresh cache on InvalidMutation

rhuffy:
The logic that checks for invalid mutations relies on a cached map owner endpoints for tokens. In production, we've observed false positives for InvalidMutation errors when the cache out of sync. 

This PR refreshes that cache when an invalid mutation is detected, and checks again before throwing.

The cache will be refreshed at most once every 10 minutes.

### DIFF
--- a/src/java/com/palantir/cassandra/utils/MutationVerificationUtils.java
+++ b/src/java/com/palantir/cassandra/utils/MutationVerificationUtils.java
@@ -73,13 +73,19 @@ public class MutationVerificationUtils
             if (mutationIsInvalid(naturalEndpoints, pendingEndpoints))
             {
                 keyspace.metric.invalidMutations.inc();
-                logger.error("Cannot apply mutation as this host {} does not contain key {} in keyspace {}. Only hosts {} and {} do.",
+                logger.error("InvalidMutation! Cannot apply mutation as this host {} does not contain key {} in keyspace {}. Only hosts {} and {} do.",
                         FBUtilities.getBroadcastAddress(),
                         Hex.bytesToHex(mutation.key().array()),
                         mutation.getKeyspaceName(),
                         naturalEndpoints,
                         pendingEndpoints);
-                throw new RuntimeException("Cannot apply mutation as this host does not contain key.");
+                throw new RuntimeException("InvalidMutation! Cannot apply mutation as this host does not contain key.");
+            }
+            else
+            {
+                logger.warn("Ignoring InvalidMutation error detected using stale token ring cache. Error was originally detected for key {} in keyspace {}.",
+                        Hex.bytesToHex(mutation.key().array()),
+                        mutation.getKeyspaceName());
             }
 
             keyspace.metric.validMutations.inc();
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/locator/AbstractReplicationStrategy.java
```diff
@@ -94,6 +94,16 @@ public ArrayList<InetAddress> getCachedEndpoints(Token t)
...
+    public void clearCachedEndpoints()
+    {
+        synchronized (this)
+        {
+            logger.trace("clearing cached endpoints");
+            cachedEndpoints.clear();
+            lastInvalidatedVersion = tokenMetadata.getRingVersion();
+        }
+    }
+
```
rhuffy: do we need this addition to be synchronized, like the method above it?

zpear: need to think about it more. concurrency is tricky + semi-dangerous since we don't want to affect https://github.com/palantir/cassandra/pull/336 (which fixes race conditions that lead to corruption).

Relevant to the cache is this race fix, which we don't have. Though I don't really see how this would give us some ranges with 2 instead of 3 endpoints https://issues.apache.org/jira/browse/CASSANDRA-16673

zpear: more generally, I wonder if we can/should just call TokenMetadata#invalidateCachedRings(), which will end up resetting the endpoint cache and incrementing the ring version

rhuffy: Looking at where `cachedEndpoints` is queried, it's always safe for a `get` to return null, because that will force the code to re-compute. The double-check and lock from the method above seems to exist to prevent double-clearing the cache.

rhuffy: oh, invalidateCachedRings() is for sure what we should be calling instead. Good find

src/java/com/palantir/cassandra/utils/MutationVerificationUtils.java
```diff
@@ -38,30 +40,55 @@ public class MutationVerificationUtils
...
-                             naturalEndpoints,
-                             pendingEndpoints);
+                logger.error("Cannot apply mutation as this host {} does not contain key {} in keyspace {}. Only hosts {} and {} do.",
+                        FBUtilities.getBroadcastAddress(),
+                        Hex.bytesToHex(mutation.key().array()),
+                        mutation.getKeyspaceName(),
+                        naturalEndpoints,
+                        pendingEndpoints);
                 throw new RuntimeException("Cannot apply mutation as this host does not contain key.");
             }
```
zpear, line 89: can you log in an else here that clearing the cache worked

src/java/org/apache/cassandra/locator/AbstractReplicationStrategy.java
zpear: trying to think through how this could affect #336 -- all other invocations of invalidateCachedRings are called while we have taken out the token metadata read/write locks while this one will not. I think that is OK, b/c those invocations are trying to synchronize updates to multiple in-memory maps that are used as a source of truth while here we're just forcing the next call to recompute the cached info and only touching this map. Anything else you can think of?

rhuffy: I would think the lock changes in that PR are there because the topology is changing, and the cache is invalidated because the ring is changing.

Since we're just invalidating a cache I don't see why we'd need to lock.

The call we're making also just invalidates the cache, it doesn't clear the Map itself, so we don't have to worry about concurrent mutations of a collection.

zpear (approved):

PR #452, opened 2023-11-07
Fix logging in MutationVerificationUtils

rhuffy:
#450 introduced a bug where warnings were incorrectly logged on every mutation.

This PR corrects the logic to verify mutations once if the cache is fresh, and only emits a warning if refreshing the cache makes a previously invalid mutation appear valid.

Behavior should be as follows:
1. Check the mutation is valid
2. If valid, do nothing
3. If invalid
  a. If the cache is fresh, throw an exception
  b. If the cache is stale
     1. Refresh the cache
     2. Check the mutation is valid
     3. If valid, log a warning
     4. If invalid, throw an exception

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1309,7 +1309,7 @@
         <jvmarg value="-ea"/>
         <jvmarg value="-Djava.io.tmpdir=${tmp.dir}"/>
         <jvmarg value="-Dcassandra.debugrefcount=true"/>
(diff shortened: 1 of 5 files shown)


rhuffy:
look at this commit to skip the mostly whitespace changes 06830cb3

zpear:
looks good -- let's add a couple basic unit tests too

zpear (approved):

PR #455, opened 2023-11-20
Only insert hostnames if not present for internal broadcast addrs

zpear:
We discovered that two hostnames in a separate VPC can resolve to the same proxy IP and we were mistakenly swapping them, since we were tracking proxy/public IPs as well as private/broadcast addresses.

This PR rips out some direct IP-mapping logic, as it was quite convoluted, and focuses on hostname-related mapping only. 

We now only track internal IP / broadcast address -> hostname, and will only update the endpoint we are trying to hit if the InetAddress does not contain a hostname, or has a hostname + endpoint which is from the other VPC (and therefore in our map).

If there is a hostname + we do not have the IP in the map, it must be a "public" / proxy IP, so we should leave the InetAddress as-is.

example:

For a node in another VPC hostA with broadcast address 10.100.0.1 that resolves to 10.9.0.1, 10.9.0.2, and 10.9.0.3 in the other VPC, you could have the following:
* hostA/10.100.0.1 -> (DNS) hostA/10.9.0.(1,2,3)
* hostA/10.9.0.(1,2,3) -> no-op
* /10.100.0.1 -> look up in our map -> hostA/10.9.0.(1,2,3)

### DIFF
--- a/src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
+++ b/src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
@@ -114,7 +114,7 @@ public class CrossVpcIpMappingHandshaker
         }
         if (!resolved.contains(endpoint))
         {
-            logger.warn("DNS-resolved address different than provided endpoint. This should mean that the endpoint " +
+            logger.debug("DNS-resolved address different than provided endpoint. This should mean that the endpoint " +
                         "includes a VPC-internal IP. Swapping. provided: {} resolved: {}",
                          endpoint, resolved);
             return resolved.stream().findFirst().get();
--- a/src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
+++ b/src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
@@ -21,6 +21,7 @@ package com.palantir.cassandra.cvim;
 import java.net.InetAddress;
 import java.time.Duration;
 import java.net.UnknownHostException;
+import java.util.Arrays;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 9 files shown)


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1398,6 +1398,7 @@ public static void setCrossVpcInternodeCommunication(boolean setting)
         conf.cross_vpc_internode_communication_enabled = setting;
     }
 
+    @Deprecated
```
inespot, line 1401: Let's file a github issue to remove this entirely at some point?

src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingHandshaker.java
```diff
@@ -57,111 +58,77 @@ public class CrossVpcIpMappingHandshaker
...
             return endpoint;
         }
         InetAddressIp proposedAddress = new InetAddressIp(endpoint.getHostAddress());
-        if (DatabaseDescriptor.isCrossVpcHostnameSwappingEnabled() && ipHostnameMappings.containsKey(proposedAddress)) {
-            return maybeSwapHostname(endpoint);
-        }
-        if (DatabaseDescriptor.isCrossVpcIpSwappingEnabled() && privatePublicIpMappings.containsKey(proposedAddress)) {
-            return maybeSwapIp(endpoint);
+        if (DatabaseDescriptor.isCrossVpcHostnameSwappingEnabled() && privateIpToHostname.containsKey(proposedAddress)) {
+            return maybeInsertHostname(endpoint);
```
inespot, line 98: Nit: `{` should be on the next line

inespot: Same thing on a couple other places in this class

inespot (approved):

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
zpear: done https://github.com/palantir/cassandra/issues/456

PR #457, opened 2023-11-28
Ignore cross vpc messages sent to incorrect nodes

zpear:
If we send a Syn to hostA, but include hostB as the target in the cross vpc syn body, hostA would respond with its internal IP as if it were hostB's. This can result in an incorrect broadcast address -> hostname mapping across the cluster

### DIFF
--- a/test/unit/com/palantir/cassandra/cvim/CrossVpcIpMappingSynVerbHandlerTest.java
+++ b/test/unit/com/palantir/cassandra/cvim/CrossVpcIpMappingSynVerbHandlerTest.java
@@ -119,7 +119,7 @@ public class CrossVpcIpMappingSynVerbHandlerTest
     }
 
     @Test
-    public void doVerb_doesUpdateMappingWhenTargetDoesNotMatchBroadcastAddress() throws UnknownHostException
+    public void doVerb_doesNotUpdateMappingWhenTargetDoesNotMatchBroadcastAddress() throws UnknownHostException
     {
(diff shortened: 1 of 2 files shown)


test/unit/com/palantir/cassandra/cvim/CrossVpcIpMappingSynVerbHandlerTest.java
```diff
@@ -107,4 +108,25 @@ public void doVerb_doesNotReplyWhenCrossVpcComm_disabled() throws UnknownHostExc
...
+    public void doVerb_doesNotReplyWhenTargetDoesNotMatchBroadcastAddress() throws UnknownHostException
+    {
+        doReturn(new InetAddressHostname("other")).when(handler).getBroadcastHostname();
+        DatabaseDescriptor.setCrossVpcInternodeCommunication(true);
+        handler.doVerb(messageIn, 0);
+        verify(handler, never()).reply(any(), any());
+    }
+
+    @Test
+    public void doVerb_doesUpdateMappingWhenTargetDoesNotMatchBroadcastAddress() throws UnknownHostException
```
wi11dey: supposed to be
```suggestion
    public void doVerb_doesNotUpdateMappingWhenTargetDoesNotMatchBroadcastAddress() throws UnknownHostException
```
?

zpear: yep!

wi11dey (approved):

PR #458, opened 2023-12-12
Add multislicemultiget counter metric per CF

lyubent:
Tracks usage of `o.a.c.thrift.CassandraServer#multiget_multislice` custom thrift API.

### DIFF
--- a/src/java/org/apache/cassandra/thrift/CassandraServer.java
+++ b/src/java/org/apache/cassandra/thrift/CassandraServer.java
@@ -443,7 +443,7 @@ public class CassandraServer implements Cassandra.Iface
         {
             ClientState cState = state();
             String keyspace = cState.getKeyspace();
-            Keyspace.open(keyspace).getColumnFamilyStore(column_parent.column_family).metric.largePartitionsCompacted.inc();
+            Keyspace.open(keyspace).getColumnFamilyStore(column_parent.column_family).metric.multiGetMultiSliceRequestQueries.inc();
             markRequestMeter(keyspace);
             cState.hasColumnFamilyAccess(keyspace, column_parent.column_family, Permission.SELECT);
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/thrift/CassandraServer.java
```diff
@@ -443,6 +443,7 @@ public Map<ByteBuffer, List<List<ColumnOrSuperColumn>>> multiget_multislice(List
         {
             ClientState cState = state();
             String keyspace = cState.getKeyspace();
+            Keyspace.open(keyspace).getColumnFamilyStore(column_parent.column_family).metric.multiGetMultiSliceRequestQueries.inc();
             markRequestMeter(keyspace);
```
rhuffy, line 447: can we confirm that this Keyspace.open() call doesn't introduce additional overhead?

lyubent: The cost of the call is free once the KS is initialized (and initialization happens before we enable the thrift interface), and initialization happens once and only once. The initialization itself isn't cheap (runs through a sync block where the monitor only allows a single thread in). There is also a map lookup and a map put on the `NonBlockingHashMap` used by o.a.c.config.Schema.

[source](https://github.com/palantir/cassandra/blob/2bcaf4adf07309a1cc9c11da03f44136204d6cf6/src/java/org/apache/cassandra/db/Keyspace.java#L111-L131)

However, we should benchmark to sanity test given that this is in the critical read path.

lyubent: It's worth noting that there is some safety that is hidden behind an assert:

```java
public static Keyspace open(String keyspaceName)
{
    assert initialized || keyspaceName.equals(SystemKeyspace.NAME);
    return open(keyspaceName, Schema.instance, true); // calls the private Keyspace#open
}
```

PR #461, opened 2024-01-26
Block requests to join ring after failed bootstraps

zpear:
Had to mock reads from System keyspace b/c it was messing with some unit tests

zombie mode should be fine b/c it goes through join ring 

Validated enable interfaces endpoint throws. Also circumvented w/ nodetool and validated we don't set tokens as part of enabling gossip either:

```
./service/bin/nodetool enablegossip
error: Cannot set tokens for a non-bootstrapped node
-- StackTrace --
com.palantir.cassandra.db.BootstrappingSafetyException: Cannot set tokens for a non-bootstrapped node
        at org.apache.cassandra.service.StorageService.setGossipTokens(StorageService.java:252)
```

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -650,7 +650,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
         logger.info("Thrift API version: {}", cassandraConstants.VERSION);
         logger.info("CQL supported versions: {} (default: {})",
                     StringUtils.join(ClientState.getCQLSupportedVersion(), ","), ClientState.DEFAULT_CQL_VERSION);
-        isBootstrapMode = SystemKeyspace.getBootstrapState() == SystemKeyspace.BootstrapState.IN_PROGRESS;
+        isBootstrapMode = SystemKeyspace.bootstrapInProgress();
 
         initialized = true;
 
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -247,6 +247,10 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
 
     public void setGossipTokens(Collection<Token> tokens)
     {
+        if (DatabaseDescriptor.isAutoBootstrap() && !bootstrapComplete())
+        {
+            throw new BootstrappingSafetyException("Cannot set tokens for a non-bootstrapped node");
+        }
         List<Pair<ApplicationState, VersionedValue>> states = new ArrayList<Pair<ApplicationState, VersionedValue>>();
         states.add(Pair.create(ApplicationState.TOKENS, valueFactory.tokens(tokens)));
         states.add(Pair.create(ApplicationState.STATUS, valueFactory.normal(tokens)));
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -641,6 +650,7 @@ public synchronized void initServer(int delay) throws ConfigurationException
         logger.info("Thrift API version: {}", cassandraConstants.VERSION);
         logger.info("CQL supported versions: {} (default: {})",
                     StringUtils.join(ClientState.getCQLSupportedVersion(), ","), ClientState.DEFAULT_CQL_VERSION);
+        isBootstrapMode = SystemKeyspace.getBootstrapState() == SystemKeyspace.BootstrapState.IN_PROGRESS;
```
zpear: this is one of the critical lines: some assertions based on isBootstrapMode, like read verb handler: https://github.com/palantir/cassandra/blob/2bcaf4adf07309a1cc9c11da03f44136204d6cf6/src/java/org/apache/cassandra/db/ReadVerbHandler.java#L37 may not reject reads otherwise, as `isBootstrapMode` would be null instead of `true` on second startup

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -247,6 +247,10 @@ public void setTokens(Collection<Token> tokens)
 
     public void setGossipTokens(Collection<Token> tokens)
     {
+        if (DatabaseDescriptor.isAutoBootstrap() && !bootstrapComplete())
```
inespot, line 250: Do we ever expect auto_bootstrap to be set to false?

src/java/org/apache/cassandra/service/StorageService.java
inespot: Nit: there is an existing function for this check: `SystemKeyspace.bootstrapInProgress()`

test/unit/org/apache/cassandra/service/ActiveRepairServiceTest.java
```diff
@@ -90,6 +87,7 @@ public void prepare() throws Exception
 
         TokenMetadata tmd = StorageService.instance.getTokenMetadata();
         tmd.clearUnsafe();
+        SystemKeyspace.setBootstrapState(SystemKeyspace.BootstrapState.COMPLETED);
```
inespot, line 90: Why is this change needed for this and the MutationsVerificationUtils tests?

src/java/org/apache/cassandra/service/StorageService.java
zpear: we never set it, but I included here to be consistent w/ the rest of the codebase

test/unit/org/apache/cassandra/service/ActiveRepairServiceTest.java
zpear: I believe the test initialization requires it here and the other test suite w/ the new assertions

inespot: Because `setTokens` calls `setGossipTokens`, got it

inespot (approved):

PR #464, opened 2024-02-06
additional checks to prevent bootstrapping if a cleanup is in progress

rhuffy:
These checks are already implemented in cassandra-manager. Also adding here in cassandra as an additional safety measure.


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1632,6 +1641,7 @@ public void onFailure(Throwable e)
     @Override
     public void startBootstrap()
     {
+        Preconditions.checkState(!isCleanupRunning(), "Cleanup is running. Refusing to start bootstrap.");
```
inespot, line 1644: Should we move this to sidecar? We'll hit the NTE after we signal to start bootstrap anyway

PR #465, opened 2024-02-09
Avoid NPE on cross vpc handshake if broadcast address is not explicitly set

zpear:
FBUtilities#getBroadcastAddress falls back to local address if not set in config. DatabaseDescriptor just passes along null

### DIFF
--- a/src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingSynVerbHandler.java
+++ b/src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingSynVerbHandler.java
@@ -52,7 +52,8 @@ public class CrossVpcIpMappingSynVerbHandler implements IVerbHandler<CrossVpcIpM
         InetAddressIp sourceExternalIp = new InetAddressIp(InetAddress.getByName(sourceName.toString())
                                                                       .getHostAddress());
 
-        InetAddressIp targetInternalIp = new InetAddressIp(FBUtilities.getBroadcastAddress().getHostAddress());
+        InetAddress broadcastAddress = FBUtilities.getBroadcastAddress();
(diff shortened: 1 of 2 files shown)


src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingSynVerbHandler.java
```diff
@@ -70,7 +71,7 @@ public void doVerb(MessageIn<CrossVpcIpMappingSyn> message, int id) throws Unkno
         if (!proposedTargetName.equals(getBroadcastHostname()))
         {
             logger.warn("Received a cross VPC Syn intended for another host. Ignoring. Intended: {} Actual: {}",
-                        proposedTargetName, DatabaseDescriptor.getBroadcastAddress());
+                        proposedTargetName, broadcastAddress);
```
d-guo, line 74: instead of storing `broadcastAddress`, why not use `FBUtilities.getBroadcastAddress()` here too? avoids any potential races

d-guo (approved):

src/java/com/palantir/cassandra/cvim/CrossVpcIpMappingSynVerbHandler.java
zpear: broadcast address is set in cassandra.yaml + should be install config, so no race concern. If we only called it in a single place in this class to begin with (though I left the call in #getBroadcastHostname b/c we use it for unit tests), we may not have even had this issue

zpear: in general though even if the value could change, it could still be advantageous to lock in the value at the start of a method so it doesn't swap under the hood at a later point within a single execution (though again, the function we're using for mocking unit tests means that's not really the case here anyway)

PR #473, opened 2024-03-13
finishBootstrap check times out after 30 minutes as a BootstrapError

rhuffy:
When WAITING_TO_FINISH_BOOTSTRAP was added, we required operators to manually ACK the successfully bootstrapped node before it could join the ring. Now, this is done by our automation, but cases where the automated safety check report an error, Cassandra will remain in this state indefinitely. This could be unsafe if the node is added manually, so this PR adds a timeout to fail fast.

The expectation is that our automation will evaluate and trigger `finishBootstrap` immediately if the checks pass, and if the checks fail, waiting some time will not cause them to pass.

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -1042,7 +1042,9 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                 {
                     recordNonTransientError(NonTransientError.BOOTSTRAP_ERROR, ImmutableMap.of("bootstrapSafetyCheckFailed", "true"));
                     unsafeDisableNode();
-                    throw new BootstrappingSafetyException("Bootstrap safety check failed.");
+                    String message = "Finish bootstrap was not called within 30 minutes. Bootstrap safety check failed.";
+                    logger.error(message);
+                    throw new BootstrappingSafetyException(message);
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1037,7 +1037,13 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
             try
             {
                 setMode(Mode.WAITING_TO_FINISH_BOOTSTRAP, "Awaiting finish bootstrap call", true);
-                finishBootstrapCondition.await();
+                boolean timeoutExceeded = !finishBootstrapCondition.await(30, MINUTES);
+                if (timeoutExceeded)
+                {
+                    recordNonTransientError(NonTransientError.BOOTSTRAP_ERROR, ImmutableMap.of("bootstrapSafetyCheckFailed", "true"));
+                    unsafeDisableNode();
+                    throw new BootstrappingSafetyException("Bootstrap safety check failed.");
```
zpear: nit: because finishBootstrap was not called before timeout was exceeded

zpear: could also be blocked on something other than safety check failure

zpear (approved):

PR #477, opened 2024-03-22
Add safeguard to NTE node incase local pause leads to gossip timeout

lyubent:
When a node bootstraps, it is viewed as fat clients by the rest of the cluster. This means if a local pause prevents heart-beating for longer than [MAX_LOCAL_PAUSE_IN_NANOS](https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/gms/FailureDetector.java#L275C20-L275C44), the bootstrapping node will be evicted from gossip from the rest of the nodes in the cluster. This leads to a state where the bootstrapping node continues to receive streams of data in but  isn't propagating its gossip state.

The problem occurs when the node completes streaming. It will think that it should be part of the cluster and ACK reads / writes / range-scans. This leads to a split-brain where not all nodes in the cluster agree on ownership.

This patch sets the node to non-transient-error mode should if hit the gossip timeout and adds a check once streaming completes to verify no NTEs for bootstrap errors were recorded. If any were observed, the node will fail the bootstrap.

### DIFF
--- a/src/java/org/apache/cassandra/gms/FailureDetector.java
+++ b/src/java/org/apache/cassandra/gms/FailureDetector.java
@@ -54,6 +54,7 @@ public class FailureDetector implements IFailureDetector, FailureDetectorMBean
     private static final int DEBUG_PERCENTAGE = 80; // if the phi is larger than this percentage of the max, log a debug message
     private static final long DEFAULT_MAX_PAUSE = 5000L * 1000000L; // 5 seconds
     private static final long MAX_LOCAL_PAUSE_IN_NANOS = getMaxLocalPause();
+    private static final long BOOSTRAP_SAFEGUARD_PAUSE_IN_NANOS = Long.getLong("palantir_cassandra.boostrap_safeguard_pause_in_ms", 30000L) * 1000000L;
     private long lastInterpret = System.nanoTime();
     private long lastPause = 0L;
 
@@ -268,7 +269,7 @@ public class FailureDetector implements IFailureDetector, FailureDetectorMBean
 
     private void safeguardBootstrapTimeout()
     {
-        if (StorageService.instance.isJoining())
+        if (StorageService.instance.isJoiningOrWaitingToFinishBootstrap())
         {
             StorageService.instance.unsafeDisableNode();
             logger.error("Detected local pause longer than MAX_LOCAL_PAUSE_IN_NANOS {} whilst bootstrapping", MAX_LOCAL_PAUSE_IN_NANOS);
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/gms/FailureDetector.java
```diff
@@ -276,6 +292,7 @@ public void interpret(InetAddress ep)
         {
             logger.warn("Not marking nodes down due to local pause of {} > {}", diff, MAX_LOCAL_PAUSE_IN_NANOS);
             lastPause = now;
+            safeguardBootstrapTimeout();
```
zpear: I think as-is this might be too aggressive -- we'll be failing after 5s by default when we have wiggle-room until 30s. could also make it a separately configured timeout so we can mess with it after seeing how it performs

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1558,7 +1558,7 @@ private boolean bootstrap(final Collection<Token> tokens)
             bootstrapStream.get();
             isBootstrapMode = false;
             logger.info("Bootstrap streaming completed for tokens {}", tokens);
-            return true;
+            return nonTransientErrors.stream().anyMatch(nte -> nte.containsKey(NonTransientError.BOOTSTRAP_ERROR.toString()));
```
zpear: can be consistent w/ what we do elsewhere, like for commitlog corruption:
```
StorageService.instance.hasNonTransientError(StorageServiceMBean.NonTransientError.COMMIT_LOG_CORRUPTION)
```

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4541,6 +4541,11 @@ public boolean isStarting()
         return operationMode == Mode.STARTING;
     }
 
+    public boolean isJoining()
+    {
+        return operationMode == Mode.JOINING;
+    }
```
zpear: we should also guard during waiting_to_finish_bootstrap

lyubent: I'm debating with myself if we want to expose a separate function for this but given that we only care about joining and waiting_to_finish_bootstrap together within a single context outside of SS, I'll refactor the existing isJoining to isJoiningOrWaitingToFinishBootstrap

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4541,6 +4541,11 @@ public boolean isStarting()
         return operationMode == Mode.STARTING;
     }
 
+    public boolean isJoiningOrWaitingToFinishBootstrap()
+    {
+        return operationMode == Mode.JOINING || operationMode == Mode.WAITING_TO_FINISH_BOOTSTRAP;
```
zpear, line 4546: should we make operationMode volatile?

zpear: looks like mainline does: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/service/StorageService.java#L454

lyubent: Yeah it was changed in https://issues.apache.org/jira/browse/CASSANDRA-12509 to improve the shutdown procedure. I've pushed a change to make opmode volatile

zpear (approved):

PR #479, opened 2024-03-25
Do not gossip to unreachable endpoints that have been removed

inespot:
A race in Cassandra Gossiper code can cause an endpoint to be re-added to `unreachableEndpoints` after it was removed from the node's state and quarantined.

### DIFF
--- a/src/java/org/apache/cassandra/gms/Gossiper.java
+++ b/src/java/org/apache/cassandra/gms/Gossiper.java
@@ -673,7 +673,7 @@ public class Gossiper implements IFailureDetectionEventListener, GossiperMBean
             double prob = unreachableEndpointCount / (liveEndpointCount + 1);
             double randDbl = random.nextDouble();
             if (randDbl < prob)
-                // Theoritically justRemoved endpoints should not be added unreachableEndpoints
+                // Theoritically justRemovedEndpoints and unreachableEndpoints should have no intersections
                 // but the Gossiper is prone to races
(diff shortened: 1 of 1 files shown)


d-guo (approved):

src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -673,7 +673,10 @@ private void maybeGossipToUnreachableMember(MessageOut<GossipDigestSyn> message)
             double prob = unreachableEndpointCount / (liveEndpointCount + 1);
             double randDbl = random.nextDouble();
             if (randDbl < prob)
-                sendGossip(message, unreachableEndpoints.keySet());
+                // Theoritically justRemovedEndpoints and unreachableEndpoints should have no intersections
+                // but the Gossiper is prone to races
+                sendGossip(message, Sets.filter(unreachableEndpoints.keySet(),
```
d-guo, line 678: optional nit: can we do this instead
```
sendGossip(message, Sets.difference(unreachableEndpoints.keySet(), justRemovedEndpoints.keySet()));
```

inespot: Discussed offline, keeping filter as it's closer to pattern in later version of C*: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/gms/Gossiper.java#L807

PR #483, opened 2024-04-23
[CASSANDRA-15439] Add option to override the FatClient timeout for Bootstrapping nodes

rhuffy:
Backport of [CASSANDRA-15439](https://issues.apache.org/jira/browse/CASSANDRA-15439)

This allows us to override the timeout for node removal while a node is bootstrapping. We know that we never want a bootstrapping node that has been removed by other nodes while streaming to succeed bootstrapping, and have already implemented safeguards to prevent that node from joining the ring after completing streaming.

Setting this parameter to a higher value makes this error state less likely to occur. This change takes the default set in 5.0, which is 5 minutes. We could consider making this higher.

We also had success reducing the frequency of this issue by vertically upsizing nodes and throttling streaming.

### DIFF
--- a/src/java/org/apache/cassandra/gms/VersionedValue.java
+++ b/src/java/org/apache/cassandra/gms/VersionedValue.java
@@ -20,13 +20,11 @@ package org.apache.cassandra.gms;
 import java.io.*;
 
 import java.net.InetAddress;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.UUID;
+import java.util.*;
 
 import static java.nio.charset.StandardCharsets.ISO_8859_1;
 
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 
 import org.apache.cassandra.db.TypeSizes;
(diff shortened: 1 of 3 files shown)


rhuffy:
committed to trunk here: https://github.com/apache/cassandra/commit/057d082e00f7d10b8e9b127cfabd9b8cd228da3d

src/java/org/apache/cassandra/gms/VersionedValue.java
```diff
@@ -79,6 +81,8 @@ public class VersionedValue implements Comparable<VersionedValue>
 
     // values for ApplicationState.REMOVAL_COORDINATOR
     public final static String REMOVAL_COORDINATOR = "REMOVER";
+    public final static List<String> BOOTSTRAPPING_STATES = Arrays.asList(VersionedValue.STATUS_BOOTSTRAPPING,
```
zpear: nit: status to match the upstream patch?

zpear (approved):

src/java/org/apache/cassandra/gms/VersionedValue.java
rhuffy: fixed

PR #488, opened 2024-04-29
Don’t purge tombstones when streaming for a DC rebuild

wi11dey:


rhuffy:
nit can you format this properly for cassandra. See the pinned comment in #help-cassandra for how to do this.

src/java/org/apache/cassandra/db/compaction/CompactionController.java
```diff
@@ -188,13 +197,34 @@ public static Set<SSTableReader> getFullyExpiredSSTables(ColumnFamilyStore cfSto
...
+     * @param cfs
+     * @param compacting          we take the drop-candidates from this set, it is usually the sstables included in the compaction
+     * @param overlappingSSTables the sstables that overlap the ones in compacting.
+     * @param gcBefore
+     * @return
+     */
+    public static Set<SSTableReader> getFullyExpiredSSTables(ColumnFamilyStore cfs, Iterable<SSTableReader> compacting, Iterable<SSTableReader> overlappingSSTables, int gcBefore)
+    {
+        return new CompactionController(cfs, compacting, gcBefore, overlappingSSTables).getFullyExpiredSSTables();
+    }
```
zpear: switched just so we can use a mocked storageservice?

wonder if there's another way we can avoid this -- as it stands we're not using a try-with when creating this compaction controller which could be problematic

wi11dey: yeah

considered using MockedStatic, but that requires mockito 3.4.0+ and seemed like upgrading broke existing tests

the try-with is only used to close overlappingSSTablesRefs, which doesn't exist when this method is called directly since it takes overlappingSSTables as a parameter and doesn't calculate them -- the instance created from the package-private constructor is just a pojo

wi11dey: we can also rely on the internal integration tests for coverage, which would simplify this pr a lot

src/java/org/apache/cassandra/db/compaction/CompactionController.java
```diff
@@ -17,10 +17,12 @@
  */
 package org.apache.cassandra.db.compaction;
 
+import java.io.Closeable;
```
jkozlowski: Unused import?

src/java/org/apache/cassandra/db/compaction/CompactionController.java
```diff
@@ -100,78 +112,75 @@ void maybeRefreshOverlaps()
     private void refreshOverlaps()
```
jkozlowski, line 95: There's a LOT of duplication in these two methods, refreshOverlaps and maybeRefreshOverlaps

src/java/org/apache/cassandra/db/compaction/CompactionController.java
wi11dey: ah nice thanks, forgot there's no spotless in this repo

src/java/org/apache/cassandra/db/compaction/CompactionController.java
wi11dey: the duplicate if () {return} checks ? i think it was originally done like this in https://github.com/palantir/cassandra/pull/49 to match the existing NEVER_PURGE_TOMBSTONES check from upstream https://github.com/apache/cassandra/blob/cassandra-2.2/src/java/org/apache/cassandra/db/compaction/CompactionController.java#L95

refreshOverlaps is also called from the constructor, so i can guard the call with checks there and remove them from refreshOverlaps if you prefer ?

jkozlowski: or put them really in refreshOverlaps only? Or really have a separate method shouldRefreshOverlaps and refreshOverlaps, and then maybeRefresh is:

```
if (shouldRefreshOverlaps()) {
  refreshOverlaps();
} 
```

src/java/org/apache/cassandra/db/compaction/CompactionController.java
zpear: If you combine this PR w/ the other PR for streams, you could make a static function like `areTombstonesSafeToPurge()` which checks pendingRanges, rebuilds, and for receiving streams. Then you could probably spoof a receiving stream using StreamManager#registerReceiving to force areTombstonesSafeToPurge to be true and just test against that.

src/java/org/apache/cassandra/db/compaction/CompactionController.java
wi11dey: merged with @zpear's requested changes https://github.com/palantir/cassandra/pull/488#discussion_r1586361583

src/java/org/apache/cassandra/db/compaction/CompactionController.java
wi11dey: tried with StreamManager#registerReceiving, but it requires a StreamResultFuture and checks it's planId field, which is hard to mock without changes. instead of changing StreamManager, i just added a unit-testing-only method unsafeSetRebuilding to control isRebuilding in StorageService

precedent for these unit-testing-only methods in StorageService here https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/service/StorageService.java#L632 and here https://github.com/palantir/cassandra/pull/262

src/java/org/apache/cassandra/db/compaction/CompactionController.java
```diff
@@ -275,10 +261,26 @@ public void close()
...
+     * If we are streaming, tombstones should not be purged so that we don't pre-maturely purge those that exceed gc_grace_seconds.
+     * <p>
+     * Note that this node may receive data without streaming, e.g. during a repair.
+     */
+    private static boolean doNotPurgeTombstones(String keyspace)
+    {
+        return NEVER_PURGE_TOMBSTONES
+                || StorageService.instance.isRebuilding()
+                || pendingRangesExistForKeyspace(keyspace)
+                || StreamManager.instance.getTotalReceivingStreams() > 0;
```
zpear: one worry I have is the new checks are going to limit all purging, rather than specifically for the keyspace that's streaming. I suppose we can start with this though and refactor if it becomes an issue

src/java/org/apache/cassandra/db/compaction/CompactionController.java
```diff
@@ -142,11 +130,11 @@ public static Set<SSTableReader> getFullyExpiredSSTables(ColumnFamilyStore cfSto
...
         logger.trace("Checking droppable sstables in {}", cfStore);
 
-        if (compacting == null || NEVER_PURGE_TOMBSTONES)
+        if (compacting == null)
             return Collections.<SSTableReader>emptySet();
 
-        if (pendingRangesExistForKeyspace(cfStore.keyspace.getName())) {
-            logger.debug("not looking for droppable sstables - there are pending ranges for keyspace {}", cfStore.keyspace.getName());
+        if (doNotPurgeTombstones(cfStore.keyspace.getName())) {
+            logger.debug("not looking for droppable sstables - doNotPurgeTombstones retunred true for keyspace {}", cfStore.keyspace.getName());
```
zpear: nit: typo in returned

src/java/org/apache/cassandra/db/compaction/CompactionController.java
```diff
@@ -76,17 +77,6 @@ public CompactionController(ColumnFamilyStore cfs, Set<SSTableReader> compacting
 
     void maybeRefreshOverlaps()
     {
-        if (NEVER_PURGE_TOMBSTONES)
-        {
-            logger.debug("not refreshing overlaps - running with -Dcassandra.never_purge_tombstones=true");
-            return;
```
zpear: I think we should still keep the early return here: https://issues.apache.org/jira/browse/CASSANDRA-11831

src/java/org/apache/cassandra/db/compaction/CompactionController.java
wi11dey: nevermind, this is intentional https://issues.apache.org/jira/browse/CASSANDRA-11831

src/java/org/apache/cassandra/db/compaction/CompactionController.java
wi11dey: we could start without the `StreamManager.instance.getTotalReceivingStreams() > 0`, which would keep the same keyspace-specific behavior for bootstraps ? and for rebuilding all the keyspaces are streamed (edit: for a DC migration) so i think it'd be the same either way there

zpear: you are able to rebuild one keyspace at a time, but yeah it's only tracked as a single boolean. for our purposes w/ how we do DC switches I think that's fine for now.

down to do your suggestion in this PR

zpear (approved):

PR #489, opened 2024-04-29
Don't purge tombstones when any streams are being received at all

wi11dey:
extension of https://github.com/palantir/cassandra/pull/488


wi11dey:
merged into https://github.com/palantir/cassandra/pull/489

PR #493, opened 2024-05-03
Don’t affect other tests in the rebuild tombstone purging test

wi11dey:
The test in https://github.com/palantir/cassandra/pull/488 may affect other tests depending on the order in which they're run


test/unit/org/apache/cassandra/db/compaction/CompactionControllerTest.java
```diff
@@ -206,6 +206,8 @@ public void testTombstonesNotPurgedWhenRebuilding()
         StorageService.instance.unsafeSetRebuilding(true);
 
         assertEquals(CompactionController.getFullyExpiredSSTables(cfs, compacting, overlapping, gcBefore), Collections.emptySet());
+
+        StorageService.instance.unsafeSetRebuilding(true);
```
zpear: should this be false? can also do a try finally so if this test fails it doesn't affect other ones then too

wi11dey: oops yes, changed to try finally too

zpear (approved):

PR #495, opened 2024-05-09
Verify keys on read

rhuffy:
Similar to the motivation for our verify_keys_on_write feature, it appears to be possible for a node to respond to a ReadVerb even if it does not think it owns the data. This PR adds an invalidReads metric behind a similar feature flag so we can start to see if this is actually occurring.

This would cause logical corruption if a client were to make a read request on a coordinator with an out of data view of the ring and received a null after committing a write, and could potentially result in resurrected data if an invalid read is written as part of a read repair.

This PR also adds SafeLogging arguments to the existing `VerifyKeysOnWrite` errors.


src/java/org/apache/cassandra/db/ReadVerbHandler.java
```diff
@@ -49,6 +50,8 @@ public void doVerb(MessageIn<ReadCommand> message, int id)
                                                                       ReadResponse.serializer);
         Tracing.trace("Enqueuing response to {}", message.from);
         MessagingService.instance().sendReply(reply, id, message.from);
+
+        OwnershipVerificationUtils.verifyRead(keyspace, command.key);
     }
```
rhuffy, line 55: note that this verification is intentionally after enqueuing the reply in order to avoid introducing additional latency on the read path. We should consider moving it earlier and throwing an exception after understanding in what circumstances this violation is logged, if any.

src/java/com/palantir/cassandra/utils/OwnershipVerificationUtils.java
```diff
@@ -37,72 +39,75 @@
...
+        List<InetAddress> cachedNaturalEndpoints = StorageService.instance.getNaturalEndpoints(keyspaceName, tk);
+        Collection<InetAddress> pendingEndpoints = StorageService.instance.getTokenMetadata().pendingEndpointsFor(tk, keyspaceName);
 
-        if (mutationIsInvalid(cachedNaturalEndpoints, pendingEndpoints))
+        if (operationIsInvalid(cachedNaturalEndpoints, pendingEndpoints))
         {
             if (cacheWasRecentlyRefreshed())
             {
-                throwInvalidMutationException(mutation, keyspace, cachedNaturalEndpoints, pendingEndpoints);
+                handler.onViolation(keyspace, key, cachedNaturalEndpoints, pendingEndpoints);
```
d-guo, line 88: an invalid read won't throw an exception here so it will constantly trigger `refreshCache` on every read, which could potentially be problematic. similarly it will still call `handler.onValid(keyspace);` at the botton

i think we should define a `InvalidReadException` and handle it here and below without a refresh

rhuffy: good catch! To avoid depending on the presence of an exception for correct control flow, I just added some returns in the correct place.

src/java/com/palantir/cassandra/utils/MutationVerificationHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import java.net.InetAddress;
+import java.nio.ByteBuffer;
+import java.util.Collection;
+import java.util.List;
+
+public class MutationVerificationHandler implements OwnershipVerificationHandler
+{
+    private static final Logger logger = LoggerFactory.getLogger(OwnershipVerificationUtils.class);
+
+    public static final OwnershipVerificationHandler INSTANCE = new MutationVerificationHandler();
```
d-guo, line 39: why wouldn't this be of type `MutationVerificationHandler` and similarly for `ReadVerificationHandler`

d-guo (approved):

src/java/com/palantir/cassandra/utils/MutationVerificationHandler.java
rhuffy: discussed offline, I'm just following the `List l = new ArrayList();` pattern where you always use the most generic interface where possible.

PR #504, opened 2024-06-06
MVP for handling connection resets

jkozlowski:


src/java/org/apache/cassandra/net/CallbackInfo.java
```diff
@@ -18,12 +18,13 @@
...
 
 import java.net.InetAddress;
+import javax.annotation.Nullable;
 
 import org.apache.cassandra.io.IVersionedSerializer;
 
 /**
  * Encapsulates the callback information.
- * The ability to set the message is useful in cases for when a hint needs 
+ * The ability to set the message is useful in cases for when a hint needs
```
jkozlowski, line 27: Cannot stop intellij from changing this

src/java/org/apache/cassandra/net/OutboundTcpConnectionPool.java
```diff
@@ -46,13 +47,16 @@ public class OutboundTcpConnectionPool
...
     public final OutboundTcpConnection gossipMessages;
 
+    public final Consumer<OutboundTcpConnection> socketDisconnectListener;
+
     // pointer to the reset Address.
     private InetAddress resetEndpoint;
     private ConnectionMetrics metrics;
 
-    OutboundTcpConnectionPool(InetAddress remoteEp)
+    OutboundTcpConnectionPool(InetAddress remoteEp, Consumer<OutboundTcpConnection> socketDisconnectListener)
```
jkozlowski: Looking at this class, it actually has retries for messages that it gets IOException WHEN trying to buffer them to the socket, so technically we'll be overly eager here and just cancel these anyway (even though they'll be retried.

Handling this would probably be a bit more invasive than this

src/java/org/apache/cassandra/net/OutboundTcpConnection.java
```diff
@@ -387,6 +387,7 @@ private void disconnect()
             }
             out = null;
             socket = null;
+            poolReference.socketDisconnectListener.accept(this);
```
jkozlowski: Reading a bit more, this will be quite aggressive, but we really don't expect the resets to be happening too much, so cancelling overly aggressively, is possibly ok.

If not, we may need to track which requests have actually been sent vs are still in our queue, and maybe not cancel those we haven't sent.

PR #508, opened 2024-06-21
Compaction throttling based on disk usage

lyubent:
Alter the rate limit set from `compaction_throughput_mb_per_sec` depending on disk utilization.

### DIFF
--- a/src/java/org/apache/cassandra/db/Directories.java
+++ b/src/java/org/apache/cassandra/db/Directories.java
@@ -111,7 +111,7 @@ public class Directories
     private static final Logger logger = LoggerFactory.getLogger(Directories.class);
 
     private static final double MAX_COMPACTION_DISK_USAGE = System.getProperty("palantir_cassandra.max_compaction_disk_usage") == null
-                                                            ? 0.95
+                                                            ? 0.90
                                                             : Double.parseDouble(System.getProperty("palantir_cassandra.max_compaction_disk_usage"));
     private static final ReadWriteLock SNAPSHOT_LOCK = new ReentrantReadWriteLock();
     private static final int LOCK_TIMEOUT_SECONDS = 60;
--- a/src/java/org/apache/cassandra/db/compaction/AbstractCompactionStrategy.java
+++ b/src/java/org/apache/cassandra/db/compaction/AbstractCompactionStrategy.java
@@ -279,8 +279,6 @@ public abstract class AbstractCompactionStrategy
                 scanners.add(sstable.getScanner(range, CompactionThroughputThrottler.getRateLimiter(Pair.create(sstable.getKeyspaceName(),
                                                                                                                 sstable.getColumnFamilyName()))));
             }
-
-
         }
         catch (Throwable t)
         {
--- a/src/java/org/apache/cassandra/db/compaction/AbstractCompactionStrategy.java
+++ b/src/java/org/apache/cassandra/db/compaction/AbstractCompactionStrategy.java
@@ -18,18 +18,15 @@
 package org.apache.cassandra.db.compaction;
 
 import java.util.*;
-import java.util.stream.Collectors;
 
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Iterables;
-import com.google.common.util.concurrent.RateLimiter;
-import org.apache.cassandra.io.sstable.format.SSTableReader;
-import org.apache.cassandra.utils.Pair;
+import com.palantir.cassandra.db.compaction.CompactionThroughputThrottler;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import org.apache.cassandra.db.ColumnFamilyStore;
 import org.apache.cassandra.db.Memtable;
 import org.apache.cassandra.db.lifecycle.LifecycleTransaction;
@@ -38,6 +35,8 @@ import org.apache.cassandra.dht.Token;
 import org.apache.cassandra.exceptions.ConfigurationException;
 import org.apache.cassandra.io.sstable.Component;
 import org.apache.cassandra.io.sstable.ISSTableScanner;
+import org.apache.cassandra.io.sstable.format.SSTableReader;
+import org.apache.cassandra.utils.Pair;
 import org.apache.cassandra.utils.JVMStabilityInspector;
 
 /**
--- a/src/java/org/apache/cassandra/db/compaction/AbstractCompactionStrategy.java
+++ b/src/java/org/apache/cassandra/db/compaction/AbstractCompactionStrategy.java
@@ -20,13 +20,11 @@ package org.apache.cassandra.db.compaction;
 import java.util.*;
 import java.util.stream.Collectors;
 
-import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Iterables;
 import com.google.common.util.concurrent.RateLimiter;
-import com.palantir.logsafe.SafeArg;
 import org.apache.cassandra.io.sstable.format.SSTableReader;
 import org.apache.cassandra.utils.Pair;
 import org.slf4j.Logger;
@@ -274,12 +272,16 @@ public abstract class AbstractCompactionStrategy
     @SuppressWarnings("resource")
     public ScannerList getScanners(Collection<SSTableReader> sstables, Range<Token> range)
     {
-        RateLimiter limiter = CompactionThroughputThrottler.getRateLimiter(sstablesToKsCfPair(sstables));
         ArrayList<ISSTableScanner> scanners = new ArrayList<ISSTableScanner>();
         try
         {
             for (SSTableReader sstable : sstables)
-                scanners.add(sstable.getScanner(range, limiter));
+            {
+                scanners.add(sstable.getScanner(range, CompactionThroughputThrottler.getRateLimiter(Pair.create(sstable.getKeyspaceName(),
+                                                                                                                sstable.getColumnFamilyName()))));
+            }
+
+
         }
         catch (Throwable t)
         {
@@ -537,16 +539,5 @@ public abstract class AbstractCompactionStrategy
         if (currGroup.size() != 0)
             groupedSSTables.add(currGroup);
         return groupedSSTables;
-    }
-
-    protected static Pair<String, String> sstablesToKsCfPair(Collection<SSTableReader> sstableReaders)
-    {
-        Set<Pair<String, String>> ksCfPairs = sstableReaders.stream()
-                                                            .map(sstable -> Pair.create(sstable.getKeyspaceName(), sstable.getColumnFamilyName()))
-                                                            .collect(Collectors.toSet());
-        Preconditions.checkArgument(ksCfPairs.size() == 1,
-                "An individual compaction should span a single keyspace/table pair {}",
-                SafeArg.of("keyspaceTablePair", ksCfPairs));
-        return ksCfPairs.stream().findFirst().get();
     }
 }
--- a/src/java/org/apache/cassandra/db/compaction/AbstractCompactionStrategy.java
+++ b/src/java/org/apache/cassandra/db/compaction/AbstractCompactionStrategy.java
@@ -18,13 +18,17 @@
 package org.apache.cassandra.db.compaction;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
+import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Iterables;
 import com.google.common.util.concurrent.RateLimiter;
+import com.palantir.logsafe.SafeArg;
 import org.apache.cassandra.io.sstable.format.SSTableReader;
+import org.apache.cassandra.utils.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -270,7 +274,7 @@ public abstract class AbstractCompactionStrategy
     @SuppressWarnings("resource")
     public ScannerList getScanners(Collection<SSTableReader> sstables, Range<Token> range)
     {
-        RateLimiter limiter = CompactionManager.instance.getRateLimiter();
+        RateLimiter limiter = CompactionThroughputThrottler.getRateLimiter(sstablesToKsCfPair(sstables));
         ArrayList<ISSTableScanner> scanners = new ArrayList<ISSTableScanner>();
         try
         {
@@ -533,5 +537,16 @@ public abstract class AbstractCompactionStrategy
         if (currGroup.size() != 0)
             groupedSSTables.add(currGroup);
         return groupedSSTables;
+    }
+
+    protected static Pair<String, String> sstablesToKsCfPair(Collection<SSTableReader> sstableReaders)
+    {
+        Set<Pair<String, String>> ksCfPairs = sstableReaders.stream()
+                                                            .map(sstable -> Pair.create(sstable.getKeyspaceName(), sstable.getColumnFamilyName()))
+                                                            .collect(Collectors.toSet());
+        Preconditions.checkArgument(ksCfPairs.size() == 1,
+                "An individual compaction should span a single keyspace/table pair {}",
+                SafeArg.of("keyspaceTablePair", ksCfPairs));
+        return ksCfPairs.stream().findFirst().get();
     }
 }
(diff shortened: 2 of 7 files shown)


src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -111,7 +111,7 @@ public class Directories
     private static final Logger logger = LoggerFactory.getLogger(Directories.class);
 
     private static final double MAX_COMPACTION_DISK_USAGE = System.getProperty("palantir_cassandra.max_compaction_disk_usage") == null
-                                                            ? 0.95
+                                                            ? 0.90
```
d-guo, line 114: what's the context for this change?

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public class CompactionThroughputThrottler
+{
+    private static final boolean ENABLE_COMPACTION_THROUGHPUT_THROTTLING = Boolean.getBoolean("palantir_cassandra.enable_compaction_throughput_throttling");
+    private static final long THROTTLER_TABLE_SIZE_BYTES = Long.getLong("palantir_cassandra.throttler_table_size_gb", 10L) * 1024L * 1024L * 1024L;
+    private static final Logger logger = LoggerFactory.getLogger(CompactionThroughputThrottler.class);
+
+    public static RateLimiter getRateLimiter(String keyspace, String columnFamily)
+    {
+        RateLimiter defaultThroughput = RateLimiter.create(CompactionManager.instance.getRateLimiter().getRate());
```
d-guo, line 22: nit: defaultRateLimiter

d-guo: nit: `defaultRateLimiter`

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,72 @@
...
+        RateLimiter recommendedThroughput = defaultThroughput;
+        if (diskUsage >= 0.90d)
+        {
+            // 12.5% throughput - at least 10MB/s
+            recommendedThroughput = RateLimiter.create(Math.max(defaultThroughput.getRate() / 8, 10d * 1024 * 1024));
+        }
+        else if (diskUsage >= 0.85d)
+        {
+            // 25% throughput - at least 20MB/s
+            recommendedThroughput = RateLimiter.create(Math.max(defaultThroughput.getRate() / 8, 20d * 1024 * 1024));
```
d-guo: `defaultThroughput.getRate() / 4`?

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+        else if (diskUsage >= 0.80d)
+        {
+            // 50% throughput - at least 40MB/s
+            recommendedThroughput = RateLimiter.create(Math.max(defaultThroughput.getRate() / 8, 40d * 1024 * 1024));
```
d-guo: same thing `defaultThroughput.getRate() / 2`?

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,72 @@
...
+            logger.trace("No disk pressure or table {} too small {} - using default compaction throughput {} with rate limit of {} for disk usage {}.",
+                         SafeArg.of("ksCfPair", String.format("%s/%s", keyspace, columnFamily)),
+                         SafeArg.of("liveSpaceUsedBytes", liveSpaceUsedBytes),
+                         SafeArg.of("defaultCompactionThroughputMbPerSec", StorageService.instance.getCompactionThroughputMbPerSec()),
+                         SafeArg.of("defaultThroughput", defaultThroughput),
+                         SafeArg.of("diskUsage", diskUsage));
+            return defaultThroughput;
+        }
+
+        RateLimiter recommendedThroughput = defaultThroughput;
```
d-guo: nit: `recommendedRateLimiter`

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,72 @@
...
+        else if (diskUsage >= 0.85d)
+        {
+            // 25% throughput - at least 20MB/s
+            recommendedThroughput = RateLimiter.create(Math.max(defaultThroughput.getRate() / 8, 20d * 1024 * 1024));
+        }
+        else if (diskUsage >= 0.80d)
+        {
+            // 50% throughput - at least 40MB/s
+            recommendedThroughput = RateLimiter.create(Math.max(defaultThroughput.getRate() / 8, 40d * 1024 * 1024));
+        }
```
d-guo: the default `compaction_throughput_mb_per_sec` is `16` unless we override this somewhere else. does this mean we're okay with actually increasing the compaction throughput to be greater than the configured throughput? seems like this isn't desirable behavior

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,72 @@
...
+
+        logger.debug("Due to disk usage of {} for table {}, recommending throughput of {}MBs with rate limit of {} for keyspace/table {}/{}. " +
+                     "Overwritten defaults were throughput {}MBs and rate limit  {}.",
```
d-guo: nit: extra space

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,72 @@
...
+                     SafeArg.of("ksCfPair", String.format("%s/%s", keyspace, columnFamily)),
+                     SafeArg.of("diskUsage", diskUsage),
+                     SafeArg.of("recommendedThroughputMbs", recommendedThroughput.getRate() / 1024/ 1024),
+                     SafeArg.of("recommendedRateLimiter", recommendedThroughput),
+                     SafeArg.of("defaultCompactionThroughput", StorageService.instance.getCompactionThroughputMbPerSec()),
+                     SafeArg.of("defaultRateLimiter", CompactionManager.instance.getRateLimiter()),
+                     SafeArg.of("keyspace", keyspace),
+                     SafeArg.of("table", columnFamily));
```
d-guo: i think the order of these arguments needs to be fixed

src/java/org/apache/cassandra/db/compaction/LeveledCompactionStrategy.java
```diff
@@ -372,6 +380,16 @@ protected OnDiskAtomIterator computeNext()
...
+                                                                         .collect(Collectors.toSet());
+            Preconditions.checkArgument(ksCfPairs.size() == 1,
+                                        "An individual compaction should span a single keyspace/table pair {}",
+                                        SafeArg.of("keyspaceTablePair", ksCfPairs));
+            return ksCfPairs.stream().findFirst().get();
+        }
+
         public void close() throws IOException
         {
             if (currentScanner != null)
```
d-guo: is there any reason you didn't also apply the changes to `AbstractCompactionStrategy`?

src/java/org/apache/cassandra/db/Directories.java
lyubent: This setting dictates how much of the available space we can use for a single compaction. 0.95 means that, at most, disk will reach 95% usage as the compaction takes place. The setting doesn't however account for backups via snapshots. Its effect is at the start of the compaction for scoping how many of the existing sstables for the specified ks.cf being compacted can be part of the compaction. If there are too many sstables to start the compaction and stay below 95% multiple passes will be done with some sstables being removed out of the initial pass.

The problem with 95% is, its too high. It will lead to disk filling up more often than not for clusters that are above 80% that have any sort of single big column family. This is mainly due to snapshotting taking place as the compaction is in flight.

src/java/org/apache/cassandra/db/compaction/LeveledCompactionStrategy.java
lyubent: We don't use STCS / DTCS / TWCS to any capacity other than L0 fallback, beyond that no. Happy to put it in if its desirable.

d-guo: how hard do you think it would be to add to the non-LCS ones (presumably just to `AbstractCompactionStrategy`)? i'd prefer to be consistent throughout and add them in this PR, but defer to you on what you want to do here

lyubent: Not hard. It _can_ be a one liner but I've refactored the `sstablesToKsCfPair` to live in ACS.

d-guo (approved):

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,71 @@
+package org.apache.cassandra.db.compaction;
+
```
rhuffy: could you move this to a package called `com.palantir.cassandra.db.compaction`. This makes it easier to track changes in the fork.

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,71 @@
...
+        long liveSpaceUsedBytes = ColumnFamilyStore.getIfExists(keyspace, columnFamily).metric.liveDiskSpaceUsed.getCount();
+
+        if (liveSpaceUsedBytes < THROTTLER_TABLE_SIZE_BYTES || diskUsage < 0.80d)
+        {
+            logger.trace("No disk pressure or table {} too small {} - using default compaction throughput {} with rate limit of {} for disk usage {}.",
+                         SafeArg.of("ksCfPair", String.format("%s/%s", keyspace, columnFamily)),
+                         SafeArg.of("liveSpaceUsedBytes", liveSpaceUsedBytes),
+                         SafeArg.of("defaultCompactionThroughputMbPerSec", StorageService.instance.getCompactionThroughputMbPerSec()),
+                         SafeArg.of("defaultRateLimit", defaultRateLimit),
+                         SafeArg.of("diskUsage", diskUsage));
```
rhuffy: can you wrap this in an "is tracing enabled" check?

This is to avoid needless allocations of SafeArg objects.

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,71 @@
...
+
+        logger.debug("Due to disk usage of {} and table size of {} for ks/cf {}, recommending throughput of {}MBs with rate limit of {}. " +
+                     "Overwritten defaults were throughput {}MBs and rate limit {}.",
+                     SafeArg.of("diskUsage", diskUsage),
+                     SafeArg.of("liveSpaceUsedBytes", liveSpaceUsedBytes),
+                     SafeArg.of("ksCfPair", String.format("%s/%s", keyspace, columnFamily)),
+                     SafeArg.of("recommendedRateLimitMbs", recommendedRateLimit.getRate() / 1024/ 1024),
+                     SafeArg.of("recommendedRateLimit", recommendedRateLimit),
+                     SafeArg.of("defaultCompactionThroughput", StorageService.instance.getCompactionThroughputMbPerSec()),
+                     SafeArg.of("defaultRateLimiter", CompactionManager.instance.getRateLimiter()));
```
rhuffy: Could this be wrapped in an "is debug enabled check"? Same reasoning as above to avoid allocations.

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,71 @@
...
+            recommendedRateLimit = RateLimiter.create(Math.min(defaultRateLimit.getRate() / 2, defaultRateLimit.getRate()));
+        }
+
+        logger.debug("Due to disk usage of {} and table size of {} for ks/cf {}, recommending throughput of {}MBs with rate limit of {}. " +
+                     "Overwritten defaults were throughput {}MBs and rate limit {}.",
+                     SafeArg.of("diskUsage", diskUsage),
+                     SafeArg.of("liveSpaceUsedBytes", liveSpaceUsedBytes),
+                     SafeArg.of("ksCfPair", String.format("%s/%s", keyspace, columnFamily)),
+                     SafeArg.of("recommendedRateLimitMbs", recommendedRateLimit.getRate() / 1024/ 1024),
+                     SafeArg.of("recommendedRateLimit", recommendedRateLimit),
```
rhuffy: could you just provide the value in MBps here?, and also fix the spacing of `1024/ 1024`?

src/java/org/apache/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -0,0 +1,71 @@
...
+        String keyspace = ksCf.left;
+        String columnFamily = ksCf.right;
+        RateLimiter defaultRateLimit = RateLimiter.create(CompactionManager.instance.getRateLimiter().getRate());
+        if (!ENABLE_COMPACTION_THROUGHPUT_THROTTLING)
+        {
+            return defaultRateLimit;
+        }
+
+        double diskUsage = Directories.getMaxPathToUtilization().getValue();
+        long liveSpaceUsedBytes = ColumnFamilyStore.getIfExists(keyspace, columnFamily).metric.liveDiskSpaceUsed.getCount();
```
rhuffy: What happens if the ksCf pair doesn't exist here? Do we need to add error handling?

lyubent: We're not getting this far if the ks / cf doesn't exist. Happy to add extra try / catch or a pre-condition but it feels redundant unless we're trying to protect against a race condition, i.e. we pass everything and its only here that the ks is removed.

lyubent: It will NPE

```
java.lang.NullPointerException
	at com.palantir.cassandra.db.compaction.CompactionThroughputThrottler.getRateLimiter(CompactionThroughputThrottler.java:34)
	at org.apache.cassandra.db.compaction.LeveledCompactionStrategy$LeveledScanner.<init>(LeveledCompactionStrategy.java:338)
	at org.apache.cassandra.db.compaction.LeveledCompactionStrategy.getScanners(LeveledCompactionStrategy.java:265)
	at org.apache.cassandra.db.compaction.WrappingCompactionStrategy.getScanners(WrappingCompactionStrategy.java:458)
	at org.apache.cassandra.db.compaction.AbstractCompactionStrategy.getScanners(AbstractCompactionStrategy.java:355)
	at org.apache.cassandra.db.compaction.CompactionTask.runMayThrow(CompactionTask.java:174)
	at org.apache.cassandra.utils.WrappedRunnable.run(WrappedRunnable.java:28)
	at org.apache.cassandra.db.compaction.CompactionTask.executeInternal(CompactionTask.java:76)
	at org.apache.cassandra.db.compaction.AbstractCompactionTask.execute(AbstractCompactionTask.java:59)
	at org.apache.cassandra.db.compaction.CompactionManager$10.runMayThrow(CompactionManager.java:746)
	at org.apache.cassandra.utils.WrappedRunnable.run(WrappedRunnable.java:28)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

Since this goes through [CompactionTask#runMayThrow](https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/db/compaction/CompactionTask.java#L102) we use a [try with resources](https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/db/compaction/CompactionTask.java#L173-L174) for the table refs - which will be auto-closed when we error out. It's better for us to propagate the exception than to try to handle it.

src/java/org/apache/cassandra/db/compaction/AbstractCompactionStrategy.java
```diff
@@ -270,12 +271,16 @@ public boolean apply(SSTableReader sstable)
...
         try
         {
             for (SSTableReader sstable : sstables)
-                scanners.add(sstable.getScanner(range, limiter));
+            {
+                scanners.add(sstable.getScanner(range, CompactionThroughputThrottler.getRateLimiter(Pair.create(sstable.getKeyspaceName(),
+                                                                                                                sstable.getColumnFamilyName()))));
+            }
+
+
```
zpear: nit whitespace

zpear:
lgtm

PR #509, opened 2024-06-21
Bump guava to 33.1.0

wi11dey:
In preparation for adding the debug server


wi11dey:
SHA1: 9b7ed39143d59e8eabcc6f91ffe4d23db2efe558
SHA256: 346aec0eb8c8987360c8a264e70ff10c2fba760446eb27e8ab07e78e787a75fe

d-guo:
have you checked that there's no breaking changes?

wi11dey:
yup, these are the changes: https://github.com/google/guava/releases

combined changelog:

- Changed Gradle Metadata to include dependency versions directly. This may address https://github.com/google/guava/issues/6657 that some users have reported (which might be a result of users' excluding guava-parent). (https://github.com/google/guava/commit/c6d35cf1a5617e96f41da0056dd2bc04f76ad0e5)
- collect: Changed Multisets.unmodifiableMultiset(set).removeIf(predicate) to throw an exception always, even if nothing matches predicate. (https://github.com/google/guava/commit/61dbccfda3f75872ba426ab9953d85bd76fa35a6)
- graph: Fixed the behavior of Graph/ValueGraph views for a node when that node is removed from the graph. (https://github.com/google/guava/commit/950799691c4efdbf9f1b6d6c384b613c22beef65)
- io: Fixed Files.createTempDir and FileBackedOutputStream under https://github.com/google/guava/issues/6634. (The fix actually covers only Java 9+ because Java 8 would require an additional approach. Let us know if you need support under Java 8.) (https://github.com/google/guava/commit/f87f68cd3e032a082a7702f3143862a9b27422fe)
- net: Made MediaType.parse allow and skip over whitespace around the / and = separator tokens in addition to the ; separator, for which it was already being allowed. (https://github.com/google/guava/commit/2786f83291c9d481750c874316f3588e9e3537ff)
-  util.concurrent: Tweaked Futures.getChecked constructor-selection behavior: The method continues to prefer to call constructors with a String parameter, but now it breaks ties based on whether the constructor has a Throwable parameter. Beyond that, the choice of constructor remains undefined. (For this and other reasons, we discourage the use of getChecked.) (https://github.com/google/guava/commit/59cfb2267aa0d8538b00874328943082186857d0)
- This version of guava-android contains some package-private methods whose signature includes the Java 8 Collector API. This is a test to identify any problems before we expose those methods publicly to users. Please report any problems that you encounter. (https://github.com/google/guava/commit/73dbf7ef26db8db2593056339be38bd00daf9584)
- Changed various classes to catch Exception instead of RuntimeException even when only RuntimeException is theoretically possible. This can help code that throws undeclared exceptions, as some bytecode rewriters (e.g., Robolectric) and languages (e.g., Kotlin) do. (https://github.com/google/guava/commit/c294c23760464b87ec193fc9b995e06f3d5e907f, https://github.com/google/guava/commit/747924eca675b79c3047d204814a4bc91e58fe70, https://github.com/google/guava/commit/b2baf48809bac380c23fb585635f917a594ec748)
- Added an Automatic-Module-Name to failureaccess, [Guava's one strong runtime dependency](https://github.com/google/guava/wiki/UseGuavaInYourBuild#what-about-guavas-own-dependencies). (https://github.com/google/guava/commit/280b5d2f606b5f37304d728311cf492d45f95843)
- reflect: In guava-android only, removed Invokable.getAnnotatedReturnType() and Parameter.getAnnotatedType(). These methods never worked in an Android VM, and to reflect that, they were born @Deprecated, @Beta, and @DoNotCall. They're now preventing us from rolling out some new Android compatibility testing. This is the only binary-incompatible change in this release, and it should have no effect in practice. Still, we bump the major version number to follow Semantic Versioning. (https://github.com/google/guava/commit/045cd8428fe34ad2e340407de9167dc9adf1801f)
- util.concurrent: Changed our implementations to avoid eagerly initializing loggers during class loading. This can help performance, especially under Android. (https://github.com/google/guava/commit/4fe1df56bd74e9eec8847bdb15c5be51f528e8c8)
- Updated our Error Prone dependency to 2.26.1, which includes a JPMS-ready jar of annotations. If you use the Error Prone annotations in a modular build of your own code, you may need to [add a requires line for them](https://github.com/google/error-prone/releases/tag/v2.26.1). (https://github.com/google/guava/commit/d48c6dfbb8cc9b8360ef5c14a5f4d00b802f14ee, https://github.com/google/guava/commit/c6e91c498ced26631029d1bdfdb9154d4a217368)
- base: Added a Duration overload for Suppliers.memoizeWithExpiration. (https://github.com/google/guava/commit/76e46ec35bbc220ba56d1e9a0abd853a7df51ece)
- base: Deprecated the remaining two overloads of Throwables.propagateIfPossible. They won't be deleted, but we recommend migrating off them. (https://github.com/google/guava/commit/cf86414a87e6e0d008f9438619e97eeefd22a32f)
- cache: Fixed a bug that could cause https://github.com/google/guava/pull/6851#issuecomment-1931276822. (https://github.com/google/guava/commit/0e1aebf73ec48848b17b402a22f33f55756073e8)
- graph: Changed the return types of transitiveClosure() and reachableNodes() to Immutable* types. reachableNodes() already returned an immutable object (even though that was not reflected in the declared return type); transitiveClosure() used to return a mutable object. The old signatures remain available, so this change does not break binary compatibility. (https://github.com/google/guava/commit/09e655f6c13bbd6644592552084597d6bce124f1)
- graph: Changed the behavior of views returned by graph accessor methods that take a graph element as input: They now throw IllegalStateException when that element is removed from the graph. (https://github.com/google/guava/commit/8dca77634125cbaf53365000d1c9f7fb32f6dcea)
- hash: Optimized Checksum-based hash functions for Java 9+. (https://github.com/google/guava/commit/afb35a5d1be4a9718376098ef991496d5548dc4d)
- testing: Exposed FakeTicker Duration methods to Android users. (https://github.com/google/guava/commit/f346bbb6a704bc365be576146c040abc296d15da)
- util.concurrent: Deprecated the constructors of UncheckedExecutionException and ExecutionError that don't accept a cause. We won't remove these constructors, but we recommend migrating off them, as users of those classes often assume that instances will contain a cause. (https://github.com/google/guava/commit/1bb3c4386bd5cc7a3cf4a6aa817aaeb2a179622d)
- util.concurrent: Improved the correctness of racy accesses for J2ObjC users. (https://github.com/google/guava/commit/d3232b71ce5eee2fb861a7f75bcebe1290afafdc)

also ensured Cassandra starts up fine after the bump; the rolling-JFR cluster on the dev stack currently has 33.1.0 on its classpath due to version constraint solving with the debug server

d-guo:
nice, taking a quick glance over the changes [here](https://github.com/google/guava/releases), i think they're all fine

d-guo (approved):

PR #512, opened 2024-07-02
Enable Registering Bootstrap Listener

d-guo:
Enable Registering Bootstrap Listener


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -114,6 +115,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     private int cleanupOpsInProgress = 0;
 
     private final RepairTracker repairTracker = new RepairTracker();
+    private final List<ProgressListener> bootstrapListeners = new CopyOnWriteArrayList<>();
```
rhuffy, line 118: why use a CopyOnWriteArrayList?

d-guo: thread-safe + follows existing usage for `lifecycleSubscribers`

rhuffy (approved):

PR #513, opened 2024-07-03
Add jmx endpoint to load new schema from disk

estherg77:
When performing a restore, this will allow us to load schema restored to system tables without restarting CassandraDaemon.


estherg77:
Closing in favor of #514

PR #516, opened 2024-07-17
Relax configuration thresholds

wi11dey:


inespot:
Keeping as a rc until we better flush out long term solution

PR #519, opened 2024-08-07
move SeedProvider and AbstractNetworkTopologySnitch to InetAddressAndPort for forward compatibility

rhuffy:
Modern versions of Cassandra use InetAdderssAndPort in place of InetAddress. To avoid maintaining two implementations of every custom SeedProvider and AbstractNetworkTopologySnitch we use internally, this PR switches some usage of InetAddress to InetAddressAndPort.

This allows us to maintain implementations of SeedProvider and AbstractNetworkTopologySnitch that are compile-time compatible with this fork, and with OSS trunk.

### DIFF
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -295,6 +295,8 @@ public class DatabaseDescriptor
     public static void applyConfig(Config config) throws ConfigurationException
     {
         conf = config;
+        
+        InetAddressAndPort.initializeDefaultPort(getSSLStoragePort());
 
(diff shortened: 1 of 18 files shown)


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -295,6 +295,8 @@ else if (config.rpc_interface != null)
     public static void applyConfig(Config config) throws ConfigurationException
     {
         conf = config;
+        
+        InetAddressAndPort.initializeDefaultPort(getSSLStoragePort());
```
rhuffy, line 299: note that on trunk the default port is `getStoragePort()` so 7000. I'm not sure if this matters.

PR #521, opened 2024-08-16
Persist isCompacted status of SSTables and scrub on startup

wi11dey:
When a table finished compaction as it is being streamed, the streaming will still hold a ref on the compaction ancestor. If the node crashes in this state, the next reboot will read both the ancestor and the compacted SSTable without any tombstones that arrived and got compacted during streaming.

### DIFF
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -608,13 +608,8 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
             Descriptor desc = sstableFiles.getKey();
             Set<Component> components = sstableFiles.getValue();
 
-            if (desc.type.isTemporary)
+            if (desc.type.isTemporary || components.contains(Component.OBSOLETE))
             {
-                SSTable.delete(desc, components);
-                continue;
-            }
-
-            if (components.contains(Component.IS_COMPACTED)) {
                 SSTable.delete(desc, components);
                 continue;
             }
@@ -843,7 +838,7 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
                 continue; // old (initialized) SSTable found, skipping
             if (descriptor.type.isTemporary) // in the process of being written
                 continue;
-            if (entry.getValue().contains(Component.IS_COMPACTED))
+            if (entry.getValue().contains(Component.OBSOLETE))
                 continue;
 
             if (!descriptor.isCompatible())
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -614,6 +614,11 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
                 continue;
             }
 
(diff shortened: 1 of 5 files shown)


src/java/org/apache/cassandra/io/sstable/format/SSTableReader.java
```diff
@@ -1691,7 +1691,7 @@ public boolean markObsolete(Tracker tracker)
         {
             assert !tidy.isReplaced;
         }
-        if (!tidy.global.isCompacted.getAndSet(true))
+        if (!tidy.global.setCompacted())
```
tpetracca, line 1694: note during review:
- this is the only place in the code `isCompacted` was getting set to true
- `setCompacted` is synchronized

thus the former atomic call to getAndSet is still mirrored in this change

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -838,6 +838,8 @@ public synchronized int loadNewSSTablesWithCount(boolean assumeCfIsEmpty)
                 continue; // old (initialized) SSTable found, skipping
             if (descriptor.type.isTemporary) // in the process of being written
                 continue;
+            if (entry.getValue().contains(Component.OBSOLETE))
+                continue;
```
tpetracca, line 848: is the implication here that by continuing, we're not initializing the sstable and we *will* delete it later? or do we need to update something elsewhere? what ensure "tmp" sstables get deleted?

tpetracca: also thoughts on if it's worth us logging (either by service log or metrics) some info about the fact that we've encountered an "obsolete" sstable (either here or later when we would delete it...) for transparenty/debuggability in the future

wi11dey: this section runs in parallel with the scrubDataDirectories code that does the actual deletion unfortunately https://github.com/palantir/cassandra/pull/521/files#diff-9db147d2e310e83ba17e4b4c756ce0aa90ce1742006fa3bc2418aaaf1fe4b7c9L573-L611 which is why i assume descriptor.type.isTemporary tables are also skipped here (they're also scrubbed in scrubDataDirectories). without this part, you get concurrent deletion issues like in these test failures: https://app.circleci.com/pipelines/github/palantir/cassandra/3931/workflows/00830712-dd20-49e5-a02f-0b2f70fb3429/jobs/11440/tests

wi11dey: added warn log when the scrubber encounters an obsolete sstable

tpetracca: 👍🏻

src/java/org/apache/cassandra/io/sstable/Component.java
```diff
@@ -55,6 +55,7 @@ public enum Type
         SUMMARY("Summary.db"),
         // table of contents, stores the list of all components for the sstable
         TOC("TOC.txt"),
+        OBSOLETE("Obsolete.txt"),
```
rhuffy: can you add a comment here to match the other documentation?

rhuffy (approved):

PR #522, opened 2024-08-16
Close client sockets on synchronous thrift server shutdown

zpear:
Previously shutdown would just close the server socket, but not the client sockets spawned via accept. This could result in client services blocking on reading from a socket which will never return a reply until some timeout gets exceeded.


src/java/org/apache/cassandra/thrift/ThriftSessionManager.java
```diff
@@ -82,4 +97,21 @@ public int getConnectedClients()
     {
         return activeSocketSessions.size();
     }
+
+    /**
+     * Currently only implemented for synchronous thrift server. This ensures that when the server socket is closed,
+     * spawned client sockets will also be closed instead of waiting for TCP timeouts.
+     */
+    public void closeActiveClients() throws IOException
```
nickbar01234, line 105: I might be missing some details here, so I was hoping you can provide some additional context.

My understanding is that the maximum number of active threads can be [quite large](https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/thrift/CustomTThreadPoolServer.java#L110). Can we potentially run into the following race:

1. The maximum number of possible threads are scheduled.
2. JVM schedule each thread one at a time, but before they register with `ThriftSessionManager`, they are context switched.
3. The client signals `stop()`, and we remove all "active clients", but they haven't been registered in (2).
4. Scheduler context switched to the threads in (2) and they start a connection with the remote client.

My assumption: Many threads can access `CustomTThreadPoolServer`.
* Should we have a check to prevent new clients from joining after `stop()` has been called?
* I'm not sure where exactly in the code we instantiate connection with the remote client the previous point isn't necessary.
* ~Additionally, does `client.close()` forcefully close the connection on the other side?~ I think your test case answers this question.

test/unit/org/apache/cassandra/thrift/CustomTThreadPoolServerTest.java
```diff
@@ -0,0 +1,72 @@
...
+{
+    @Test
+    public void stop_closesActiveSockets() throws TException, IOException
+    {
+        InetAddress ip = InetAddress.getLocalHost();
+        int port = DatabaseDescriptor.getRpcPort();
+        ThriftServer server = new ThriftServer(ip, port, 10);
+        server.start();
+
+        Cassandra.Client clientThatHasSentARequest = getClient(port);
```
nickbar01234, line 51: Which specific part of the code sends the request?

src/java/org/apache/cassandra/thrift/ThriftSessionManager.java
```diff
@@ -82,4 +97,21 @@ public int getConnectedClients()
...
+    {
+        Set<TTransport> currentActive = ImmutableSet.copyOf(activeClients);
+        for (TTransport client : currentActive)
```
nickbar01234: nit: My preference is to use functional style here.

src/java/org/apache/cassandra/thrift/ThriftSessionManager.java
zpear: I think most of that is possible. My opinion however is at the exact moment "closeActiveClients" is called those other non-registered connections aren't yet "active clients" (at least w/ my read of how these classes implicitly define active) b/c despite having a connection they aren't yet being processed. 

If you called stop, then another thread registered + tracked a client for an earlier created socket, line 218 in CustomTThreadPoolServer should prevent it from doing any work and then have the socket closed shortly after anyway

test/unit/org/apache/cassandra/thrift/CustomTThreadPoolServerTest.java
zpear: line 54

src/java/org/apache/cassandra/thrift/CustomTThreadPoolServer.java
```diff
@@ -198,6 +204,7 @@ public void run()
             TProtocol inputProtocol = null;
             TProtocol outputProtocol = null;
             SocketAddress socket = null;
+            ThriftSessionManager.instance.trackClient(client_);
```
inespot, line 207: Instead of tracking all clients in a central map, can we not just close the client socket "if stopped" in the `finally` ?

zpear: the problem is you can still have clients being processed for minutes after shutting down the server socket, so closing the client socket there would be too late

inespot (approved):

PR #524, opened 2024-08-20
In-situ tombstone metrics

wi11dey:
Testing marking metrics even on non-successful range scans

### DIFF
--- a/src/java/com/palantir/cassandra/utils/CountingCellIterator.java
+++ b/src/java/com/palantir/cassandra/utils/CountingCellIterator.java
@@ -20,6 +20,7 @@ import java.util.Iterator;
 
 import org.apache.cassandra.db.Cell;
 import org.apache.cassandra.db.ExpiringCell;
+import org.apache.cassandra.metrics.ColumnFamilyMetrics;
 import org.apache.cassandra.utils.MergeIterator;
 
 /**
@@ -38,6 +39,7 @@ import org.apache.cassandra.utils.MergeIterator;
 public class CountingCellIterator implements Iterator<Cell> {
     private final Iterator<Cell> delegate;
 
+    private final ColumnFamilyMetrics metrics;
     /*
      * Given the way SliceQueryFilter is passed these values, per ColumnFamilyStore.getColumnFamily(QueryFilter filter),
      * it is generally expected that gcBefore will be equivalent to { (now / 1000) - gc_grace_seconds }
@@ -50,12 +52,13 @@ public class CountingCellIterator implements Iterator<Cell> {
     private int liveCells = 0;
     private int tombstones = 0;
 
-    public static CountingCellIterator wrapIterator(Iterator<Cell> delegate, long timestamp, long gcBefore) {
-        return new CountingCellIterator(delegate, timestamp, gcBefore);
+    public static CountingCellIterator wrapIterator(Iterator<Cell> delegate, ColumnFamilyMetrics metrics, long timestamp, long gcBefore) {
+        return new CountingCellIterator(delegate, metrics, timestamp, gcBefore);
     }
 
-    private CountingCellIterator(Iterator<Cell> delegate, long timestamp, long gcBefore) {
+    private CountingCellIterator(Iterator<Cell> delegate, ColumnFamilyMetrics metrics, long timestamp, long gcBefore) {
         this.delegate = delegate;
+        this.metrics = metrics;
         this.gcBeforeSeconds = gcBefore;
         this.nowMillis = timestamp;
     }
@@ -103,6 +106,7 @@ public class CountingCellIterator implements Iterator<Cell> {
         } else if (ExpiringCell.class.isAssignableFrom(cell.getClass())) {
             droppableTtls++;
         } else if (cell.getLocalDeletionTime() < gcBeforeSeconds) {
+            metrics.droppableTombstones.mark();
             droppableTombstones++;
         } else {
             tombstones++;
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2018,6 +2018,10 @@ public ColumnFamily getColumnFamily(QueryFilter filter)
 
         ColumnFamily result = null;
 
+        if (filter.filter instanceof SliceQueryFilter) {
+            ((SliceQueryFilter) filter.filter).setMetrics(metric);
+        }
+
```
wi11dey, line 2024: this method of passing through the metrics is unfortunate; however, it is how metrics are currently recorded after-the-fact https://github.com/palantir/cassandra/pull/524/files#diff-9db147d2e310e83ba17e4b4c756ce0aa90ce1742006fa3bc2418aaaf1fe4b7c9R2082-R2084

zpear (dismissed):
for rc

d-guo:
can you post a grafana link here for the metric wherever you've been testing?

wi11dey:
https://pl.ntr/2oh

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2102,21 +2105,6 @@ private void recordMetrics(SliceQueryFilter filter) {
         metric.liveReadHistogram.update(filter.lastReadLive());
         metric.tombstonesReadHistogram.update(filter.lastReadTombstones());
```
d-guo, line 2098: right now, the droppableTombstones and ranged tombstone metrics will be updated in-situ, but these metrics will still be updated after the fact, right? any reason you don't move these into `CountingCellIterator ` and `ColumnCounter`?

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -276,10 +281,22 @@ public Comparator<Cell> getColumnComparator(CellNameType comparator)
 
     public void collectReducedColumns(ColumnFamily container, Iterator<Cell> reducedColumns, DecoratedKey key, int gcBefore, long now)
     {
-        reducedCells = CountingCellIterator.wrapIterator(reducedColumns, now, gcBefore);
+        reducedCells = CountingCellIterator.wrapIterator(reducedColumns, metrics, now, gcBefore);
         columnCounter = columnCounter(container.getComparator(), now);
-        deletionInfo = container.deletionInfo();
```
d-guo, line 281: how does `deletionInfo` get set here before we've itereated through the `reducedCells`?

src/java/com/palantir/cassandra/utils/CountingCellIterator.java
```diff
@@ -103,6 +106,9 @@ private void count(Cell cell) {
         } else if (ExpiringCell.class.isAssignableFrom(cell.getClass())) {
             droppableTtls++;
         } else if (cell.getLocalDeletionTime() < gcBeforeSeconds) {
+            if (metrics != null) {
+                metrics.droppableTombstones.mark();
```
d-guo, line 110: can you explain why this is correct to mark the metrics here and likewise in `SliceQueryFilter#collectReducedColumns`? previously, we set these at the level of `ColumnFamilyStore ` in `getColumnFamily`. are there other codepaths that use the filter that aren't from `getColumnFamily`?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
wi11dey: the histograms can only be updated after the fact since they need the total unlike the droppable tombstone rate

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
wi11dey: discussed offline

src/java/com/palantir/cassandra/utils/CountingCellIterator.java
wi11dey: discussed offline -- this is strictly an improvement as long as we're not double-counting. if do double-count someday, there is no risk of badness

wi11dey:
responding to offline discussion: the droppable tombstones metric we're marking in-situ here is registered as DroppableTombstonesRead over JMX https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java#L743, which we do indeed compact based on: https://pl.ntr/2oi https://pl.ntr/2oj

d-guo:
agree this is strictly an improvement so giving my +1. i am wondering though if in manager we should be checking the actual `droppableTombstones` instead of `droppableTombstonesRead`. it looks like cassandra already calculates this when computing the ratio
```
    private double getDroppableTombstoneRatio(boolean useGcGrace)
    {
        double allDroppable = 0;
        long allColumns = 0;
        int localTime = (int)(System.currentTimeMillis()/1000);

        for (SSTableReader sstable : getSSTables())
        {
            int gcBefore = localTime;
            if (useGcGrace) gcBefore = localTime - sstable.metadata.getGcGraceSeconds();
            allDroppable += sstable.getDroppableTombstonesBefore(gcBefore);
            allColumns += sstable.getEstimatedColumnCount().mean() * sstable.getEstimatedColumnCount().count();
        }
        return allColumns > 0 ? allDroppable / allColumns : 0;
    }
```
so it should be simple for us to just expose `allDroppable` and use that in manager

d-guo (dismissed):

d-guo (dismissed):

d-guo (approved):

PR #525, opened 2024-08-22
Revert "Persist isCompacted status of SSTables and scrub on startup"

rhuffy:
Reverts palantir/cassandra#521 in favor of #523


wi11dey (approved):
need to merge https://github.com/palantir/cassandra/pull/523

PR #526, opened 2024-08-22
Persist volume metadata

nickbar01234:
# Description

Write metadata to data drive and commit log on startup to ensure volume integrity.

## Details

image

## Additional Notes

I adapted the implementation for `KeyspaceTableOpStatePersister` ([link](https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java)) for `FileParser<T>`. In future iteration, we should move `KeyspaceTableOpStatePersister` to `com.palantir` package.


src/java/com/palantir/cassandra/actions/VolumeMetadata.java
```diff
@@ -0,0 +1,60 @@
...
+    public UUID getHostId()
+    {
+        return hostId;
+    }
+
+    public String getPodName()
+    {
+        return podName;
+    }
+}
```
wi11dey: discussed offline -- this likely needs @JsonCreator and @JsonValue to de/ser properly, and Wrapper<String> would not be appropriate since this is actually a tuple of two strings

src/java/com/palantir/cassandra/actions/Action.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.cassandra.actions;
+
+public interface Action
```
wi11dey: discussed offline: there is no reason to have the Action interface until we're actually doing something generically with them -- it's effectively the same as Runnable right now

src/java/com/palantir/cassandra/actions/DefaultVolumeIntegrityCheckActionWrapper.java
```diff
@@ -0,0 +1,55 @@
...
+        String commitLogDirectory = DatabaseDescriptor.getCommitLogLocation();
+
+        Action delegate = new DefaultVolumeIntegrityCheckAction(hostId, withParser(dataDirectory), withParser(commitLogDirectory));
+        delegate.execute();
+    }
+
+    private FileParser<VolumeMetadata> withParser(String path)
+    {
+        return new FileParser<>(Paths.get(path, METADATA_NAME), VolumeMetadata.class);
+    }
```
wi11dey: i would just make this a constructor or static initializer of DefaultVolumeIntegrityCheck

src/java/com/palantir/cassandra/actions/DefaultVolumeIntegrityCheckAction.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.UUID;
+
+import com.palantir.cassandra.utils.FileParser;
+import com.palantir.logsafe.Preconditions;
+
+/**
+ * Check if the backing volume has been tampered with. While Cassandra supports multiple data drive, we only
+ * use 1 data drive.
+ */
+public final class DefaultVolumeIntegrityCheckAction implements Action
```
wi11dey: nit

```suggestion
public final class VolumeIntegrityCheck
```

i don't see a reason to name it Default* when it's not implementing a VolumeIntegrityCheck interface and there's no way for there to be a non-default volume integrity check at the moment

src/java/com/palantir/cassandra/utils/FileParser.java
```diff
@@ -0,0 +1,77 @@
...
+    public void write(T value) throws IOException
+    {
+        String content = OBJECT_MAPPER.writeValueAsString(value);
+
+        String tmp = path.toFile().getAbsolutePath() + ".tmp";
+        File tmpFile = new File(tmp);
+
+        try
+        {
+            tmpFile.createNewFile();
```
wi11dey: is this line necessary ?

src/java/com/palantir/cassandra/utils/FileParser.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.nio.file.StandardOpenOption;
+import java.util.Optional;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class FileParser<T>
```
wi11dey, line 32: discussed offline -- https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java should inherit from this to reduce duplication

src/java/com/palantir/cassandra/actions/DefaultVolumeIntegrityCheckActionWrapper.java
nickbar01234: I went with a static initializer. Doing it as a constructor makes it more difficult for testing.

src/java/com/palantir/cassandra/utils/FileParser.java
nickbar01234: Addressed in #530!

src/java/com/palantir/cassandra/check/VolumeMetadata.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        VolumeMetadata volumeMetadata = (VolumeMetadata) other;
+        return hostId.equals(volumeMetadata.hostId) && podName.equals(volumeMetadata.podName);
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(hostId, podName);
+    }
```
inespot, line 84: Do we need the `toString` `equals` and `hashCode` overrides?

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private final UUID hostId;
+
+    private final FileParser<VolumeMetadata> dataDriveMetadataFileParser;
+
+    private final FileParser<VolumeMetadata> commitLogMetadataFileParser;
+
+    static class VolumeMetadataType extends TypeReference<VolumeMetadata>
+    {
+    }
```
inespot: Might be overkill, considering we seem to be only using this in one place

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+import com.palantir.logsafe.Preconditions;
+import org.apache.cassandra.config.DatabaseDescriptor;
+
+/**
+ * Check if the backing volume has been tampered with. While Cassandra supports multiple data drive, we only
+ * use 1 data drive.
+ */
+public final class VolumeIntegrityCheck
+{
+    public static final String METADATA_NAME = "cassandra-metadata.json";
```
inespot: Nit: `VOLUME_METADATA_FILE_NAME`

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.palantir.cassandra.utils.FileParser;
+import com.palantir.logsafe.Preconditions;
+import org.apache.cassandra.config.DatabaseDescriptor;
+
+/**
+ * Check if the backing volume has been tampered with. While Cassandra supports multiple data drive, we only
+ * use 1 data drive.
+ */
+public final class VolumeIntegrityCheck
```
inespot: Nit: `VolumesIntegrityCheck`

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private final FileParser<VolumeMetadata> dataDriveMetadataFileParser;
+
+    private final FileParser<VolumeMetadata> commitLogMetadataFileParser;
+
+    static class VolumeMetadataType extends TypeReference<VolumeMetadata>
+    {
+    }
+
+    public VolumeIntegrityCheck(UUID hostId,
```
inespot: This should probably be package-private and @VisibleForTesting

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private void writeIfEmpty(FileParser<VolumeMetadata> parser, boolean present) throws IOException
+    {
+        if (!present)
+        {
+            parser.write(VolumeMetadata.of(hostId));
+        }
+    }
+
+    public static VolumeIntegrityCheck of(UUID hostId)
```
inespot: Why go with a static factory method vs overloading the constructor?

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+    public void execute()
+    {
+        try
+        {
+            Optional<VolumeMetadata> maybeDataDriveMetadata = dataDriveMetadataFileParser.read();
+            Optional<VolumeMetadata> maybeCommitLogMetadata = commitLogMetadataFileParser.read();
+            check(maybeDataDriveMetadata, maybeCommitLogMetadata);
+
+            // Write order matters here. We treat write to the data drive as
+            // a "locking" operation, i.e the node is in a steady state.
```
inespot: What does "steady state" mean here?

src/java/com/palantir/cassandra/check/VolumeMetadata.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public UUID getHostId()
+    {
+        return hostId;
+    }
+
+    public String getPodName()
+    {
+        return podName;
+    }
```
inespot, line 62: Are the `getHostId()` and `getPodName()` function used anywhere?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -858,6 +860,7 @@ else if (shouldBootstrap())
             // (we won't be part of the storage ring though until we add a counterId to our state, below.)
             // Seed the host ID-to-endpoint map with our own ID.
             UUID localHostId = SystemKeyspace.getLocalHostId();
+            VolumeIntegrityCheck.of(localHostId).execute();
```
inespot: Not sure that logically, this should be part of the `prepareToJoin` function. We should move it up to `initServer`

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+                                     "Expected no metadata has been written to data drive if commitlog is empty");
+        }
+    }
+
+    private void writeIfEmpty(FileParser<VolumeMetadata> parser, boolean present) throws IOException
+    {
+        if (!present)
+        {
+            parser.write(VolumeMetadata.of(hostId));
+        }
```
inespot: When there is a single line in the if statement, you can drop the brackets, it makes the code less cluttered

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
```diff
@@ -34,127 +30,94 @@
...
         }
         catch (IOException e)
         {
-            return false;
+            logIoException("Failed to create state file.", persistentFileLocation.toFile().getAbsolutePath(), e);
         }
     }
 
-    private synchronized void atomicWritetoFile(File file, String content) throws IOException
+    private void logIoException(String message, String path, IOException e)
```
inespot: `log.warn` is actually shorter than `logIoException`? Any reason not to inline?

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
```diff
@@ -34,127 +30,94 @@
...
 
+    private final FileParser<Map<String, Long>> parser;
+
     public KeyspaceTableOpStatePersister(Path persistentFileLocation)
     {
         this.persistentFileLocation = persistentFileLocation;
-        this.persistentFile = getOrMaybeCreateStateFile();
+        this.parser = new FileParser<>(persistentFileLocation, new TypeReference<Map<String, Long>>()
+        {
+        });
```
inespot: Nit: inline bracket + parentheses

src/java/com/palantir/cassandra/utils/FileParser.java
```diff
@@ -0,0 +1,86 @@
...
+            Files.write(tmpFile.toPath(), content.getBytes(), StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING);
+            Files.move(tmpFile.toPath(), path.toFile().toPath(), StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
+        }
+        finally
+        {
+            Files.deleteIfExists(tmpFile.toPath());
+        }
+    }
+
+    public boolean create() throws IOException
```
inespot: Nit: `createFileIfNotExists`

inespot: Is this return value used anywhere?

src/java/com/palantir/cassandra/utils/FileParser.java
```diff
@@ -0,0 +1,86 @@
...
+            return Optional.empty();
+        }
+    }
+
+    public void write(T value) throws IOException
+    {
+        String content = OBJECT_MAPPER.writeValueAsString(value);
+
+        String tmp = path.toFile().getAbsolutePath() + ".tmp";
+        File tmpFile = new File(tmp);
```
inespot, line 64: Why the tmp file?

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
nickbar01234: I thought this would be more readable than trying to inline. What do you think?

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
nickbar01234: Mostly a stylistic preference; I prefer this over using `new ....`. I don't have a strong preference on this.

src/java/com/palantir/cassandra/check/VolumeMetadata.java
nickbar01234: ~It's not. Thank you for the catch!~

Edit: I realized that we actually need the getter for [proper serialization](https://www.baeldung.com/jackson-field-serializable-deserializable-or-not). There are a couple of alternatives:

1. Make all fields public (I don't think this ideal)
2. Expose a setter (I think it doesn't make sense since we would want to mark the field as final)
3. Make all fields serializable - I think it makes sense to set which fields is serializable at the POJO level

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
nickbar01234: This comes from the default formatter when I set up the repo. Is there a specific formatter that I should use?

src/java/org/apache/cassandra/service/StorageService.java
nickbar01234: At this point, we know for sure that the host ID exists locally. One concern for calling the check earlier is that we alter Cassandra's behavior, which may lead to unintended consequences.

src/java/com/palantir/cassandra/utils/FileParser.java
nickbar01234: This comes from the implementation for `KeyspaceTableOpStatePersister`. We write to a tmp file first to ensure atomicity.

src/java/com/palantir/cassandra/utils/FileParser.java
nickbar01234: It's not used anywhere currently, but I thought it maybe helpful to include since I would expect that future implementations can use this interface.

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
nickbar01234: Steady state means that the changes have been rolled onto existing nodes. I wanted to specify the distinction that a given node may be waiting to bootstrap vs a node that existed prior to this change. Without writing to data drive, we won't be able to distinguish between the two cases.

nickbar01234: Maybe I can include ^ blurb to make it more clear.

src/java/com/palantir/cassandra/check/VolumeMetadata.java
nickbar01234: Yep! 

* `toString` is mostly for utility / debugging, and I think it's a good practice to do on POJOs.
* `equals` is used in `VolumesIntegrityCheck`. I thought it was cleaner than using getters to manually compare each field.

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
nickbar01234: I intended for it to be an easy way to switch log level.

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
inespot: I am in favor of inlining when using one liners or one off logic. It makes the code more readable imo since you don't have to search for the type/function definition

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
inespot: Unless you think we'll soon need to modify log level, let's inline for now. We can always reconsider later

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
inespot: Ugh ... No the formatter is correct, this is following Cassandra style, it's just ugly imo

src/java/org/apache/cassandra/service/StorageService.java
inespot: Ah because `getLocalHostId` is what actually creates the local host id if it does not exists. Oki, sg

src/java/com/palantir/cassandra/utils/FileParser.java
inespot: Oki, fair enough

src/java/com/palantir/cassandra/check/VolumeIntegrityCheck.java
inespot: Oki, I guess we can default to what fits Cassandra's current code style best :)

test/unit/com/palantir/cassandra/check/VolumesIntegrityCheckTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    @BeforeClass
+    public static void beforeAll() throws IOException
+    {
+        Files.createDirectories(Paths.get(getDataDrive()));
+        Files.createDirectories(Paths.get(DatabaseDescriptor.getCommitLogLocation()));
+    }
+
+    @Before
+    public void beforeEach() throws IOException
```
inespot: Nit: `before()` and `beforeClass()` to make it consistent with the rest

src/java/com/palantir/cassandra/check/VolumesIntegrityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+            throw new RuntimeException("Fail to execute check for volume integrity", cause);
+        }
+    }
+
+    private void check(Optional<VolumeMetadata> maybeDataDriveMetadata, Optional<VolumeMetadata> maybeCommitLogMetadata)
+    {
+        if (maybeCommitLogMetadata.isPresent())
+        {
+            Preconditions.checkState(VolumeMetadata.of(hostId).equals(maybeCommitLogMetadata.get()),
+                                     "Expected SystemKeyspace HostId and POD_NAME to match with commitlog");
```
inespot, line 93: Remind me why we don't also check `!maybeDataDriveMetadata.isPresent() 
 || VolumeMetadata.of(hostId).equals(maybeDataDriveMetadata.get())` here? To prevent data pvcs from switching between two C* pods?

src/java/com/palantir/cassandra/utils/FileParser.java
```diff
@@ -0,0 +1,86 @@
...
+        {
+            return Optional.of(OBJECT_MAPPER.readValue(file, cls));
+        }
+        else
+        {
+            return Optional.empty();
+        }
+    }
+
+    public void write(T value) throws IOException
```
inespot, line 59: I think we should be good not making this synchronized since `CleanupStateTracker` functions are are synchronized and `VolumesIntegrityCheck` does writes sequentially but might be worth double checking

src/java/com/palantir/cassandra/check/VolumesIntegrityCheck.java
nickbar01234: If we swap data pvcs, then commit log metadata is present, so `Preconditions.checkState(VolumeMetadata.of(hostId).equals(maybeCommitLogMetadata.get())` throws.

src/java/com/palantir/cassandra/utils/FileParser.java
nickbar01234: From reading the code, I believe it should be fine. The only edge case I see is if multiple threads try to call `StorageService.init` which will hit `prepareToJoin` codepath multiple time. Is there any reason we should account for this?

src/java/com/palantir/cassandra/check/VolumesIntegrityCheck.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private void writeIfEmpty(FileParser<VolumeMetadata> parser, boolean present) throws IOException
+    {
+        if (!present) parser.write(VolumeMetadata.of(hostId));
+    }
+
+    public static VolumesIntegrityCheck getInstance(UUID hostId)
+    {
+        if (instance == null) instance = new VolumesIntegrityCheck(hostId);
```
inespot, line 109: Style nit: Eventhough you can remove brackets for "single line if", C* format still puts the action in a new line

src/java/com/palantir/cassandra/check/VolumesIntegrityCheck.java
```diff
@@ -0,0 +1,125 @@
...
+                                     "Expected no metadata has been written to data drive if commitlog is empty");
+        }
+    }
+
+    private void writeIfEmpty(FileParser<VolumeMetadata> parser, boolean present) throws IOException
+    {
+        if (!present) parser.write(VolumeMetadata.of(hostId));
+    }
+
+    public static VolumesIntegrityCheck getInstance(UUID hostId)
```
inespot, line 107: Not strongly opinionated but you could just set the instance as part of the execute function and save a function call

src/java/com/palantir/cassandra/check/VolumesIntegrityCheck.java
```diff
@@ -0,0 +1,125 @@
...
+    {
+        return new FileParser<>(Paths.get(path, VOLUME_METADATA_NAME), new TypeReference<VolumeMetadata>() {});
+    }
+
+    private static String getDataDirectory()
+    {
+        return Arrays
+               .stream(DatabaseDescriptor.getAllDataFileLocations())
+               .findFirst()
+               .orElseThrow(() -> new RuntimeException("No data directory found"));
```
inespot, line 123: I don't think anything currently prevents Cassandra to be configured with multiple directories on prem. Could we support multiple data dirs?

inespot:
Did you also test this on IL/Skylab managed setup?

nickbar01234:
Close since we're changing the solution logic.

PR #528, opened 2024-08-25
Backport CASSANDRA-16057: Pass output stream from probe

wi11dey:
So that it can be reliably redirected for nodetool commands to a different buffer, without intercepting any unwanted logs

Backport of https://github.com/apache/cassandra/commit/83e1e9e45193322f18f57aa7cc4ad31d9d5a152d

### DIFF
--- a/src/java/org/apache/cassandra/tools/NodeProbe.java
+++ b/src/java/org/apache/cassandra/tools/NodeProbe.java
@@ -53,7 +53,6 @@ import javax.management.remote.JMXConnectorFactory;
 import javax.management.remote.JMXServiceURL;
 import javax.rmi.ssl.SslRMIClientSocketFactory;
 
-import net.nicoulaj.compilecommand.annotations.Print;
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.db.ColumnFamilyStoreMBean;
 import org.apache.cassandra.db.HintedHandOffManager;
@@ -102,7 +101,6 @@ public class NodeProbe implements AutoCloseable
     final int port;
     private String username;
     private String password;
-    private final PrintStream output;
 
     protected JMXConnector jmxc;
     protected MBeanServerConnection mbeanServerConn;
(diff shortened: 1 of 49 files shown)


inespot (approved):
For rc

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -53,6 +53,7 @@
 import javax.management.remote.JMXServiceURL;
 import javax.rmi.ssl.SslRMIClientSocketFactory;
 
+import net.nicoulaj.compilecommand.annotations.Print;
```
rhuffy: unused import

rhuffy:
Why not backport everything instead of a partial backport? Excluding the files that don't exist of course. I think the only parts missing are the inner interface `NodeToolCmdRunnable` and `Output.java`

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -126,6 +128,11 @@ public class NodeProbe implements AutoCloseable
      * @throws IOException on connection failures
      */
     public NodeProbe(String host, int port, String username, String password) throws IOException
+    {
+        this(host, port, username, password, System.out);
+    }
+
+    public NodeProbe(String host, int port, String username, String password, PrintStream output) throws IOException
```
rhuffy: Can we avoid breaking the constructor API here? The oss patch gets around this with the `setOutput` method
https://github.com/apache/cassandra/commit/83e1e9e45193322f18f57aa7cc4ad31d9d5a152d#diff-92f8748902f03f37a4f7db56b4dfb7d226adcf3839141e6adb8ebbc575020d57R277

rhuffy (changes requested):
I would prefer if we just backported the patch you linked in the description, specifically to avoid adding a new NodeProbe constructor. I don't want to rely on a custom constructor as this would break forward compatibility in our internal repo.

rhuffy (approved):

PR #531, opened 2024-08-29
Expose Actual Tombstone Counts as Metric

d-guo:
We typically look at the read droppable tombstones metric to evaluate performance and compaction progress. This PR exposes Cassandra's internal estimate of droppable tombstones, regardless if they are being scanned. This is helpful for making sure compactions are keeping up as expected.

### DIFF
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -3429,7 +3429,8 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
         return getDroppableTombstoneCount(getSSTables(), useGcGrace);
     }
 
-    private double getDroppableTombstoneCount(Collection<SSTableReader> sstables, boolean useGcGrace) {
+    private double getDroppableTombstoneCount(Collection<SSTableReader> sstables, boolean useGcGrace)
+    {
         double allDroppable = 0;
         int localTime = (int) (System.currentTimeMillis() / 1000);
 
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -3413,6 +3413,35 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
         return allColumns > 0 ? allDroppable / allColumns : 0;
     }
 
+    public double getDroppableTombstoneCount()
+    {
+        return getDroppableTombstoneCount(true);
+    }
+
+    public double getTombstoneCount()
+    {
+        return getDroppableTombstoneCount(false);
+    }
+
+    public double getLiveTombstoneCount()
+    {
+        return getTombstoneCount() - getDroppableTombstoneCount();
+    }
+
+    private double getDroppableTombstoneCount(boolean useGcGrace)
+    {
+        double allDroppable = 0;
+        int localTime = (int) (System.currentTimeMillis() / 1000);
+
+        for (SSTableReader sstable : getSSTables())
+        {
+            int gcBefore = localTime;
+            if (useGcGrace) gcBefore = localTime - sstable.metadata.getGcGraceSeconds();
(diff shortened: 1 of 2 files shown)


rhuffy (dismissed):

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
```diff
@@ -737,6 +743,27 @@ public Double getValue()
...
+                return cfs.getLiveTombstoneCount();
+            }
+        });
+        tombstoneCount = createColumnFamilyGauge("TombstoneCount", new Gauge<Double>()
+        {
+            public Double getValue()
+            {
+                return cfs.getTombstoneCount();
+            }
+        });
```
wi11dey: these can all be method refs

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -3413,6 +3413,35 @@ private double getDroppableTombstoneRatio(boolean useGcGrace)
...
+    {
+        return getDroppableTombstoneCount(false);
+    }
+
+    public double getLiveTombstoneCount()
+    {
+        return getTombstoneCount() - getDroppableTombstoneCount();
+    }
+
+    private double getDroppableTombstoneCount(boolean useGcGrace)
```
wi11dey, line 3427: can you refactor `getDroppableTombstoneRatio` to use this instead for the numerator and only calculate the denominator itself ?

d-guo: actually i think it's more correct to calculate them together so they reference the same sstables instead of two separate calls to `getSSTables()`

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
d-guo: ah i wanted to stay consistent with the current style. i think it'd look weird to have `tombstoneCount = createColumnFamilyGauge("TombstoneCount", cfs::tombstoneCount);` here

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
wi11dey: in that case you can factor out a method that returns the ratio as a Pair and have the ratio do the division while the count just takes numerator; i don't want to be duplicating code that does actual calculation

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
wi11dey: intellij has a thing where you can just go through and convert the existing in the method for consistency; there's other method refs in the codebase so i think this is just a holdover

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -3399,18 +3399,47 @@ public double getLiveTombstoneRatio()
...
+    {
+        return getDroppableTombstoneCount(getSSTables(), useGcGrace);
+    }
+
+    private double getDroppableTombstoneCount(Collection<SSTableReader> sstables, boolean useGcGrace) {
```
wi11dey: nit to match c* style
```suggestion
    private double getDroppableTombstoneCount(Collection<SSTableReader> sstables, boolean useGcGrace)
    {
```

wi11dey (approved):

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
d-guo: to avoid code duplication and unnecessary calculation, decided to use helper which takes in a view of sstables

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
d-guo: converted just the 6 metrics that we added

PR #541, opened 2024-09-04
Fix resurrection by hanging ref and deletion by invalidated ancestry

wi11dey:
This PR
* Does not trust the ancestors of any stats file written before this version
  * A `VALID_ANCESTORS` component is added as a flag from this version onward
  * For older stats files, treated as no ancestors
* Skips SSTable renumbering when `assumeCfIsEmpty`
* When SSTables need to be renumbered, the stats file gets deleted
* Adds `palantir_cassandra.dry_run_ancestor_deletion` system parameter to only dry-run deleting leftover compaction ancestors.


d-guo (approved):
RC

PR #549, opened 2024-09-23
Expose CF Validation Hook and re-enable full ancestor cleanup

zpear:
Re-enables https://github.com/palantir/cassandra/pull/523 with configurable ancestor filtering

This also fixes a couple of other preexisting bugs related to compaction scheduling and cleanup

### DIFF
--- a/src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
+++ b/src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
@@ -33,7 +33,8 @@ public class ColumnFamilyStoreManager implements IColumnFamilyStoreValidator
     public static final ColumnFamilyStoreManager instance = new ColumnFamilyStoreManager();
     private final List<IColumnFamilyStoreValidator> validators;
 
-    private ColumnFamilyStoreManager() {
+    private ColumnFamilyStoreManager()
+    {
         this.validators = new CopyOnWriteArrayList<>();
     }
 
--- a/src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
+++ b/src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
@@ -43,7 +43,7 @@ public class ColumnFamilyStoreManager implements IColumnFamilyStoreValidator
         validators.add(validator);
     }
 
-    public void deregisterValidator(IColumnFamilyStoreValidator validator)
+    public void unregisterValidator(IColumnFamilyStoreValidator validator)
     {
         validators.remove(validator);
     }
--- a/src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
+++ b/src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
@@ -43,6 +43,11 @@ public class ColumnFamilyStoreManager implements IColumnFamilyStoreValidator
         validators.add(validator);
     }
 
+    public void deregisterValidator(IColumnFamilyStoreValidator validator)
+    {
+        validators.remove(validator);
+    }
+
     public Map<Descriptor, Set<Integer>> filterValidAncestors(CFMetaData cfMetaData, Map<Descriptor, Set<Integer>> sstableToCompletedAncestors, Map<Integer, UUID> unfinishedCompactions)
     {
         Map<Descriptor, Set<Integer>> filtered = sstableToCompletedAncestors;
--- a/src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
+++ b/src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
@@ -1,0 +1,54 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.cassandra.db;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import org.apache.cassandra.config.CFMetaData;
+import org.apache.cassandra.io.sstable.Descriptor;
+
+
+public class ColumnFamilyStoreManager implements IColumnFamilyStoreValidator
+{
+    public static final ColumnFamilyStoreManager instance = new ColumnFamilyStoreManager();
+    private final List<IColumnFamilyStoreValidator> validators;
+
+    private ColumnFamilyStoreManager() {
+        this.validators = new CopyOnWriteArrayList<>();
+    }
+
+    public void registerValidator(IColumnFamilyStoreValidator validator)
+    {
+        validators.add(validator);
+    }
+
+    public Map<Descriptor, Set<Integer>> filterValidAncestors(CFMetaData cfMetaData, Map<Descriptor, Set<Integer>> sstableToCompletedAncestors, Map<Integer, UUID> unfinishedCompactions)
+    {
+        Map<Descriptor, Set<Integer>> filtered = sstableToCompletedAncestors;
+        for (IColumnFamilyStoreValidator validator : validators)
+        {
+            filtered = validator.filterValidAncestors(cfMetaData, filtered, unfinishedCompactions);
+        }
+        return filtered;
+    }
+}
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -836,7 +836,7 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
     {
         if (assumeCfIsEmpty)
         {
-            throw new UnsupportedOperationException("Loading new SSTables is not supported on this version");
+            throw new UnsupportedOperationException("Loading new SSTables is not supported on version 2.2.18-1.164.0+.");
         }
         logger.info("Loading new SSTables for {}/{}{}...",
                 keyspace.getName(), name,
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -104,6 +104,9 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
 {
     private static final Logger logger = LoggerFactory.getLogger(ColumnFamilyStore.class);
 
+    private static final boolean DRY_RUN_NON_COMPACTING_UNUSED_SSTABLE_CLEANUP = Boolean.getBoolean(
+                                            "palantir_cassandra.dry_run_non_compacting_unused_sstable_cleanup");
+
     private static final ExecutorService flushExecutor = new JMXEnabledThreadPoolExecutor(DatabaseDescriptor.getFlushWriters(),
                                                                                           StageManager.KEEPALIVE,
                                                                                           TimeUnit.SECONDS,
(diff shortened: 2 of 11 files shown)


src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -710,7 +713,20 @@ public static void removeUnfinishedCompactionLeftovers(CFMetaData metadata, Map<
...
             }
+            sstableToAncestors.put(desc, ancestors);
+        }
+
+        sstableToAncestors = ColumnFamilyStoreManager.instance.filterValidAncestors(metadata, sstableToAncestors, unfinishedCompactions);
+
+        for (Map.Entry<Descriptor, Set<Component>> sstableFiles : directories.sstableLister().skipTemporary(true).list().entrySet())
+        {
+            Descriptor desc = sstableFiles.getKey();
+            if (!sstableToAncestors.containsKey(desc))
```
inespot: Instead of re-listing tables, why not loop over the `sstableToAncestors` result considering the `filterValidAncestors` already operates on the original `sstableToAncestors` map which is derived from `directories.sstableLister().skipTemporary(true).list().entrySet()` above?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -686,6 +688,7 @@ public static void removeUnfinishedCompactionLeftovers(CFMetaData metadata, Map<
```
inespot, line 695: Let's bump the `logger.trace("Unfinished compactions of {}.{} reference missing sstables of generations {}"` to info?

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -259,27 +260,21 @@ private void completeSetupMayThrowSstableException() {
...
         for (String keyspaceName : Schema.instance.getKeyspaces())
         {
-            // Skip system as we've already cleaned it
+            // Skip system as we'll already clean it after the other tables
             if (keyspaceName.equals(SystemKeyspace.NAME))
                 continue;
 
             for (CFMetaData cfm : Schema.instance.getKeyspaceMetaData(keyspaceName).values())
+            {
+                ColumnFamilyStore.removeUnfinishedCompactionLeftovers(cfm, unfinishedCompactions.getOrDefault(cfm.ksAndCFName, ImmutableMap.of()));
```
inespot: Nit: We should rename this to `deleteUnusedSSTables` or sth along those lines

src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
```diff
@@ -0,0 +1,55 @@
...
+
+
+public class ColumnFamilyStoreManager implements IColumnFamilyStoreValidator
+{
+    public static final ColumnFamilyStoreManager instance = new ColumnFamilyStoreManager();
+    private final List<IColumnFamilyStoreValidator> validators;
+
+    private ColumnFamilyStoreManager()
+    {
+        this.validators = new CopyOnWriteArrayList<>();
```
inespot, line 38: Are you planning on extending this to add other validators? Or is that a design choice in case we do?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5045,31 +5045,35 @@ public void rescheduleFailedDeletions()
      */
     public void loadNewSSTables(String ksName, String cfName)
     {
-        ColumnFamilyStore.loadNewSSTables(ksName, cfName, false);
+        //ColumnFamilyStore.loadNewSSTables(ksName, cfName, false);
+        throw new UnsupportedOperationException("Cannot load SSTables on this version.");
```
inespot: Nit: Add the version once we release

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
zpear: sure

src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
zpear: the latter

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -820,6 +832,10 @@ public synchronized int loadNewSSTablesWithCount()
 
     public synchronized int loadNewSSTablesWithCount(boolean assumeCfIsEmpty)
     {
+        if (assumeCfIsEmpty)
+        {
+            throw new UnsupportedOperationException("Loading new SSTables is not supported on this version");
```
inespot: Nit: Put version in this error as well

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -259,27 +260,21 @@ private void completeSetupMayThrowSstableException() {
...
-        for (Pair<String, String> kscf : unfinishedCompactions.keySet())
-        {
-            CFMetaData cfm = Schema.instance.getCFMetaData(kscf.left, kscf.right);
-            // CFMetaData can be null if CF is already dropped
-            if (cfm != null)
-                ColumnFamilyStore.removeUnfinishedCompactionLeftovers(cfm, unfinishedCompactions.get(kscf));
-        }
-        SystemKeyspace.discardCompactionsInProgress();
-
         // clean up debris in the rest of the keyspaces
```
inespot: Nit: obsolete comment

inespot (dismissed):

d-guo (dismissed):

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -205,24 +209,31 @@ public boolean apply(SSTableReader sstable)
...
             }
+            finally
+            {
+                Directories.removeExpectedSpaceUsedByCompaction(expectedWriteSize, CONSIDER_CONCURRENT_COMPACTIONS);
+                if (taskId != null && (!abortFailed))
+                    SystemKeyspace.finishCompaction(taskId);
+
+                if (collector != null && ci != null)
+                    collector.finishCompaction(ci);
+            }
```
wi11dey, line 236: the abortFailed/readyToFinish flags and checking for suppressed exceptions are working against the intention of the try-with-resources statement, which is to always run all close() method regardless of if the previous ones failed, and suppress the failures from closing https://docs.oracle.com/javase/8/docs/technotes/guides/language/try-with-resources.html, whereas here we explicitly want to skip the SystemKeyspace.finishCompaction() if closing failed. don't need to do this in this PR, but i think this critical section would be less of a workaround + more idiomatic/readable like this:
```
            Optional<UUID> taskId = Optional.empty();
            String taskIdLoggerMsg = UUIDGen.getTimeUUID().toString();
            List<SSTableReader> newSStables;
            Refs<SSTableReader> refs = Refs.ref(actuallyCompact);
            try (AbstractCompactionStrategy.ScannerList scanners = strategy.getScanners(actuallyCompact))
            {
                if (!offline) {
                    taskId = Optional.of(SystemKeyspace.startCompaction(cfs, transaction.originals()));
                    taskIdLoggerMsg = taskId.toString();
                }
                logger.debug("Compacting ({}) {}", taskIdLoggerMsg, ssTableLoggerMsg);
                AbstractCompactionIterable ci = new CompactionIterable(compactionType, scanners.scanners, controller, sstableFormat, taskId);
                try (CloseableIterator<AbstractCompactedRow> iter = ci.iterator())
                {
                    long lastCheckObsoletion = start;

                    if (!controller.cfs.getCompactionStrategy().isActive)
                        throw new CompactionInterruptedException(ci.getCompactionInfo());

                    CompactionAwareWriter writer = getCompactionAwareWriter(cfs, transaction, actuallyCompact);
                    try (CompactionExecutorStatsTracker statsTracker = new CompactionExecutorStatsTracker(ci))
                    {
                        estimatedKeys = writer.estimatedKeys();
                        while (iter.hasNext())
                        {
                            if (ci.isStopRequested())
                                throw new CompactionInterruptedException(ci.getCompactionInfo());

                            try (AbstractCompactedRow row = iter.next())
                            {
                                if (writer.append(row))
                                    totalKeysWritten++;

                                if (System.nanoTime() - lastCheckObsoletion > TimeUnit.MINUTES.toNanos(1L))
                                {
                                    controller.maybeRefreshOverlaps();
                                    lastCheckObsoletion = System.nanoTime();
                                }
                            }
                        }

                        newSStables = writer.finish();
                    } finally {
                        refs.close();
                        writer.close();
                        taskId.ifPresent(SystemKeyspace::finishCompaction);
                    }
                }
            } catch (CompactionInterruptedException e) {
                throw e;
            } catch (Exception e) {
                throw new CompactionException(taskIdLoggerMsg, ssTableLoggerMsg.toString(), e);
            } finally {
                Directories.removeExpectedSpaceUsedByCompaction(expectedWriteSize, CONSIDER_CONCURRENT_COMPACTIONS);
            }

    // ... later inside the toplevel class ...
    private CompactionExecutorStatsTracker implements AutoCloseable {
        private final AbstractCompactionIterable ci;

        public ComapctionExecutorStatsTracker(AbstractCompactionIterable ci) {
            this.ci = ci;
            if (collector != null) {
                collector.beginCompaction(this.ci);
            }
        }
        
        @Override
            public void close() {
            if (collector != null) {
                collector.finishCompaction(ci);
            }
        }
    }
```
should have the same behavior but without needing to reason about checkpoint flags

this way we also don't remove the compaction from compactions_in_progress before the refs are closed (=ancestors deleted). technically we're safe from resurrection either way because after this PR we're planning to run the cleanup unconditionally but this way we're not relying on that change to save us from the ordering issues in this method

src/java/com/palantir/cassandra/db/ColumnFamilyStoreManager.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public void registerValidator(IColumnFamilyStoreValidator validator)
+    {
+        validators.add(validator);
+    }
+
+    public void deregisterValidator(IColumnFamilyStoreValidator validator)
```
wi11dey: nit
```suggestion
    public void unregisterValidator(IColumnFamilyStoreValidator validator)
```
for consistency with other register/unregister methods in the repo e.g. https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/gms/Gossiper.java#L278

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -820,6 +859,10 @@ public synchronized int loadNewSSTablesWithCount()
 
     public synchronized int loadNewSSTablesWithCount(boolean assumeCfIsEmpty)
     {
+        if (assumeCfIsEmpty)
+        {
+            throw new UnsupportedOperationException("Loading new SSTables is not supported on version 2.2.18-1.164.0+.");
+        }
```
wi11dey: why only throw if assumeCfIsEmpty ? wouldn't loading sstables dropped into a directory and having this method renumber this still corrupt the ancestor metadata ?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -710,36 +721,64 @@ public static void removeUnfinishedCompactionLeftovers(CFMetaData metadata, Map<
...
+                } else
+                {
+                    // if any of the ancestors were participating in a compaction, finish that compaction
+                    logger.warn("Going to delete leftover compaction ancestor", UnsafeArg.of("desc", desc),
+                                SafeArg.of("keyspace", desc.ksname), SafeArg.of("cf", desc.cfname),
+                                SafeArg.of("generation", desc.generation), ancestorsArg);
+                    SSTable.delete(desc, sstableFiles.getValue());
+                    UUID compactionTaskID = unfinishedCompactions.get(desc.generation);
+                    if (compactionTaskID != null)
+                        SystemKeyspace.finishCompaction(unfinishedCompactions.get(desc.generation));
```
wi11dey: nit
```suggestion
                    UUID compactionTaskID = unfinishedCompactions.get(desc.generation);
                    if (compactionTaskID != null)
                        SystemKeyspace.finishCompaction(compactionTaskID);
```
i don't know why the original code didn't already do this

or optionally
```suggestion
                    Optional.ofNullable(unfinishedCompactions.get(desc.generation)).ifPresent(SystemKeyspace::finishCompaction);
```

wi11dey (dismissed):
approving 1262d50aedc55615090086fd02f6a2ffe5bd4be4 onwards

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
zpear: we only run restores with assumeCfIsEmpty true, and I didn't want to break all the other tests which leverage this

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
zpear: I will follow up with this one!

wi11dey (approved):

PR #552, opened 2024-09-30
Handle failed abort from in-progress compaction

zpear:
If we have a failure while changing sstables from tmp to final during a compaction, this should normally be handled correctly by the Transactional interface.

However, as @inespot noticed if closing the writer also fails, so we do not fully abort all the new sstables, the in-progress compaction entry will still be removed despite the compaction not being fully rolled back.

If you're doing a leveled compaction where you may have multiple product sstables per ancestor, the lack of in-progress compaction log means you'll prefer to delete the ancestors than the product on startup when cleaning unused sstables. This, plus deleting all the tmp sstables, can mean data loss.

Instead if the compaction log is retained if the abort doesn't succeed, on startup the couple non-tmp products will be deleted, all tmp files deleted, and ancestor files retained

Also note that if the process just crashed mid-compaction, this is fine on first startup since we'll still have the record. However you could still hit the case where if you failed mid-startup and cleaned the record, the next startup would have problems.

### DIFF
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -719,7 +719,7 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
 
         allSstableToAncestors = ColumnFamilyStoreManager.instance.filterValidAncestors(metadata, allSstableToAncestors, unfinishedCompactions);
 
-        Set<UUID> cleanedCompactions = new HashSet<>();
+        Set<UUID> cleanedUnfinishedCompactions = new HashSet<>();
         for (Map.Entry<Descriptor, Set<Integer>> sstableToAncestors : allSstableToAncestors.entrySet())
         {
             Descriptor desc = sstableToAncestors.getKey();
@@ -733,14 +733,14 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
                 assert compactionTaskID != null;
                 logger.info("Going to delete unfinished compaction product {}", desc);
                 SSTable.delete(desc, allNonTempSstableFiles.get(desc));
-                cleanedCompactions.add(compactionTaskID);
+                cleanedUnfinishedCompactions.add(compactionTaskID);
             }
             else
             {
                 completedAncestors.addAll(ancestors);
             }
         }
-        cleanedCompactions.forEach(SystemKeyspace::finishCompaction);
+        cleanedUnfinishedCompactions.forEach(SystemKeyspace::finishCompaction);
 
         // remove old sstables from compactions that did complete
         for (Map.Entry<Descriptor, Set<Component>> sstableFiles : directories.sstableLister().list().entrySet())
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -719,6 +719,7 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
 
         allSstableToAncestors = ColumnFamilyStoreManager.instance.filterValidAncestors(metadata, allSstableToAncestors, unfinishedCompactions);
 
+        Set<UUID> cleanedCompactions = new HashSet<>();
         for (Map.Entry<Descriptor, Set<Integer>> sstableToAncestors : allSstableToAncestors.entrySet())
         {
             Descriptor desc = sstableToAncestors.getKey();
@@ -732,13 +733,14 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
                 assert compactionTaskID != null;
                 logger.info("Going to delete unfinished compaction product {}", desc);
                 SSTable.delete(desc, allNonTempSstableFiles.get(desc));
-                SystemKeyspace.finishCompaction(compactionTaskID);
+                cleanedCompactions.add(compactionTaskID);
             }
             else
             {
                 completedAncestors.addAll(ancestors);
             }
         }
+        cleanedCompactions.forEach(SystemKeyspace::finishCompaction);
 
         // remove old sstables from compactions that did complete
         for (Map.Entry<Descriptor, Set<Component>> sstableFiles : directories.sstableLister().list().entrySet())
--- a/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
@@ -184,6 +184,8 @@ public class CompactionTask extends AbstractCompactionTask
                     if (collector != null)
                         collector.beginCompaction(ci);
 
+                    boolean abortFailed = false;
+                    boolean readyToFinish = false;
                     try (CompactionAwareWriter writer = getCompactionAwareWriter(cfs, transaction, actuallyCompact))
                     {
                         estimatedKeys = writer.estimatedKeys();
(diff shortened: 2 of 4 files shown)


src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -732,13 +733,14 @@ public static void removeUnusedSstables(CFMetaData metadata, Map<Integer, UUID>
...
+                cleanedCompactions.add(compactionTaskID);
             }
             else
             {
                 completedAncestors.addAll(ancestors);
             }
         }
+        cleanedCompactions.forEach(SystemKeyspace::finishCompaction);
 
         // remove old sstables from compactions that did complete
```
zpear: necessary so if we crash mid-startup we won't nuke the compaction log and corrupt anyway

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -205,17 +207,24 @@ public boolean apply(SSTableReader sstable)
...
                     }
                     finally
                     {
                         Directories.removeExpectedSpaceUsedByCompaction(expectedWriteSize, CONSIDER_CONCURRENT_COMPACTIONS);
-                        // point of no return -- the new sstables are live on disk; next we'll start deleting the old ones
-                        // (in replaceCompactedSSTables)
-                        if (taskId != null)
+                        boolean abortMayHaveFailed = readyToFinish && writerClosureFailed;
+                        if (taskId != null && (!abortMayHaveFailed))
+                        // if (taskId != null)
```
zpear: this is the critical point

test/unit/org/apache/cassandra/io/sstable/SSTableRewriterTest.java
```diff
@@ -103,6 +104,121 @@ public static void revertDiskAccess()
         DatabaseDescriptor.setIndexAccessMode(indexMode);
     }
 
+    public static void truncate(ColumnFamilyStore cfs)
```
zpear: ignore changes here: largely autoformat

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
inespot: And so we don't make contrary cleanup decisions on "single compaction create multiple sstables" resulting sstables that still have their relevant compaction id in the `compaction_in_progress` table

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -205,17 +207,24 @@ public boolean apply(SSTableReader sstable)
                             }
                         }
 
-                        // don't replace old sstables yet, as we need to mark the compaction finished in the system table
+                        readyToFinish = true;
                         newSStables = writer.finish();
                     } catch (Exception e) {
-                        throw new CompactionException(taskIdLoggerMsg, ssTableLoggerMsg.toString(), e);
+                        CompactionException exception = new CompactionException(taskIdLoggerMsg, ssTableLoggerMsg.toString(), e);
+                        if (readyToFinish && e.getSuppressed() != null && e.getSuppressed().length != 0)
```
inespot, line 214: Can a suppressed exception not be null but have a size of 0?

inespot: And we have confirmed that the only time a suppressed exception is present is when the close statement fails?

zpear: Throwable should return us an empty array instead of null, but I figure the null check is still nice to have in case this changes between java versions since it's only agreed to in javadoc

zpear: > And we have confirmed that the only time a suppressed exception is present is when the close statement fails?

This is how suppressed is supposed to be used, yes. In my looking this appears to be the case as well. Ideally even if we accidentally keep the compaction log around it should be relatively infrequent + in theory harmless then removed the next startup

src/java/org/apache/cassandra/io/sstable/SSTableRewriter.java
```diff
@@ -279,6 +279,10 @@ public void run()
         }
     }
 
+    public void insertWriterTestingOnly(int index, SSTableWriter newWriter) {
```
inespot, line 283: `@VisibleForTesting` ?

test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
```diff
@@ -533,6 +548,104 @@ private void testDontPurgeAccidentaly(String k, String cfname) throws Interrupte
...
+    {
+
+        public FailedAbortCompactionWriter(ColumnFamilyStore cfs, LifecycleTransaction txn, Set<SSTableReader> nonExpiredSSTables, long maxSSTableSize, int level, boolean offline, OperationType compactionType)
+        {
+            super(cfs, txn, nonExpiredSSTables, maxSSTableSize, level, offline, compactionType);
+        }
+
+        @Override
+        public void doPrepare() {
+            // figure out how to make this crash in the rewriter
```
inespot: You could extend the SsTableRewriter and override the SstableTableRewriter to use your new class after the `super()` call in the constructor?

test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
```diff
@@ -533,6 +548,104 @@ private void testDontPurgeAccidentaly(String k, String cfname) throws Interrupte
...
+        Set<SSTableReader> compacting = Sets.newHashSet(s, s2);
+        LifecycleTransaction txn = cfs.getTracker().tryModify(compacting, OperationType.UNKNOWN);
+        CompactionTask compaction = spy(new FailedAbortCompactionTask(cfs, txn, 0, CompactionManager.NO_GC, 1024 * 1024, true));
+        try
+        {
+            compaction.runMayThrow();
+        }
+        catch (Throwable e)
+        {
+            assertNotNull(e);
```
inespot, line 580: Assert exception is what you expect as well? Either by adding a message and/or checking the suppressed exception?

test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
inespot: Or is this comment obsolete?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -719,6 +719,7 @@ public static void removeUnusedSstables(CFMetaData metadata, Map<Integer, UUID>
 
         allSstableToAncestors = ColumnFamilyStoreManager.instance.filterValidAncestors(metadata, allSstableToAncestors, unfinishedCompactions);
 
+        Set<UUID> cleanedCompactions = new HashSet<>();
```
inespot: Nit: `cleanedUnfinishedCompactions` if we wanna be super precise

src/java/org/apache/cassandra/io/sstable/SSTableRewriter.java
zpear: I can add the annotation, but it still can only be accessed from the same package unless it's public

test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
zpear: woops. obsolete

inespot (approved):

PR #553, opened 2024-10-04
Disable is_new_cluster 4 day check when not join ring

estherg77:


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -701,9 +701,13 @@ else if (conf.num_tokens > MAX_NUM_TOKENS)
                 Path systemDirectoryPath = Paths.get(conf.data_file_directories[0] + "/system");
                 Instant fourDaysAgo = Instant.now().minus(4, ChronoUnit.DAYS);
 
-                if (Files.exists(systemDirectoryPath) && Files.getLastModifiedTime(systemDirectoryPath).toInstant().isBefore(fourDaysAgo))
+                if (Files.exists(systemDirectoryPath)
+                    && Files.getLastModifiedTime(systemDirectoryPath).toInstant().isBefore(fourDaysAgo)
+                    && Boolean.parseBoolean(System.getProperty("cassandra.join_ring", "true")))
```
wi11dey: nit: is it possible to consolidate this with https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/service/StorageService.java#L221 ?
```suggestion
                    && StorageService.joinRing)
```
i'm fine making `StorageService.joinRing` public for this purpose, as that's already the case for ring delay https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/service/StorageService.java#L112

wi11dey (approved):

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
estherg77: Updated to use StorageService.joinRing  (I didn't see a specific ordering for public/private variable in StorageService so I simply made joinRing public in place)

PR #556, opened 2024-10-11
writePredictedSpeculativeRetryPerformanceMetrics computes extra replica snapshot once

schlosna:
While reviewing some JFRs, `AbstractReadExecutor.writePredictedSpeculativeRetryPerformanceMetrics()` and `PredictedSpeculativeRetryPerformanceMetrics.maybeWriteMetrics(ColumnFamilyStore cfs, Collection<Long> latencies, Optional<Snapshot> extraReplicaSnapshot)` pop up with heavy allocations that we can reduce by computing the extra replica snapshot once per `writePredictedSpeculativeRetryPerformanceMetrics` invocation

image
image
image

### DIFF
--- a/src/java/com/palantir/cassandra/utils/RangeTombstoneCountingIterator.java
+++ b/src/java/com/palantir/cassandra/utils/RangeTombstoneCountingIterator.java
@@ -23,6 +23,7 @@ import java.util.Iterator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.codahale.metrics.Snapshot;
 import org.apache.cassandra.db.ColumnFamily;
 import org.apache.cassandra.db.ColumnFamilyStore;
 import org.apache.cassandra.db.DeletionInfo;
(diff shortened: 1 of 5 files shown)


wi11dey:
lgtm, cutting rc to smoke test

src/java/com/palantir/cassandra/utils/RangeTombstoneCountingIterator.java
```diff
@@ -23,6 +23,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.codahale.metrics.Snapshot;
 import org.apache.cassandra.db.ColumnFamily;
```
rhuffy: nit: unused import

schlosna: Will update.

Is there a trick to getting this repo to play nicely with IntelliJ to autofix imports? I assume we don't want to apply the Palantir-java-format as Cassandra has a very different source style and backports would be much more difficult.

schlosna: updated

PR #557, opened 2024-10-17
Write-ahead log for SSTable deletions

wi11dey:


src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -155,114 +158,124 @@ public boolean apply(SSTableReader sstable)
...
-
-                if (collector != null && ci != null)
-                    collector.finishCompaction(ci);
+        }
+        finally
+        {
+            if (!readyToFinish) {
+                // TODO(wdey): refactor all of the trys
+                try (Refs closedRefs = refs; CompactionController closedController = controller) {}
             }
```
wi11dey, line 234: this is unfortunate; we should do something like https://github.com/palantir/cassandra/pull/549#discussion_r1794454098 as a follow up

src/java/org/apache/cassandra/db/lifecycle/LifecycleTransaction.java
```diff
@@ -104,7 +104,7 @@ public String toString()
     private final Set<UniqueIdentifier> identities = Collections.newSetFromMap(new IdentityHashMap<UniqueIdentifier, Boolean>());
 
     // changes that have been made visible
-    private final State logged = new State();
+    public final State logged = new State();
```
zpear: nit: can you just make a method that returns an immutablecopy of logged.obsolete instead of changing access here

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -99,6 +101,7 @@ public boolean reduceScopeForLimitedSpace(long expectedSize)
      * which are properly serialized.
      * Caller is in charge of marking/unmarking the sstables as compacting.
      */
+    @SuppressWarnings("resource") // It is dangerous to close refs for a failed transaction
```
zpear, line 103: we should add a larger comment detailing the specifics of what we're trying to avoid/fix here

src/java/com/palantir/cassandra/db/IColumnFamilyStoreValidator.java
```diff
@@ -33,4 +33,22 @@ public interface IColumnFamilyStoreValidator
      */
     Map<Descriptor, Set<Integer>> filterValidAncestors(CFMetaData cfMetaData,
                                                        Map<Descriptor, Set<Integer>> sstableToCompletedAncestors, Map<Integer, UUID> unfinishedCompactions);
+
+    /**
+     * @return true if Cassandra should use ancestry metdata to cleanup unused SSTables on startup by running
+     * {@link org.apache.cassandra.db.ColumnFamilyStore#removeUnusedSstables(CFMetaData, Map)}, false otherwise
+     * (e.g. if a different cleanup system is being used outside of {@link org.apache.cassandra.service.CassandraDaemon}).
+     */
+    default boolean shouldRemoveUnusedSstables() {
```
zpear: nit: can the method name include some reference to this being for the native/default/ancestor-derived behavior, instead of just in the javadoc?

wi11dey: renamed to `shouldRemoveUnusedSstablesBasedOnAncestorMetadata `

PR #561, opened 2024-10-23
Log changes to TokenMetadata

nickbar01234:
# Description

v0 implementation for logging changes to `TokenMetadata`. The changes include:

* For `TokenMetadata#updateNormalTokens`, log the argument after `StorageService#joined` evaluates to true.
* For `TokenMetadata#calculatePendingRanges`, differentiate between endpoint leaving and bootstrap. When a node is bootstrapping, we calculate the previous owner that this node would have taken the ranges from.

## updateNormalTokens

We only log the changes after the set up has completed to reduce noises on [first startup](https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/service/StorageService.java#L691), where `StorageService` update tokens from `System.peers`. If this is an existing node that is restarting, we will get an extra log line for ownership that we already know about from `System.local`, but I reason this one extra log line should be fine.

### DIFF
--- a/src/java/org/apache/cassandra/locator/TokenMetadata.java
+++ b/src/java/org/apache/cassandra/locator/TokenMetadata.java
@@ -25,6 +25,7 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
 
 import com.google.common.collect.*;
 import org.apache.commons.lang3.StringUtils;
@@ -962,11 +963,9 @@ public class TokenMetadata
         // For each of the bootstrapping nodes, simply add and remove them one by one to
         // allLeftMetadata and check in between what their ranges would be.
         Multimap<InetAddress, Token> bootstrapAddresses = bootstrapTokens.inverse();
+        Multimap<Range<Token>, InetAddress> addressRangesSnapshot = strategy.getRangeAddresses(allLeftMetadata);
         for (InetAddress endpoint : bootstrapAddresses.keySet())
         {
-            // Needs to before call to updateNormalTokens
-            Multimap<Range<Token>, InetAddress> addressRangesSnapshot = strategy.getRangeAddresses(allLeftMetadata);
-
             Collection<Token> tokens = bootstrapAddresses.get(endpoint);
             allLeftMetadata.updateNormalTokens(tokens, endpoint);
             Collection<Range<Token>> addressRangesForEndpoint = strategy.getAddressRanges(allLeftMetadata).get(endpoint);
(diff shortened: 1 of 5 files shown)


src/java/org/apache/cassandra/locator/TokenMetadata.java
```diff
@@ -351,6 +370,8 @@ private void addBootstrapTokens(Collection<Token> tokens, InetAddress endpoint,
 
             for (Token token : tokens)
                 bootstrapTokens.put(token, endpoint);
+
+            logMultiMapDifference(String.format("bootstrapTokens changes for %s", original == null ? "bootstrap" : "node replacement"), bootstrapTokensSnapshot, bootstrapTokens.inverse());
```
rhuffy: this is technically not a CompileTimeConstant, so we might have some issues marking these log messages as safe. Could you instead make "bootstrap" or "node replacement" a separate SafeArg that is logged?

src/java/org/apache/cassandra/locator/TokenMetadata.java
```diff
@@ -112,13 +117,21 @@ public TokenMetadata()
     {
         this(SortedBiMultiValMap.<Token, InetAddress>create(null, inetaddressCmp),
              HashBiMap.<InetAddress, UUID>create(),
-             new Topology());
+             new Topology(), false);
     }
 
-    private TokenMetadata(BiMultiValMap<Token, InetAddress> tokenToEndpointMap, BiMap<InetAddress, UUID> endpointsMap, Topology topology)
+    public TokenMetadata(boolean shouldLogTokenChanges)
+    {
```
rhuffy: instead of plumbing though a boolean variable `shouldLogTokenChanges`, what about instead adding a jvm flag `palantir_cassandra.log_token_changes`.

Then, in the logging methods, you can just call `Boolean.getBoolean("palantir_cassandra.log_token_changes")`

In general I want to avoid making changes to method specs where possible.

nickbar01234: `TokenMetadata` internally also create instances of `TokenMetadata`, which calls methods relevant to the logging ([ref](https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/locator/TokenMetadata.java#L649)). Adding this flag in the constructor ensures that we only get the logs from the singleton `TokenMetadata` that's used by `StorageService`.

rhuffy:
Looking at all the places that `tokenToEndpointMap` is mutated, there is only one call to `put` and two calls to `removeValue`. For the call to put, there is already a warning
```
logger.warn("Token {TOKEN} changing ownership from {ENDPOINT} to {ENDPOINT}"
```

why not just add similar statements around the one call to `bootstrapTokens.put`?

What additional information are these calculations doing?

src/java/org/apache/cassandra/locator/TokenMetadata.java
```diff
@@ -112,13 +118,21 @@ public TokenMetadata()
...
+
+    public TokenMetadata(boolean shouldLogTokenChanges)
+    {
+        this(SortedBiMultiValMap.<Token, InetAddress>create(null, inetaddressCmp),
+             HashBiMap.<InetAddress, UUID>create(),
+             new Topology(), shouldLogTokenChanges);
     }
 
-    private TokenMetadata(BiMultiValMap<Token, InetAddress> tokenToEndpointMap, BiMap<InetAddress, UUID> endpointsMap, Topology topology)
+    private TokenMetadata(BiMultiValMap<Token, InetAddress> tokenToEndpointMap, BiMap<InetAddress, UUID> endpointsMap , Topology topology, boolean shouldLogTokenChanges)
```
rhuffy: nit whitespace after `endpointsMap`

rhuffy (dismissed):
LGTM, just left a whitespace nit

rhuffy (approved):

PR #562, opened 2024-10-24
Write-ahead log dry run

wi11dey:
Create the interface for the write-ahead log and call it at the right place (after a successful transaction). This PR does not expect the wal to be used and will continue closing refs/deleting inputs as usual (not blocking on a successful wal append) and cannot benefit from what the wal fixes. https://github.com/palantir/cassandra/pull/557 will actually block on deletions until a successful wal append). This pr is only for dry-running the wal system and logging

### DIFF
--- a/src/java/com/palantir/cassandra/db/IColumnFamilyStoreValidator.java
+++ b/src/java/com/palantir/cassandra/db/IColumnFamilyStoreValidator.java
@@ -39,7 +39,7 @@ public interface IColumnFamilyStoreValidator
      * {@link org.apache.cassandra.db.ColumnFamilyStore#removeUnusedSstables(CFMetaData, Map)}, false otherwise
      * (e.g. if a different cleanup system is being used outside of {@link org.apache.cassandra.service.CassandraDaemon}).
      */
-    default boolean shouldRemoveUnusedSstables() {
+    default boolean shouldRemoveUnusedSstablesBasedOnAncestorMetadata() {
         return true;
     }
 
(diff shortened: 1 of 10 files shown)


src/java/org/apache/cassandra/io/sstable/SSTable.java
```diff
@@ -105,20 +106,24 @@ public static boolean delete(Descriptor desc, Set<Component> components)
...
-        FileUtils.delete(desc.filenameFor(Component.SUMMARY));
+        deleteWithLog(FileUtils::delete, desc.filenameFor(Component.SUMMARY));
 
-        logger.trace("Deleted {}", desc);
         return true;
     }
 
+    private static void deleteWithLog(Consumer<String> deleter, String filename) {
+        deleter.accept(filename);
+        logger.info("Deleted {}", filename);
```
zpear: nit: safearg the ks, cf, and filename (not full path) of what we're deleting

src/java/com/palantir/cassandra/db/IColumnFamilyStoreValidator.java
```diff
@@ -33,4 +33,22 @@ public interface IColumnFamilyStoreValidator
...
+     * (e.g. if a different cleanup system is being used outside of {@link org.apache.cassandra.service.CassandraDaemon}).
+     */
+    default boolean shouldRemoveUnusedSstablesBasedOnAncestorMetadata() {
+        return true;
+    }
+
+    /**
+     * @return true if Cassandra should skip cleaning up ancestors during
+     * {@link org.apache.cassandra.db.ColumnFamilyStore#removeUnusedSstables(CFMetaData, Map)}, false otherwise. Note
+     * that this flag does not control whe
```
zpear: nit: incomplete comment

zpear (approved):

PR #563, opened 2024-10-24
Prohibit counter columns

wi11dey:
Having this precondition will allow us to assume that it is always safe to keep all inputs to a compaction


zpear (approved):

tpetracca:
should we also fail on attempts to modify the schema at runtime which introduces a counter column? this PR only touches validating the existing schema on startup.

wi11dey:
yeah good idea, this one was to ensure nobody was currently using counter columns

done https://github.com/palantir/cassandra/pull/572

PR #565, opened 2024-10-25
Add Ancestor Filter to sstablemetadata

d-guo:
Add Ancestor Filter to sstablemetadata. It is often of interest to search for sstables that have a specific generation as an ancestor. Specifically this will allow us to find all compactions that an ancestor has participated in.

### DIFF
--- a/src/java/org/apache/cassandra/tools/SSTableMetadataViewer.java
+++ b/src/java/org/apache/cassandra/tools/SSTableMetadataViewer.java
@@ -69,7 +69,7 @@ public class SSTableMetadataViewer
         Util.initDatabaseDescriptor();
 
         String[] ancestorsMaybeNull = cmd.getOptionValues(ANCESTOR_OPTION);
-        Set<String> ancestors = Arrays.stream(ancestorsMaybeNull != null ? ancestorsMaybeNull : new String[0]).collect(Collectors.toSet());
+        Set<Integer> ancestors = Arrays.stream(ancestorsMaybeNull != null ? ancestorsMaybeNull : new String[0]).map(Integer::parseInt).collect(Collectors.toSet());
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/tools/SSTableMetadataViewer.java
```diff
@@ -20,32 +20,58 @@
...
-            out.println("Usage: sstablemetadata <sstable filenames>");
+            System.err.println(e1.getMessage());
+            System.err.println(usage);
             System.exit(1);
         }
 
         Util.initDatabaseDescriptor();
 
-        for (String fname : args)
+        String[] ancestorsMaybeNull = cmd.getOptionValues(ANCESTOR_OPTION);
```
zpear: nit: Optional.ofNullable and .orElseGet

src/java/org/apache/cassandra/tools/SSTableMetadataViewer.java
```diff
@@ -20,32 +20,58 @@
...
+        Set<Integer> ancestors = Arrays.stream(ancestorsMaybeNull != null ? ancestorsMaybeNull : new String[0]).map(Integer::parseInt).collect(Collectors.toSet());
+        boolean filterAncestors = !ancestors.isEmpty();
```
zpear, line 72: any case where filtering for sstables with no ancestors would be useful?

d-guo: i can imagine that this would be useful, but that should be done by a new option like `-e`. `boolean filterAncestors = !ancestors.isEmpty();` maintains prior behavior where if you don't specify `-a` options, no filtering will be applied

zpear (approved):

PR #576, opened 2024-11-07
Remove commons-codec

nickbar01234:
# Description

`commons-codec` 1.2 has [vulnerability](https://issues.apache.org/jira/browse/CODEC-114), which is resolved in 1.5. During investigation, I found that we don't use this package directly. In fact, Cassandra 5 removes this [dependency](https://issues.apache.org/jira/browse/CASSANDRA-18772). The closest usage I found for `commons-codec` is for metrics-reporter in [Cassandra 3](https://issues.apache.org/jira/browse/CASSANDRA-12790) - however, upon inspecting Cassandra startup flags, I don't see any usage of `-Dcassandra.metricsReporterConfigFile`.

I've confirmed that logs and metrics are still shipped with an RC.


rhuffy:
Also need to remove the jar itself if it's checked in

nickbar01234:
Closing in favor of bumping to fixed version

PR #577, opened 2024-11-08
regenerate interface code on thrift 0.17.0

tpetracca:
In https://github.com/palantir/cassandra/pull/432 we manually modified the generated code in order to bump the runtime dependency on libthrift to 0.17.0, so now we can't make changes to thrift and re-generate without having to re-apply the manual changes.

This change updates the code-gen to be generated from thrift 0.17.0, which shouldn't require the manual changes anymore.

In order to install thrift 0.17.0 locally I used an old git commit of the Homebrew source and installed from like this...
```
homebrew-core tpetracca$ git remote -v
origin	https://github.com/Homebrew/homebrew-core.git (fetch)
origin	https://github.com/Homebrew/homebrew-core.git (push)
### this commit hash is the last change to the thrift formula before they upgrade from 0.17.0 to 0.18.0
homebrew-core tpetracca$ git checkout 8f0b9dde69b560dbffad719e3a341fa3018cca7d
HEAD is now at 8f0b9dde69b thrift: brew style --fix
homebrew-core tpetracca$ brew install ./Formula/thrift.rb
... it installing ...
```


build.xml
```diff
@@ -794,7 +794,7 @@
             description="Generate Thrift Java artifacts">
       <echo>Generating Thrift Java code from ${basedir}/interface/cassandra.thrift...</echo>
       <exec executable="thrift" dir="${basedir}/interface" failonerror="true">
-        <arg line="--gen java:hashcode" />
+        <arg line="--gen java" />
```
tpetracca: This argument was removed in `0.9.2` for [THRIFT-2263](https://issues.apache.org/jira/browse/THRIFT-2263) via https://github.com/apache/thrift/commit/acdac816659c88e7b8b601b4ad42dc43bf7d48e2

Before `0.9.2` thrift's java impl did not generate `equals()` and `hashcode()` methods for classes by default unless the `hashcode` flag was set. As of `0.9.2` those methods are generated by default.

interface/cassandra.thrift
```diff
@@ -25,7 +25,6 @@
 
 namespace java org.apache.cassandra.thrift
 namespace cpp org.apache.cassandra
-namespace csharp Apache.Cassandra
```
tpetracca, line 28: The `csharp` generator was deprecated in thrift `0.13.0` and removed in `0.14.0`. We don't have any C# cassandra thrift clients, so it doesn't matter.

rhuffy:
I attempted the codegen locally and it included a license header on each file. Are you intentionally removing those headers?

rhuffy:
I've also created a homebrew tap that should make it easier install thrift in thrift in the future 

`brew install rhuffy/thrift/thrift@0.17`

edit: this may now be installed via an internal tap here https://pl.ntr/2pG

rhuffy:
👍

rhuffy (dismissed):

inespot:
@rhuffy it seems your latest rc tracked: https://github.com/palantir/cassandra/commit/0c57e05064988e8426b38b2fe07d0cd9a4218787. Can you give more context on the changes since then? More specifically this commit: https://github.com/palantir/cassandra/pull/577/commits/a82c1c864dd1f2b111a07ee3d8981bf410764f82
[Edit] Ah nvm, seems like most of the changes from this commit are just dates

inespot (approved):
Soaked on swirl for 3 days

rhuffy:
> Ah nvm, seems like most of the changes from this commit are just dates

Yes, that's correct. I also set an option in `build.xml` to not include the dates, so future re-runs of `ant gen-thrift-java` can be no-ops if the interface or generator has not changed.

PR #578, opened 2024-11-08
Time the runtime of Drain and DrainOnShutdown

rhuffy:
These runtimes will be useful to determine reasonable settings for K8s termination grace period.

LogQL can be used to treat these logs lines as metrics for aggregating these values across many clusters.

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -706,6 +706,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
             public void runMayThrow() throws InterruptedException
             {
                 Stopwatch watch = Stopwatch.createStarted();
+                logger.info("Executing drainOnShutdown hook");
 
                 inShutdownHook = true;
                 ExecutorService counterMutationStage = StageManager.getStage(Stage.COUNTER_MUTATION);
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -37,6 +37,7 @@ import javax.management.openmbean.TabularDataSupport;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
+import com.google.common.base.Stopwatch;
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -754,6 +757,8 @@ public void runMayThrow() throws InterruptedException
                 ScheduledExecutors.nonPeriodicTasks.shutdown();
                 if (!ScheduledExecutors.nonPeriodicTasks.awaitTermination(1, MINUTES))
                     logger.warn("Miscellaneous task executor still busy after one minute; proceeding with shutdown");
+
+                logger.info("DrainOnShutdown completed in {} ms", SafeArg.of("ms", watch.elapsed(TimeUnit.MILLISECONDS)));
```
inespot, line 762: Would it not be simpler to just add metrics? Or is there any advantages to using the logs?

rhuffy: Since we poll the metrics endpoints, it's likely that a metric reported just before shutdown would be missed.

LogQL supports querying logs as if they were metrics, so there is no benefit to using metrics here over a log.

Metrics are also a bit more complicated to set up, and are better suited for values that should be continuously sampled, instead of just an infrequently fired event.

inespot: Can you add logs at the start of each function (so we understand if a SIGKILL was sent before the thread completed) flush each of the logs?

rhuffy: added a log to drainOnShutdown

the `setMode` calls i.e:`setMode(Mode.DRAINING, "starting drain process", true);` will emit a log

```
    void setMode(Mode m, @Safe String msg, boolean log)
    {
        operationMode = m;
        if (log)
            logger.info(m.toString(), SafeArg.of("msg", msg));
        else
            logger.debug(m.toString(), SafeArg.of("msg", msg));
    }
```

inespot (approved):
From offline conv: as a FLUP we need to properly shutdown the logger in sls in order to make sure those logs do not get dropped when Cassandra shuts down gracefully

PR #579, opened 2024-11-11
Phi threshold can be set via nodetool

inespot:
Add support for dynamically setting phi threshold.
Removes the need to perform a rolling restart of a C* cluster in the event of a migration rollback.


src/java/com/palantir/cassandra/metrics/FailureDetectorMetrics.java
```diff
@@ -19,18 +19,35 @@
...
+    static
+    {
+        Metrics.register(
+            createMetricName("FailureDetectorPhiThreshold"), new Gauge<Double>()
+        {
+            public Double getValue()
+            {
+                return DatabaseDescriptor.getPhiConvictThreshold();
+            }
+        });
```
wi11dey:
```suggestion
        Metrics.register(createMetricName("FailureDetectorPhiThreshold"), (Gauge<Double>) DatabaseDescriptor::getPhiConvictThreshold);
```

src/java/com/palantir/cassandra/tools/nodetool/GetPhiConvictThreshold.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.palantir.cassandra.tools.nodetool;
+
+import org.apache.cassandra.tools.NodeProbe;
+import org.apache.cassandra.tools.NodeTool;
+
+import io.airlift.command.Command;
+
+@Command(name = "getphiconvictthreshold", description = "Get failure detector phi threshold value")
+public class GetPhiConvictThreshold extends NodeTool.NodeToolCmd
```
wi11dey, line 25: why do we need this instead of just checking cassandra.yaml ?

wi11dey: from offline: to grab the result of a previous in-memory set from nodetool

wi11dey (approved):

PR #583, opened 2024-11-20
Ensure schema is consistent while bootstrapping

rhuffy:
This PR addresses a potential edge case where the Schema is updated while bootstrapping. The following changes are made

- Ensure that the local schema version matches the versions for all peers reported by the `Gossiper` before Bootstrapping
- After creating bootstrap streams, ensure that the local schema version has not changed, and is not in the process of changing.


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1048,6 +1060,18 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
                 unsafeDisableNode();
                 throw new BootstrappingSafetyException("Bootstrap streaming failed.");
             }
+
+            if(!localSchemaVersion.equals(Schema.instance.getVersion().toString()) || !isSchemaConsistent(localSchemaVersion))
```
d-guo: i wonder if we are being too restrictive by requiring this. as far as i can tell, it's actually okay for the schema to change bootstrap has started as long as the joining node is receiving forwarded writes + the new keyspace/columnfamily. this is already required for correctness in normal bootstraps (without schema changes), so i'm not sure we should be more strict here.

maybe the more "correct" check here is that other nodes see the joining node and are forwarding writes (presumably if the joining node is in their gossip, this is true). i think the current code already tries to do this via sleeping for 30 seconds to let gossip work and checking that gossip has seen the seeds
```
            Gossiper.instance.addLocalApplicationStates(states);
            setMode(Mode.JOINING, "sleeping " + RING_DELAY + " ms for pending range setup", true);
            Uninterruptibles.sleepUninterruptibly(RING_DELAY, TimeUnit.MILLISECONDS);
[...]

        if (!Gossiper.instance.seenAnySeed())
            throw new IllegalStateException("Unable to contact any seeds!");
```
but doesn't appear to guarantee that all writes are forwarded

rhuffy: I think you're right that this check is more restrictive than we need to be.

src/java/org/apache/cassandra/dht/BootStrapper.java
```diff
@@ -83,6 +86,11 @@ public ListenableFuture<StreamState> bootstrap(StreamStateStore stateStore, bool
             streamer.addRanges(keyspaceName, strategy.getPendingAddressRanges(tokenMetadata, tokens, address));
         }
 
+        if (!initialLocalSchemaVersion.equals(Schema.instance.getVersion()) || !MigrationManager.isReadyForBootstrap())
```
d-guo: i think the order of this check is actually important. `MigrationManager.isReadyForBootstrap()` must be evaluated to true before `initialLocalSchemaVersion.equals(Schema.instance.getVersion())` is evaluated to true. the way it's written now, it's possible for a partially applied schema to complete in between the two checks and not throw here

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -966,6 +962,15 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
                 setMode(Mode.JOINING, "waiting for schema information to complete", true);
                 Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
             }
+
+            while(!isSchemaConsistent())
```
d-guo: i don't see a reason not to just combine this with the previous check and have it be

```
            while (!MigrationManager.isReadyForBootstrap() || !isSchemaConsistent())
            {
                setMode(Mode.JOINING, "waiting for schema information to complete... local schema version is {Schema.instance.getVersion().toString()}", true);
                Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
            }
```

d-guo:
nice! i think these checks are sufficient to guard us against the theoretical edge case we talked about. as discussed, let's hold off on merging this "fix" until we have a repro of the edge case to make sure we aren't missing something here

src/java/org/apache/cassandra/service/StorageService.java
d-guo: this also means we can remove the above `// first sleep the delay to make sure we see all our peers`, right?

src/java/org/apache/cassandra/dht/BootStrapper.java
d-guo: also what's the reason for putting these checks in `Bootstrapper` as opposed to `StorageService#bootstrap`?

d-guo:
and to double check, this doesn't affect spinning up new clusters because they don't bootstrap, right?

d-guo:
does this affect spinning up zombie or new clusters?

src/java/org/apache/cassandra/dht/BootStrapper.java
rhuffy: The original idea was to put the second check in the callback for handling a PREPARE, but I decided that that was not necessary.

Moved the checks to SS

src/java/org/apache/cassandra/service/StorageService.java
rhuffy: I've decided to switch the first loop to just a wait with no short circuit condition

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -949,23 +945,19 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
...
-            {
-                // if we see schema, we can proceed to the next check directly
-                if (!Schema.instance.getVersion().equals(Schema.emptyVersion))
-                {
-                    logger.debug("got schema: {}", Schema.instance.getVersion());
-                    break;
-                }
-                Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
-            }
+            Uninterruptibles.sleepUninterruptibly(delay, TimeUnit.MILLISECONDS);
```
d-guo, line 950: i think we might as well just remove this given the below

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1130,6 +1118,21 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
         }
     }
 
+    private static boolean isSchemaConsistent()
+    {
+        String localSchemaVersion = Schema.instance.getVersion().toString();
+        Set<Entry<InetAddress, EndpointState>> endpointStates = Gossiper.instance.getEndpointStates();
```
d-guo: nit: `return Gossiper.instance.getEndpointStates().stream().map(Entry::getValue).map(entry -> entry.getApplicationState(ApplicationState.SCHEMA).value).allMatch(Schema.instance.getVersion().toString()::equals);`

src/java/org/apache/cassandra/service/StorageService.java
rhuffy: I think having some delay here is still important because we know from experience that you should wait at least some amount of time here before proceeding. The intention of this change is to make the bootstrapping node wait 
- at least 30s
- wait longer if it thinks the schema is consistent

A bad scenario would be if we get the non-empty schema by gossiping with a node in a network partition, think we have gossiped with everyone because we haven't waited long enough, and then pass the subsequent checks.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1130,6 +1116,14 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
         }
     }
 
+    private static boolean isSchemaConsistent()
+    {
+        String localSchemaVersion = Schema.instance.getVersion().toString();
+        return Gossiper.instance.getEndpointStates().stream()
+                                .map(entry -> entry.getValue().getApplicationState(ApplicationState.SCHEMA).value)
+                                .allMatch(localSchemaVersion::equals);
```
d-guo: ah wait doesn't this check cause problems bc the endpoint states can have `LEFT` nodes?

src/java/org/apache/cassandra/service/StorageService.java
d-guo: does that risk exist even if we wait for `isSchemaConsistent`? in that case, we should make our check more strict instead of just relying on a 30 second wait to hope it's safe

rhuffy:
> does this affect spinning up zombie or new clusters?

Confirmed from testing that this does not impact spinning up new clusters. On the first startup of seed nodes, this codepath is not flexed because they do not bootstrap.

src/java/org/apache/cassandra/service/StorageService.java
rhuffy: I've made a change to exclude LEFT nodes, and also gracefully handle cases where ApplicationStates are null

src/java/org/apache/cassandra/service/StorageService.java
rhuffy: :+1:

src/java/com/palantir/cassandra/utils/SchemaAgreementCheck.java
```diff
@@ -0,0 +1,72 @@
...
+        return endpointStatesSupplier.get().stream()
+                                     .map(Map.Entry::getValue)
+                                     .filter(value -> !isLeft(value))
+                                     .allMatch(value -> schemaIsEqualToLocalVersion(localSchemaVersion, value));
+    }
+
+    private boolean isLeft(EndpointState endpointState)
+    {
+        VersionedValue status = endpointState.getApplicationState(ApplicationState.STATUS);
+        return status == null || status.value.startsWith(VersionedValue.STATUS_LEFT);
```
d-guo: is it reasonable/safe to just ignore nodes with `null`? like what happens if another node has been down for a long time does that result in `null`?

rhuffy: I've taken a second look at this logic and refactored it in c6f8db6 to use the `getStatus` helper function, which does some parsing and null checking for us.

In the event that we encounter an endpoint and have `null` for its status, we should _not_ exclude this endpoint from this check, and instead wait for either the Schema of this endpoint to show up, or for the status of the endpoint to show up.

d-guo (approved):

PR #584, opened 2024-11-21
Nodetool Drain throws immediately if the server is not initialized

rhuffy:
Since the drain method is `synchronized`, calls to `nodetool drain` will hang if the server is not initialized. This includes nodes that are bootstrapping.

This PR makes these calls to `drain()` fail fast, so the behavior is more obvious.

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -2638,7 +2638,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
             logger.info("Received removenode gossip about myself. Is this node rejoining after an explicit removenode?");
             try
             {
-                drain();
+                drainInternal();
             }
             catch (Exception e)
             {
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4606,7 +4606,19 @@ public String getDrainProgress()
...
+    public void drain() throws IOException, InterruptedException, ExecutionException
+    {
+        if (daemon.setupCompleted())
+        {
+            drainInternal();
+        }
+        else
+        {
+            throw new IllegalStateException("Cannot drain a node that is bootstrapping");
+        }
```
zpear: this would happen if we called drain during regular startup as well, not just bootstrapping right

zpear: then there's a question: if we're starting up vs bootstrapping is it desirable to wait for startup to complete then drain, or just fail the drain and continue w/ other shutdown actions later?

rhuffy: The only case we should expect Drain to be called when the server is not fully initialized is if someone is taking manual action to terminate a node that is stuck initializing, or taking manual action to abort an in-progress bootstrap or node replacement. These are probably both cases we'd want to terminate without delay.

zpear (approved):
Lgtm

PR #585, opened 2024-11-21
Add some additional logging to StorageService#drain

rhuffy:
This PR bumps some existing log statements from Debug to Info, and adds some additional log statements in the Drain codepath.

The intention here is to get better signal as to which parts of the Drain process take the most time.


nickbar01234 (commented):
LGTM

nickbar01234 (approved):

PR #586, opened 2024-11-22
Add prefix filter to sstable2json

d-guo:
Enables finding all keys with a prefix within an sstable through sstable2json.

### DIFF
--- a/src/java/org/apache/cassandra/tools/SSTableExport.java
+++ b/src/java/org/apache/cassandra/tools/SSTableExport.java
@@ -23,7 +23,9 @@ import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintStream;
+import java.nio.ByteBuffer;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import org.apache.cassandra.io.sstable.format.SSTableReader;
 import org.apache.commons.cli.*;
@@ -49,6 +51,7 @@ public class SSTableExport
     private static final ObjectMapper jsonMapper = new ObjectMapper();
 
     private static final String KEY_OPTION = "k";
+    private static final String PREFIX_OPTION = "p";
     private static final String EXCLUDEKEY_OPTION = "x";
     private static final String ENUMERATEKEYS_OPTION = "e";
 
@@ -66,6 +69,11 @@ public class SSTableExport
         // Number of times -x <key> can be passed on the command line.
         excludeKey.setArgs(500);
         options.addOption(excludeKey);
+
+        Option prefixKey = new Option(PREFIX_OPTION, true, "Prefix of row key");
+        // Number of times -p <prefix> can be passed on the command line.
+        prefixKey.setArgs(1);
+        options.addOption(prefixKey);
 
         Option optEnumerate = new Option(ENUMERATEKEYS_OPTION, false, "enumerate keys only");
         options.addOption(optEnumerate);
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/tools/SSTableExport.java
```diff
@@ -305,6 +313,66 @@ public static void export(Descriptor desc, PrintStream outs, Collection<String>
...
+            while (keysCount < 5000) {
+                if (lastKey != null && lastKey.compareTo(decoratedKey) > 0)
+                    throw new IOException("Key in sstable are out of order despite scanning in order! " + lastKey + " > " + decoratedKey);
+
+                RowIndexEntry entry;
+                if (keysCount == 0) {
+                    entry = sstable.getPosition(decoratedKey, SSTableReader.Operator.GE);
+                } else {
+                    entry = sstable.getPosition(lastKey, SSTableReader.Operator.GT);
+                }
```
wi11dey: won't this keep researching the sstable every key instead of scanning forward from the first ?

d-guo: i think there's two pieces at play here: `sstable.getPosition` will have to binary search for the index for each key but `dfile.seek` will only ever move forward from the previous seeked position. this means searching for a prefix that has 100 keys is the same as if you searched speciically for those 100 keys with `-k`

d-guo:
manual testing
```
rescue@cqlsh> SELECT * FROM dg_sstables.prefix;

 row            | col1 | col2 | value1 | value2
----------------+------+------+--------+--------
   0x6170706c65 |    1 |    1 |     0x |     0x
       0x62616d |    1 |    1 |     0x |     0x
       0x62616e |    1 |    1 |     0x |     0x
 0x62616e616e61 |    1 |    1 |     0x |     0x
   0x7a65627261 |    1 |    1 |     0x |     0x

(5 rows)
rescue@cqlsh> exit
```
ptdocker@dg-expansion-0:/app$ ./service/bin/sstable2json mnt/cassandra-data-0/cassandra/dg_sstables/prefix-c5058400a8fd11efb2cc23a21fa52e0e -p 62

{
"lb-1-big-Data.db":[
{"key": "62616d",
 "cells": [["1:1:","",1732299544590381],
           ["1:1:value1","",1732299544590381],
           ["1:1:value2","",1732299544590381]]},
{"key": "62616e",
 "cells": [["1:1:","",1732299531590633],
           ["1:1:value1","",1732299531590633],
           ["1:1:value2","",1732299531590633]]},
{"key": "62616e616e61",
 "cells": [["1:1:","",1732299525101513],
           ["1:1:value1","",1732299525101513],
           ["1:1:value2","",1732299525101513]]}
]
}
```

**Comment by d-guo** [comment]
manual testing
```
rescue@cqlsh> SELECT * FROM dg_sstables.prefix;

 row            | col1 | col2 | value1 | value2
----------------+------+------+--------+--------
   0x6170706c65 |    1 |    1 |     0x |     0x
       0x62616d |    1 |    1 |     0x |     0x
       0x62616e |    1 |    1 |     0x |     0x
 0x62616e616e61 |    1 |    1 |     0x |     0x
   0x7a65627261 |    1 |    1 |     0x |     0x

(5 rows)
rescue@cqlsh> exit
```

```
ptdocker@dg-expansion-0:/app$ ./service/bin/sstable2json mnt/cassandra-data-0/cassandra/dg_sstables/prefix-c5058400a8fd11efb2cc23a21fa52e0e -p 62

{
"lb-1-big-Data.db":[
{"key": "62616d",
 "cells": [["1:1:","",1732299544590381],
           ["1:1:value1","",1732299544590381],
           ["1:1:value2","",1732299544590381]]},
{"key": "62616e",
 "cells": [["1:1:","",1732299531590633],
           ["1:1:value1","",1732299531590633],
           ["1:1:value2","",1732299531590633]]},
{"key": "62616e616e61",
 "cells": [["1:1:","",1732299525101513],
           ["1:1:value1","",1732299525101513],
           ["1:1:value2","",1732299525101513]]}
]
}
```

File: src/java/org/apache/cassandra/tools/SSTableExport.java (hunk 1 above)
Comment (zpear) [reply src/java/org/apache/cassandra/tools/SSTableExport.java]: If you had a "next prefix" supplied I think you would be able to run this like a normal range scan right? like "prefix1" is the left bound then "prefix2" is the next?

Comment (d-guo) [reply src/java/org/apache/cassandra/tools/SSTableExport.java]: good idea, added!

Comment (wi11dey) [reply src/java/org/apache/cassandra/tools/SSTableExport.java]: i think with the update this
> this means searching for a prefix that has 100 keys is the same as if you searched speciically for those 100 keys with `-k`

is still the case
```
                RowIndexEntry entry = sstable.getPosition(decoratedKey, keysCount == 0 ? SSTableReader.Operator.GE : SSTableReader.Operator.GT);
```
i think like @zpear said we can just use the on disk atom iterator until we hit untilPrefix (or until the prefix doesn't match anymore) instead of researching for every key

zpear (approved):

d-guo:
talked offline — keeping the current implementation because `DataRange` represents a `(exclusive, inclusive]` range

> A Range is responsible for the tokens between (left, right].

we want `IncludingExcludingBounds` but it doesn't appear `SSTableReader` accepts that type of bound in its current form

PR #587, opened 2024-11-26
Make initial sleep bootstrap delay configurable

inespot:
For rc only: to mitigate schema race

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -948,8 +948,10 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                 SystemKeyspace.setBootstrapState(SystemKeyspace.BootstrapState.IN_PROGRESS);
             }
             setMode(Mode.JOINING, "waiting for ring information", true);
+
+            int sleepDelay = bootstrapDelayFactor() * delay;
             // first sleep the delay to make sure we see all our peers
-            for (int i = 0; i < delay; i += 1000)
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5290,6 +5292,12 @@ public boolean localQuorumReadsForSerialCasEnabled() {
         return LocalQuorumReadForSerialCasSetting.instance.isTrue();
     }
 
+
+    public int bootstrapDelayFactor()
+    {
+        return Integer.parseInt(System.getProperty("palantir_cassandra.bootstrap_delay_factor", "1"));
+    }
```
rhuffy, line 5299: usually these are static constants

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -948,8 +948,10 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
                 SystemKeyspace.setBootstrapState(SystemKeyspace.BootstrapState.IN_PROGRESS);
             }
             setMode(Mode.JOINING, "waiting for ring information", true);
+
+            int sleepDelay = bootstrapDelayFactor() * delay;
             // first sleep the delay to make sure we see all our peers
-            for (int i = 0; i < delay; i += 1000)
+            for (int i = 0; i < sleepDelay; i += 1000)
```
rhuffy, line 954: I would remove the loop entirely and just sleep for the entire intended delay

PR #588, opened 2024-11-26
Tweak MessagingService Shutdown and Add Internode Connect Timeout

zpear:
A couple different, but sort of related changes in this PR:

* On MessagingService shutdown we now close the incoming socket before shutting down the callback expiring map. Although in the previous ordering the callback map won't allow new callbacks to be registered, I think other C* nodes may continue successfully sending messages since at the networking layer they're still accepted. This might help a little bit, in conjunction with the next change, to avoid waiting until timeouts for requests we know will fail -- but probably not a ton.
* If one C* node sees another as up, but cannot connect for whatever reason, it will loop trying to reconnect in OutboundTcpConnection for `minRpcTimeout`, which can be the same as read/write timeouts in practice. Once this is exceeded, all the messages on the backlog are cleared. If these were needed to satisfy a read/write, this node as the coordinator just waits until the failure callback is invoked via read/write timeout (which can be a long time).
  * If we make this separately configurable and assume recoverable network problems are short-lived, and non-recoverable ones take substantially longer than read/write timeouts, we can make this connectivity timeout small and invoke the failure callbacks much faster than the write timeout, which might improve the client experience.
  * Normally you'd hope this node sees the other as unreachable in this scenario as then it will not attempt to send a request. However, we've seen this not be the case relatively frequently.
  * Also generally, whenever messages got cleared from the backlog we were not invoking any failure callbacks. We can just do that too instead of waiting for them to timeout

### DIFF
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -88,6 +88,8 @@ public class Config
 
     public long native_transport_idle_timeout_in_ms = 0L;
 
+    public volatile Long internode_connect_timeout_in_ms = 1000L;
+
     public volatile Long request_timeout_in_ms = 10000L;
 
     public volatile Long read_request_timeout_in_ms = 5000L;
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -1077,6 +1077,12 @@ public class DatabaseDescriptor
         return conf.rpc_listen_backlog;
     }
 
+
+    public static long getInternodeConnectionTimeout()
+    {
+        return conf.request_timeout_in_ms;
+    }
+
     public static long getRpcTimeout()
     {
         return conf.request_timeout_in_ms;
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -1080,7 +1080,12 @@ public class DatabaseDescriptor
 
     public static long getInternodeConnectionTimeout()
     {
-        return conf.request_timeout_in_ms;
+        return conf.internode_connect_timeout_in_ms;
+    }
+
+    public static long setInternodeConnectionTimeout(long timeoutInMillis)
+    {
+        return conf.internode_connect_timeout_in_ms = timeoutInMillis;
     }
 
     public static long getRpcTimeout()
--- a/src/java/org/apache/cassandra/net/MessagingService.java
+++ b/src/java/org/apache/cassandra/net/MessagingService.java
@@ -766,17 +766,14 @@ public final class MessagingService implements MessagingServiceMBean
         // We may need to schedule hints on the mutation stage, so it's erroneous to shut down the mutation stage first
         assert !StageManager.getStage(Stage.MUTATION).isShutdown();
 
-        // the important part
-        if (!callbacks.shutdownBlocking())
-            logger.warn("Failed to wait for messaging service callbacks shutdown");
-
         // attempt to humor tests that try to stop and restart MS
         try
         {
             clearMessageSinks();
-            for (SocketThread th : socketThreads)
+            for (SocketThread th : socketThreads) {
                 try
                 {
+                    // Close incoming connections
                     th.close();
                 }
                 catch (IOException e)
@@ -784,6 +781,11 @@ public final class MessagingService implements MessagingServiceMBean
                     // see https://issues.apache.org/jira/browse/CASSANDRA-10545
                     handleIOException(e);
                 }
+
+            }
+            // Wait to finish callbacks before closing outbound connections
+            if (!callbacks.shutdownBlocking())
+                logger.warn("Failed to wait for messaging service callbacks shutdown");
 
             connectionManagers.values().forEach(OutboundTcpConnectionPool::close);
         }
--- a/src/java/org/apache/cassandra/net/MessagingService.java
+++ b/src/java/org/apache/cassandra/net/MessagingService.java
@@ -781,7 +781,6 @@ public final class MessagingService implements MessagingServiceMBean
                     // see https://issues.apache.org/jira/browse/CASSANDRA-10545
                     handleIOException(e);
                 }
-
             }
             // Wait to finish callbacks before closing outbound connections
             if (!callbacks.shutdownBlocking())
--- a/src/java/org/apache/cassandra/net/OutboundTcpConnection.java
+++ b/src/java/org/apache/cassandra/net/OutboundTcpConnection.java
@@ -627,6 +627,7 @@ public class OutboundTcpConnection extends Thread
             if (!qm.isTimedOut())
                 return;
             iter.remove();
+            invokeFailureCallback(qm);
             dropped.incrementAndGet();
         }
     }
--- a/src/java/org/apache/cassandra/net/OutboundTcpConnection.java
+++ b/src/java/org/apache/cassandra/net/OutboundTcpConnection.java
@@ -34,6 +34,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.IntStream;
 import java.util.zip.Checksum;
 
 import javax.net.ssl.SSLHandshakeException;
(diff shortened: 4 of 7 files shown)


src/java/org/apache/cassandra/config/Config.java
```diff
@@ -88,6 +88,8 @@ public class Config
 
     public long native_transport_idle_timeout_in_ms = 0L;
 
+    public volatile Long internode_connect_timeout_in_ms = 1000L;
```
inespot, line 91: So we intentionally want to lower this value for steady state instead of just overriding it for migration? Have we seen this causing issues outside of migrations?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1077,6 +1077,17 @@ public static int getRpcListenBacklog()
...
+
+    public static long getInternodeConnectionTimeout()
+    {
+        return conf.internode_connect_timeout_in_ms;
+    }
+
+    public static long setInternodeConnectionTimeout(long timeoutInMillis)
+    {
+        return conf.internode_connect_timeout_in_ms = timeoutInMillis;
+    }
```
inespot, line 1089: If we plan on tweaking this during migrations, might be worth adding the capability to set this via nodetool

src/java/org/apache/cassandra/config/Config.java
zpear: Yes I'd like to have this just be a default config. We already run a significant portion of clusters with read/write timeouts of 1s, so I'm not worried about inserting this here since I believe socket connection should be fairly independent of cassandra query-processing-time (in other words, this is a no-op for many of our clusters. for some others, which have higher other timeouts, I don't see any reason not to bring this connect timeout in line with the first group of clusters)

If there is _any_ "longer"-lived connectivity problem between nodes (or you do a regular restart + another node doesn't yet see the restarting one as down) I expect this to be beneficial.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
zpear: from other comment: my current plan is for this to be the default and not change during migrations

src/java/org/apache/cassandra/net/OutboundTcpConnection.java
```diff
@@ -236,8 +238,10 @@ else if (socket != null || connect())
                     else
                     {
                         // clear out the queue, else gossip messages back up.
-                        drainedMessages.clear();
-                        backlog.clear();
+                        int cleared = clearQueueWithFailureCallback(i, drainedMessages, drainedMessageSize, backlog);
```
inespot, line 241: What specific callback behavior are you thinking will help here? Or is this just for visibility (eg `failuresUpdater.incrementAndGet(this)`)

zpear: if you invoke the failure callback, I believe AbstractWriteResponseHandler will be able to tell whether the latest failure means we won't get the required number of responses, then call signal() so we're no longer blocking in StorageProxy

zpear: the unit test shows this. the get() returns faster than the write timeout w/ this change

src/java/org/apache/cassandra/config/Config.java
inespot: Do we have any concerns about weird onprem setups where networking might be slower and the SSL handshake might take a little longer?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
inespot: Oki, considering we will still able to set this value via cassandra.yaml config like any of the other timeouts in case we need to override it

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -766,24 +766,25 @@ public void shutdown()
...
                 }
                 catch (IOException e)
                 {
                     // see https://issues.apache.org/jira/browse/CASSANDRA-10545
                     handleIOException(e);
                 }
+            }
+            // Wait to finish callbacks before closing outbound connections
+            if (!callbacks.shutdownBlocking())
+                logger.warn("Failed to wait for messaging service callbacks shutdown");
```
inespot, line 787: Nit: can we make this log and comment slightly more clear? The first time I read this, I thought it was waiting for callbacks to complete but it's actually just shutting down the reaper executor service that expires callbacks? Or is there some synchronization I might have missed?

src/java/org/apache/cassandra/config/Config.java
zpear: I don't think so -- we run 1s timeout clusters onprem without issue in a few places already (and the oss default is already 2s)

src/java/org/apache/cassandra/net/OutboundTcpConnection.java
```diff
@@ -255,6 +259,24 @@ else if (socket != null || connect())
         }
     }
 
+    @VisibleForTesting
+    int clearQueueWithFailureCallback(int currentMessage, List<QueuedMessage> bufferedMessages, int bufferSize, BlockingQueue<QueuedMessage> queue) {
+        bufferedMessages.stream().skip(currentMessage).forEach(this::invokeFailureCallback);
```
inespot, line 264: Are we not failing the current message as well? Considering we have skipped the `writeConnected` and entered the `else` statement above

src/java/org/apache/cassandra/net/MessagingService.java
zpear: I think it is doing what the comment + log say. It's just not obvious due to ExpiringMap's implementation:

in ExpiringMap#put()
```
       if (shutdown)
        {
            // StorageProxy isn't equipped to deal with "I'm nominally alive, but I can't send any messages out."
            // So we'll just sit on this thread until the rest of the server shutdown completes.
            //
            // See comments in CustomTThreadPoolServer.serve, CASSANDRA-3335, and CASSANDRA-3727.
            Uninterruptibles.sleepUninterruptibly(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
```

src/java/org/apache/cassandra/net/OutboundTcpConnection.java
```diff
@@ -255,6 +259,24 @@ else if (socket != null || connect())
...
+        int initialCleared = bufferedMessages.size() - currentMessage;
+        bufferedMessages.clear();
+
+        int queueSize = queue.size();
+        int remaining = queueSize;
+        while (remaining > 0) {
+            remaining -= queue.drainTo(bufferedMessages, Math.min(bufferSize, remaining));
```
inespot, line 271: Why use the same buffer as the `drainedMessageSize`? To avoid changing behavior in how many failure callbacks we could theoretically trigger at the same time?

src/java/org/apache/cassandra/net/OutboundTcpConnection.java
zpear: yeah, current message should be failed here as well, b/c currentMessage is this one's index, while we're skipping that many entries.

So currMessage 1 means we only skip index 0, and still fail index 1

src/java/org/apache/cassandra/net/OutboundTcpConnection.java
zpear: I figured it's nicer to keep the buffering consistent + reuse the same list, whether we're trying to actually send message or fail them we're still processing them.

test/unit/org/apache/cassandra/net/MessagingServiceTest.java
```diff
@@ -53,4 +92,97 @@ public void testDroppedMessages()
...
+        MessagingService.instance().sendRR(message, FBUtilities.getLocalAddress(), handler3, false);
+        handler3.get();
+        Duration handler3Time = Duration.between(start, Instant.now());
+        TestHandler handler4 = createHandler(keyspace);
+        MessagingService.instance().sendRR(message, FBUtilities.getLocalAddress(), handler4, false);
+        handler4.get();
+        Duration handler4Time = Duration.between(start, Instant.now());
+
+        // handler2 may or may not fail vs timeout. Likely due to OS level buffering, flushing the socket that has now
+        // been closed on the receiving end may or may not throw an IOException
```
inespot, line 145: Could you give some more details about this? Based on your comment, handler2 could be expired by the callback reaper? Is the idea that the OS buffering would cause the `connect()` to still succeed?

src/java/org/apache/cassandra/net/MessagingService.java
inespot: Ah. So it also turns off the "ability to add callbacks", on top of shutting down the reaper. This is ... weirdly implemented

test/unit/org/apache/cassandra/net/MessagingServiceTest.java
zpear: > Is the idea that the OS buffering would cause the connect() to still succeed?

Yes. I see the call to the java Socket succeed despite confirming on my machine that the socket has been closed.

inespot (approved):

PR #574, opened 2024-11-07
Enable using the write ahead log

wi11dey:

PR #589, opened 2024-12-02
WAL-only cleanup

wi11dey:
Moves all Cassandra SSTable deletion to only use the new write-ahead logging system, implemented in #562 and #574


inespot:
Need to remove reloadSstable block in this PR as well

PR #592, opened 2024-12-05
migrate to SafeLogging in `org.apache.cassandra.streaming`

rhuffy:
Mark logging arguments as Safe or Unsafe in `org.apache.cassandra.streaming`.

This PR also wraps all calls to `logger.debug()` in an `if (logger.isDebugEnabled())` check.

For reviewers, please confirm that you agree with my choices for SafeArgs.

Everything appears to be safe except
- `partitionKey` in `StreamReader`

Interesting is the `message` variable in `ConnectionHandler`. I validated that these are safe by checking the `toString` methods of all subclasses of `StreamMessage`, but would appreciate a second look at this.

### DIFF
--- a/src/java/org/apache/cassandra/streaming/StreamLockfile.java
+++ b/src/java/org/apache/cassandra/streaming/StreamLockfile.java
@@ -28,6 +28,8 @@ import java.util.List;
 import java.util.UUID;
 
 import com.google.common.base.Charsets;
+
+import com.palantir.logsafe.SafeArg;
 import org.apache.cassandra.io.sstable.format.SSTableWriter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -85,7 +87,7 @@ public class StreamLockfile
         }
         catch (IOException e)
         {
-            logger.warn(String.format("Could not create lockfile %s for stream session, nothing to worry too much about", lockfile), e);
+            logger.warn("Could not create lockfile {} for stream session, nothing to worry too much about", SafeArg.of("lockfile", lockfile.toString()), e);
         }
     }
 
@@ -107,7 +109,7 @@ public class StreamLockfile
             catch (Exception e)
             {
                 JVMStabilityInspector.inspectThrowable(e);
-                logger.warn("failed to delete a potentially stale sstable {}", file);
+                logger.warn("failed to delete a potentially stale sstable {}", SafeArg.of("file", file));
             }
         }
     }
--- a/src/java/org/apache/cassandra/streaming/StreamLockfile.java
+++ b/src/java/org/apache/cassandra/streaming/StreamLockfile.java
@@ -88,7 +88,7 @@ public class StreamLockfile
         }
         catch (IOException e)
         {
-            logger.warn("Could not create lockfile {} for stream session, nothing to worry too much about", SafeArg.of("lockfile", lockfile.toString()), e);
+            logger.warn("Could not create lockfile {} for stream session, nothing to worry too much about", UnsafeArg.of("lockfile", lockfile.toString()), e);
         }
     }
 
--- a/src/java/org/apache/cassandra/streaming/StreamLockfile.java
+++ b/src/java/org/apache/cassandra/streaming/StreamLockfile.java
@@ -30,6 +30,7 @@ import java.util.UUID;
 import com.google.common.base.Charsets;
 
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.UnsafeArg;
 import org.apache.cassandra.io.sstable.format.SSTableWriter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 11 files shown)


src/java/org/apache/cassandra/streaming/StreamLockfile.java
```diff
@@ -107,7 +109,7 @@ public void cleanup()
             catch (Exception e)
             {
                 JVMStabilityInspector.inspectThrowable(e);
-                logger.warn("failed to delete a potentially stale sstable {}", file);
+                logger.warn("failed to delete a potentially stale sstable {}", SafeArg.of("file", file));
```
wi11dey: is the full file safeloggable ?

src/java/org/apache/cassandra/streaming/StreamWriter.java
```diff
@@ -75,8 +76,13 @@ public StreamWriter(SSTableReader sstable, Collection<Pair<Long, Long>> sections
     public void write(DataOutputStreamPlus output) throws IOException
     {
         long totalSize = totalSize();
-        logger.debug("[Stream #{}] Start streaming file {} to {}, repairedAt = {}, totalSize = {}", session.planId(),
-                     sstable.getFilename(), session.peer, sstable.getSSTableMetadata().repairedAt, totalSize);
+        if (logger.isDebugEnabled())
+            logger.debug("[Stream #{}] Start streaming file {} to {}, repairedAt = {}, totalSize = {}",
+                         SafeArg.of("planId", session.planId()),
+                         SafeArg.of("file", sstable.getFilename()),
```
wi11dey: also would like to doublecheck this is safe

rhuffy: sstable file names are safe right?

src/java/org/apache/cassandra/streaming/StreamLockfile.java
rhuffy: from reading the above code, it appears that each line of the `lockfile` is a filename, called `file` that is passed to `Descriptor.fromFilename`.

Assuming that all sstable filenames are safe, then this log statement is safe.

src/java/org/apache/cassandra/streaming/StreamWriter.java
wi11dey: yeah the filename is safe as long as it doesn't include the absolute path

rhuffy: I feel like this could include the absolute path, so I'll mark it (and equivalents) as unsafe

src/java/org/apache/cassandra/streaming/StreamLockfile.java
rhuffy: switched to UnsafeArg

src/java/org/apache/cassandra/streaming/StreamWriter.java
rhuffy: switched to UnsafeArg

src/java/org/apache/cassandra/streaming/StreamLockfile.java
```diff
@@ -85,7 +88,7 @@ this class will not need to clean up tmp files (on restart), CassandraDaemon doe
         }
         catch (IOException e)
         {
-            logger.warn(String.format("Could not create lockfile %s for stream session, nothing to worry too much about", lockfile), e);
+            logger.warn("Could not create lockfile {} for stream session, nothing to worry too much about", SafeArg.of("lockfile", lockfile.toString()), e);
```
wi11dey: file marked as safe, inconsistent with the other files that i believe we've decided are unsafe

src/java/org/apache/cassandra/streaming/compress/CompressedStreamWriter.java
```diff
@@ -96,8 +102,13 @@ public Long apply(WritableByteChannel wbc)
                     session.progress(sstable.descriptor, ProgressInfo.Direction.OUT, progress, totalSize);
                 }
             }
-            logger.debug("[Stream #{}] Finished streaming file {} to {}, bytesTransferred = {}, totalSize = {}",
-                         session.planId(), sstable.getFilename(), session.peer, progress, totalSize);
+            if (logger.isDebugEnabled())
+                logger.debug("[Stream #{}] Finished streaming file {} to {}, bytesTransferred = {}, totalSize = {}",
+                             SafeArg.of("planId", session.planId()),
+                             SafeArg.of("file", sstable.getFilename()),
```
wi11dey: missed one

src/java/org/apache/cassandra/streaming/compress/CompressedStreamWriter.java
```diff
@@ -59,8 +60,13 @@ public CompressedStreamWriter(SSTableReader sstable, Collection<Pair<Long, Long>
     public void write(DataOutputStreamPlus out) throws IOException
     {
         long totalSize = totalSize();
-        logger.debug("[Stream #{}] Start streaming file {} to {}, repairedAt = {}, totalSize = {}", session.planId(),
-                     sstable.getFilename(), session.peer, sstable.getSSTableMetadata().repairedAt, totalSize);
+        if (logger.isDebugEnabled())
+            logger.debug("[Stream #{}] Start streaming file {} to {}, repairedAt = {}, totalSize = {}",
+                         SafeArg.of("planId", session.planId()),
+                         SafeArg.of("file", sstable.getFilename()),
```
wi11dey: another

rhuffy:
should be fixed now

wi11dey (approved):

PR #593, opened 2024-12-05
migrate to SafeLogging in `org.apache.cassandra.gms`

rhuffy:
Mark logging arguments as Safe or Unsafe in org.apache.cassandra.gms.

This PR also wraps all calls to logger.debug() in an if (logger.isDebugEnabled()) check.

For reviewers, please confirm that you agree with my choices for SafeArgs.

### DIFF
--- a/src/java/org/apache/cassandra/gms/FailureDetector.java
+++ b/src/java/org/apache/cassandra/gms/FailureDetector.java
@@ -35,6 +35,7 @@ import com.codahale.metrics.UniformSnapshot;
 import com.google.common.collect.ImmutableMap;
 import com.palantir.cassandra.db.BootstrappingSafetyException;
 import com.palantir.cassandra.metrics.FailureDetectorMetrics;
+import com.palantir.logsafe.SafeArg;
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.io.FSWriteError;
 import org.apache.cassandra.io.util.FileUtils;
@@ -68,7 +69,7 @@ public class FailureDetector implements IFailureDetector, FailureDetectorMBean
         if (System.getProperty("cassandra.max_local_pause_in_ms") != null)
         {
             long pause = Long.parseLong(System.getProperty("cassandra.max_local_pause_in_ms"));
-            logger.warn("Overriding max local pause time to {} ms", pause);
+            logger.warn("Overriding max local pause time to {} ms", SafeArg.of("timeInMs", pause));
             return pause * 1000000L;
         }
         else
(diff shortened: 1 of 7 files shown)


src/java/org/apache/cassandra/gms/FailureDetector.java
```diff
@@ -329,31 +332,37 @@ public void interpret(InetAddress ep)
...
         else if (logger.isDebugEnabled() && (PHI_FACTOR * phi * DEBUG_PERCENTAGE / 100.0 > getPhiConvictThreshold()))
         {
-            logger.debug("PHI for {} : {}", ep, phi);
+            logger.debug("PHI for {} : {}", SafeArg.of("endpoint", ep), SafeArg.of("phi", phi));
         }
         else if (logger.isTraceEnabled())
         {
-            logger.trace("PHI for {} : {}", ep, phi);
-            logger.trace("mean for {} : {}", ep, hbWnd.mean());
+            logger.trace("PHI for {} : {}", SafeArg.of("endpoint", ep), SafeArg.of("phi", phi));
```
andybradshaw, line 357: Not sure if java loggers work this way, but does `trace` imply `debug`? If so, we are double logging the PHI values here

d-guo: this is fine since we're in a `else if` right?

d-guo:
arguments look good to me!

there's a lot of `logger.isDebugEnabled()` and `logger.isTraceEnabled()` can we refactor into a `maybeDebugLog(logger, message, safeArgs)` and similar

d-guo:
also wondering if it's wrong to just maintain previous behavior and not include any new `if(logger.isXEnabled)`s

src/java/org/apache/cassandra/gms/FailureDetector.java
rhuffy: Trace != Debug. Trace logs are meant to be even less frequently enabled, and are often very noisy in this project.

andybradshaw:
> there's a lot of `logger.isDebugEnabled()` and `logger.isTraceEnabled()` can we refactor into a `maybeDebugLog(logger, message, safeArgs)` and similar

I believe the problem here is that we want to avoid the allocation of the SafeArg, since these may be on fairly hot paths, and there's not a cleaner/clearer way to do lazy allocation of these objects.

src/java/org/apache/cassandra/gms/FailureDetector.java
andybradshaw: Basically, I think the concern is that levels are inclusive of the levels above them (e.g. if I have `trace` logging on for a class, I also expect to get `debug` logging -> `isDebugEnabled` will return true and we'd never hit the trace branch here). I could be wrong about that... can write a little dummy test to confirm.I guess the other bits within the `ifs` are also relevant here... so maybe we only care about these in certain scenarios. 

That being said, we do [unconditionally log PHI](https://github.com/palantir/cassandra/pull/593/files#diff-cdf19ad177363b2c78abefb71ddb877164ad5fe0edec7fad0f622aba0e9541dcR335) when trace is enabled anyways, so can probably remove this one?

rhuffy: oh, sorry I misunderstood your question. I do believe that if you set TRACE, that also does set DEBUG

andybradshaw (approved):
Approving, since I don't think we want to actually change too much of the logging here, just add safe args to the existing logging.

PR #594, opened 2024-12-07
Remove endpoints from schema pulls if other node is down

andybradshaw:


zpear (approved):

src/java/org/apache/cassandra/service/MigrationTask.java
```diff
@@ -124,6 +126,12 @@ public boolean isLatencyForSnitch()
                 return false;
             }
         };
-       MessagingService.instance().sendRRWithFailure(message, endpoint, cb);
+        try {
+            MessagingService.instance().sendRRWithFailure(message, endpoint, cb);
+        }
+        catch (Exception e)
+        {
+            version.ifPresent(v -> MigrationManager.removeEndpointFromSchemaPullVersion(v, endpoint));
```
inespot, line 134: Do we still wanna throw the original exception here to retain existing behavior?

PR #597, opened 2024-12-09
Make SSTableExport Return Instead of Exit

d-guo:
When using SSTableExport via PCQL, we want the main program to return instead of `Exit(1)` so that we can handle any exceptions and don't exit the shell.


wi11dey:
can we move the implementation into a static method that returns `int` ? then PCQL can call that and not exit and the main method here can still exit with the right error code instead of a success error code by default

wi11dey (dismissed):

wi11dey (approved):

PR #598, opened 2024-12-09
Add guard to setMode for NTEs

d-guo:
When a node enters non-transient error, it should never regain a normal operation mode (hence non-transient). This PR adds a guard for this and additionally short-circuits the bootstrap when the node enters NTE. This addition is necessary because the bootstrap NTE can occur from different threads than the one calling `bootstrapStream.get()`.

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -1616,9 +1616,9 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                     logger.info("Stopped waiting for bootstrap streaming to complete because detected a bootstrap error.", SafeArg.of("nonTransientErrors", getNonTransientErrors()));
                     break;
                 }
-                bootstrapStream.cancel(true);
                 Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
             }
+            bootstrapStream.cancel(true);
             isBootstrapMode = false;
             return !StorageService.instance.hasNonTransientError(StorageServiceMBean.NonTransientError.BOOTSTRAP_ERROR);
         }
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -1615,6 +1615,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                     if (hasNonTransientError(NonTransientError.BOOTSTRAP_ERROR))
                     {
                         logger.info("Aborting waiting for bootstrap streaming to complete because detected a bootstrap error.", SafeArg.of("nonTransientErrors", getNonTransientErrors()));
+                        bootstrapStream.cancel(true);
                         break;
                     }
                 }
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -1544,11 +1544,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     void setMode(Mode m, @Safe String msg, boolean log)
     {
         if (operationMode == Mode.NON_TRANSIENT_ERROR && m != Mode.NON_TRANSIENT_ERROR) {
-            if (log)
-                logger.warn("Attempted to change mode from NTE to non-NTE", SafeArg.of("attemptedSetMode", m), SafeArg.of("msg", msg));
-            else
-                logger.debug("Attempted to change mode from NTE to non-NTE", SafeArg.of("attemptedSetMode", m), SafeArg.of("msg", msg));
-            return;
+            logger.warn("Attempted to change mode from NTE to non-NTE", SafeArg.of("attemptedSetMode", m), SafeArg.of("msg", msg));
         }
         operationMode = m;
         if (log)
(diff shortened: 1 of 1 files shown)


rhuffy:
please fix the formatting to match the Cassandra style. You should be able to use the auto formatter.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1543,6 +1543,13 @@ private void setMode(Mode m, boolean log)
     @VisibleForTesting
     void setMode(Mode m, @Safe String msg, boolean log)
     {
+        if (operationMode == Mode.NON_TRANSIENT_ERROR && m != Mode.NON_TRANSIENT_ERROR) {
+            if (log)
+                logger.warn("Attempted to change mode from NTE to non-NTE", SafeArg.of("attemptedSetMode", m), SafeArg.of("msg", msg));
+            else
+                logger.debug("Attempted to change mode from NTE to non-NTE", SafeArg.of("attemptedSetMode", m), SafeArg.of("msg", msg));
+            return;
```
inespot, line 1549: Simplify: log with warn no matter the value of `log`, I think this is always worth a warning log

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1599,9 +1606,19 @@ private boolean bootstrap(final Collection<Token> tokens)
...
+                    bootstrapStream.get();
+                    logger.info("Bootstrap streaming completed for tokens {}", tokens);
+                    break;
+                }
+                if (hasNonTransientError(NonTransientError.BOOTSTRAP_ERROR)) {
+                    logger.info("Stopped waiting for bootstrap streaming to complete because detected a bootstrap error.", SafeArg.of("nonTransientErrors", getNonTransientErrors()));
+                    break;
+                }
+                Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
+            }
```
inespot, line 1622: You might be able to do less aggressive polling while keeping the reactivity by doing something like
```
while (true) {
    try {
        bootstrapStream.get(timeBetweenCheck, TimeUnit.SECONDS);
        logger.info("Bootstrap streaming completed for tokens {}", tokens);
        break;
    } catch (TimeoutException e) {
        if (hasNonTransientError(NonTransientError.BOOTSTRAP_ERROR)) {
            logger.info("Aborting waiting for bootstrap streaming to complete because detected a bootstrap error.",
                    SafeArg.of("nonTransientErrors", getNonTransientErrors()));
            break;
        }
    } catch (Exception e) {
        logger.error("Error while waiting for bootstrap stream to complete", e);
        break;
    }
}
```
If we stop waiting for the streams, is there a way to also notify peers/shut down stream early? Or will the following `disableNode` handle this

d-guo: this just blocks on `bootstrapStream.get`, right? like the `while(true)` doesn't do anything here

d-guo: hm, why is this less aggressive polling? if possible, i'd like to avoid using exceptions for control flow and i think i don't recognize the benefit here

d-guo: > notify peers/shut down stream early

i'll test this tomorrow to see if `disableNode` kills the streams as we would hope. with operator, the node will be deleted after entering NTE anyway though

d-guo: `disableNode` will not kill the streams. i just realized that in the original instance that this happened, the streams continued even after the NTE blip

d-guo: confirmed that `bootstrapStream.cancel(true);` throws an interrupt on the socket that does close it. see below error message on node sending streams to the joining node when the joining node cancels the bootstrap
```
ERROR [2025-01-09T04:00:21.648599Z] org.apache.cassandra.streaming.StreamSession: [Stream #bc461090-ce3d-11ef-ab9d-314016c85ad1] Streaming error occurred (0: bc461090-ce3d-11ef-ab9d-314016c85ad1)
java.net.SocketException: End-of-stream reached
```
grafana also shows the stream progress plateau

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1543,6 +1543,10 @@ private void setMode(Mode m, boolean log)
     @VisibleForTesting
     void setMode(Mode m, @Safe String msg, boolean log)
     {
+        if (operationMode == Mode.NON_TRANSIENT_ERROR && m != Mode.NON_TRANSIENT_ERROR)
+        {
+            logger.warn("Attempted to change mode from NTE to non-NTE", SafeArg.of("attemptedSetMode", m), SafeArg.of("msg", msg));
```
inespot, line 1548: Do we not want to return here?

d-guo: yes we do, good catch

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1599,9 +1604,23 @@ private boolean bootstrap(final Collection<Token> tokens)
         ListenableFuture<StreamState> bootstrapStream = bootstrapper.bootstrap(streamStateStore, !replacing && useStrictConsistency); // handles token update
         try
         {
-            bootstrapStream.get();
+            while (true)
+            {
+                try {
+                    bootstrapStream.get(5, MINUTES);
+                    logger.info("Bootstrap streaming completed for tokens {}", tokens);
```
inespot: Will this not be marked as unsafe by default? If we don't add the SafeArg wrapper

d-guo: yeah i just copied the log line from before. at some point we'll need to fix all the safe args for this class so i didn't bother doing it here

inespot (approved):

PR #603, opened 2025-01-15
Add tracing-java integration

jkozlowski:
Diff is large cause generated code is checked in.

Log
* Cannot upgrade tracing-java, because newer version are compiled with newer Java.
* This whole setup is VERY SHAKEY, because we compile cassandra here with old tracing, but run internally with new tracing, possibly. I still don't quite understand what the actual jar on the classpath would be.

To mitigate these, I am trying to use as old and low level API as I can.

* thrift: somehow managed to compile old 0.17 thrift to be able to regenerate the interfaces. There is no way to insert middleware in TBinary thrift transport.


zpear (dismissed):
rc

PR #604, opened 2025-01-16
Add safe logging arguments to StorageService

andybradshaw:


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1107,15 +1110,17 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
...
             }
         }
         else
         {
             if (dataAvailable)
                 logger.info("Startup complete, but write survey mode is active, not becoming an active ring member. Use JMX (StorageService->joinRing()) to finalize ring joining.");
             else
-                logger.warn("Some data streaming failed. Use nodetool to check bootstrap state and resume. For more, see `nodetool help bootstrap`. {}", SystemKeyspace.getBootstrapState());
+                logger.warn("Some data streaming failed. Use nodetool to check bootstrap state and resume. For more, see `nodetool help bootstrap`. {}",
+                            SafeArg.of("", SystemKeyspace.getBootstrapState()));
```
rhuffy: arg name here?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -3779,7 +3807,7 @@ public int forceRepairAsync(String keyspace, RepairOption options, boolean legac
         if (!inProgressCommand.isPresent()) {
             new Thread(createRepairTask(cmd, arguments, legacy)).start();
         } else {
-            logger.info("Combining new repair request with in-progress (identical) repair command #{}", cmd);
+            logger.info("Combining new repair request with in-progress (identical) repair command #{}", UnsafeArg.of("command", cmd));
```
rhuffy: cmd is an int, so this is probably safe, right?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4441,12 +4480,16 @@ private void calculateToFromStreams(Collection<Token> newTokens, List<String> ke
...
+                        logger.debug("Will request range {} of keyspace {} from endpoint {}",
+                                     SafeArg.of("range", workMap.get(address)),
+                                     SafeArg.of("keyspace", keyspace),
+                                     SafeArg.of("endpoint", address));
                         InetAddress preferred = SystemKeyspace.getPreferredIP(address);
                         streamPlan.requestRanges(address, preferred, keyspace, workMap.get(address));
                     }
 
-                    logger.debug("Keyspace {}: work map {}.", keyspace, workMap);
+                    // This log line seems unnecessary given the above loop...
```
rhuffy: maybe, but I do think in grafana it's easier to analyze a single log line instead of combining multiple, especially when multiple nodes are reporting the logs simultaneously

rhuffy (dismissed):
left some small nits, otherwise LGTM

src/java/org/apache/cassandra/service/StorageService.java
andybradshaw: Done

src/java/org/apache/cassandra/service/StorageService.java
andybradshaw: Fair enough... didn't check the type, assumed this was the literal command.

src/java/org/apache/cassandra/service/StorageService.java
andybradshaw: This is true as long as `workMap` doesn't get so large that we can't actually ingest the log. I don't have any evidence to suggest that's the case, but I don't know that I necessarily agree with the "single log > multi log" statement... filtering on tags like `host` exists for a reason 😛

andybradshaw:
👍

rhuffy (approved):

PR #607, opened 2025-01-27
Add Gate for Requesting Streams During Bootstrap

d-guo:
Add Gate for Requesting Streams During Bootstrap

Part of changes in https://palantir.quip.com/HUarApyalaJa#ZSCABAYrigq

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -125,7 +125,8 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     private final List<ProgressListener> bootstrapListeners = new CopyOnWriteArrayList<>();
 
     private final Condition startBootstrapCondition = new SimpleCondition(DISABLE_WAIT_TO_BOOTSTRAP);
-    // TODO(dguo): use startRequestStreamsCondition to gate decommissions as well with `.reset()`
+    // TODO(dguo): Use startRequestStreamsCondition to gate decommissions as well.
+    //  This requires SimpleCondition to synchronize between resetting and adding new waiters or using a new Condition altogether.
     private final Condition startRequestStreamsCondition = new SimpleCondition(DISABLE_WAIT_TO_REQUEST_STREAMS);
     private final Condition finishBootstrapCondition = new SimpleCondition(DISABLE_WAIT_TO_FINISH_BOOTSTRAP);
 
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -112,7 +112,6 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     private static final boolean DISABLE_WAIT_TO_BOOTSTRAP = Boolean.getBoolean("palantir_cassandra.disable_wait_to_bootstrap");
     private static final boolean DISABLE_WAIT_TO_REQUEST_STREAMS = Boolean.getBoolean("palantir_cassandra.disable_wait_to_request_streams");
     private static final boolean DISABLE_WAIT_TO_FINISH_BOOTSTRAP = Boolean.getBoolean("palantir_cassandra.disable_wait_to_finish_bootstrap");
-    private static final Integer BOOTSTRAP_DISK_USAGE_THRESHOLD = Integer.getInteger("palantir_cassandra.bootstrap_disk_usage_threshold_percentage");
     private static final Integer STREAMS_REQUEST_SAFETY_CHECK_GRACE_PERIOD_MINUTES = Integer.getInteger("palantir_cassandra.streams_request_safety_check_grace_period_minutes", 30);
     private static final Integer BOOTSTRAP_SAFETY_CHECK_GRACE_PERIOD_MINUTES = Integer.getInteger("palantir_cassandra.bootstrap_safety_check_grace_period_minutes", 30);
 
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -946,7 +946,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
             }
             else
             {
-                SystemKeyspace.setBootstrapState(SystemKeyspace.BootstrapState.IN_PROGRESS);
+                setBootstrapState(SystemKeyspace.BootstrapState.IN_PROGRESS);
             }
             setMode(Mode.JOINING, "waiting for ring information", true);
             // first sleep the delay to make sure we see all our peers
@@ -1267,7 +1267,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     {
         // start participating in the ring.
         logger.info("Attempting to set bootstrap state to COMPLETED and to join token ring");
-        SystemKeyspace.setBootstrapState(SystemKeyspace.BootstrapState.COMPLETED);
+        setBootstrapState(SystemKeyspace.BootstrapState.COMPLETED);
         setTokens(tokens);
 
         assert tokenMetadata.sortedTokens().size() > 0;
@@ -4014,6 +4014,11 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
         return liveEps;
     }
 
+    @VisibleForTesting
+    void setBootstrapState(SystemKeyspace.BootstrapState bootstrapState) {
+        SystemKeyspace.setBootstrapState(bootstrapState);
+    }
+
     public void setLoggingLevel(String classQualifier, String rawLevel) throws Exception
     {
         ch.qos.logback.classic.Logger logBackLogger = (ch.qos.logback.classic.Logger) LoggerFactory.getLogger(classQualifier);
@@ -4167,7 +4172,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
 
     private void leaveRing()
     {
-        SystemKeyspace.setBootstrapState(SystemKeyspace.BootstrapState.NEEDS_BOOTSTRAP);
+        setBootstrapState(SystemKeyspace.BootstrapState.NEEDS_BOOTSTRAP);
         tokenMetadata.removeEndpoint(FBUtilities.getBroadcastAddress());
         PendingRangeCalculatorService.instance.update();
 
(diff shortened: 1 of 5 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -201,7 +204,7 @@ public Collection<Range<Token>> getPrimaryRangesWithinDC(String keyspace)
     private double traceProbability = 0.0;
 
     @VisibleForTesting
-    static enum Mode { STARTING, NORMAL, JOINING, LEAVING, DECOMMISSIONED, MOVING, DRAINING, DRAINED, ZOMBIE, NON_TRANSIENT_ERROR, TRANSIENT_ERROR, WAITING_TO_BOOTSTRAP, WAITING_TO_FINISH_BOOTSTRAP, DISABLED }
+    static enum Mode { STARTING, NORMAL, JOINING, LEAVING, DECOMMISSIONED, MOVING, DRAINING, DRAINED, ZOMBIE, NON_TRANSIENT_ERROR, TRANSIENT_ERROR, WAITING_TO_BOOTSTRAP, WAITING_TO_REQUEST_STREAMS, WAITING_TO_FINISH_BOOTSTRAP, DISABLED }
```
inespot, line 208: We should also add this new mode to the `FailureDetector::safeguardBootstrapTimeout`safety check

inespot: And `StorageService::takeSnapshot` and `StorageService::takeColumnFamilySnapshot` and `StorageService::takeMultipleColumnFamilySnapshot` ... etc

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -110,8 +110,10 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
 {
     private static final Logger logger = LoggerFactory.getLogger(StorageService.class);
     private static final boolean DISABLE_WAIT_TO_BOOTSTRAP = Boolean.getBoolean("palantir_cassandra.disable_wait_to_bootstrap");
+    private static final boolean DISABLE_WAIT_TO_REQUEST_STREAMS = Boolean.getBoolean("palantir_cassandra.disable_wait_to_request_streams");
     private static final boolean DISABLE_WAIT_TO_FINISH_BOOTSTRAP = Boolean.getBoolean("palantir_cassandra.disable_wait_to_finish_bootstrap");
     private static final Integer BOOTSTRAP_DISK_USAGE_THRESHOLD = Integer.getInteger("palantir_cassandra.bootstrap_disk_usage_threshold_percentage");
```
inespot: Is this used anywhere?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1586,6 +1589,22 @@ private boolean bootstrap(final Collection<Token> tokens)
             SystemKeyspace.resetAvailableRanges();
         }
 
+        try
+        {
+            setMode(Mode.WAITING_TO_REQUEST_STREAMS, "Awaiting call to proceed with requesting streams during bootstrap", true);
```
inespot, line 1594: Nitty nit: `WAITING_TO_REQUEST_BOOTSTRAP_STREAMS` or sth similar? Since theoretically there are streams in other pieces of the code

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -124,6 +126,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     private final List<ProgressListener> bootstrapListeners = new CopyOnWriteArrayList<>();
 
     private final Condition startBootstrapCondition = new SimpleCondition(DISABLE_WAIT_TO_BOOTSTRAP);
+    private final Condition startRequestStreamsCondition = new SimpleCondition(DISABLE_WAIT_TO_REQUEST_STREAMS);
```
inespot, line 130: Maybe a separate PR: I think we need to add this gate for decommissions as well

src/java/org/apache/cassandra/service/StorageService.java
d-guo: don't think so — looks like the stuff (https://github.com/palantir/cassandra/commit/a4376828cd5c7a7936d359961e6d23cc4ec7f3c1) flexing it was removed at some point. i'll remove it

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1586,6 +1588,22 @@ private boolean bootstrap(final Collection<Token> tokens)
...
+        try
+        {
+            setMode(Mode.WAITING_TO_REQUEST_STREAMS, "Awaiting call to proceed with requesting streams during bootstrap", true);
+            boolean timeoutExceeded = !startRequestStreamsCondition.await(STREAMS_REQUEST_SAFETY_CHECK_GRACE_PERIOD_MINUTES, MINUTES);
+            if (timeoutExceeded)
+            {
+                logger.error("Start signal to request streams was not given within 30 minutes. Streams request safety check failed.");
+                recordBootstrapErrorAndThrow("streamsRequestSafetyCheckFailed");
+            }
+        }
```
inespot, line 1601: Could we add a unit test in `StorageServiceServerTest` to ensure we are already gossiping our tokens and JOINING application state at that point of the gate?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1695,6 +1713,12 @@ public void startBootstrap()
         startBootstrapCondition.signalAll();
     }
 
+    @Override
+    public void startRequestingStreams()
+    {
+        startRequestStreamsCondition.signalAll();
```
inespot, line 1725: If we end up using the same Operation Mode + condition for decommissions, I think that should be `.signal()` instead of `signalAll()`

inespot: Ah nvm, `SimpleCondition` does not support that. So either we recreate the startRequestStreamsCondition object after this function is called once or we use another object type than `SimpleCondition`

d-guo: yeah imo we can implement `reset` inside `SimpleCondition` and re-use, but that detail is probably for another pr

test/unit/org/apache/cassandra/service/StorageServiceServerTest.java
```diff
@@ -152,6 +158,40 @@ public void testRegularMode() throws ConfigurationException
...
 
+    @Test
+    public void testGossipStateAtGateToRequestStreams() throws ConfigurationException, UnknownHostException
+    {
+        SchemaLoader.mkdirs();
+        SchemaLoader.cleanup();
+        StorageService instance = spy(StorageService.instance);
+        doReturn(true).when(instance).shouldBootstrap(anyBoolean());
+        doNothing().when(instance).checkGossiperSeeds();
+        Thread thread = new Thread(new Runnable()
```
inespot: I think Java8 actually does support lambda functions?
```
new Thread(() -> instance.initServer(0));
```

test/unit/org/apache/cassandra/service/StorageServiceServerTest.java
```diff
@@ -152,6 +158,40 @@ public void testRegularMode() throws ConfigurationException
         StorageService.instance.stopClient();
     }
 
+    @Test
+    public void testGossipStateAtGateToRequestStreams() throws ConfigurationException, UnknownHostException
```
inespot: Cool test!

src/java/org/apache/cassandra/service/StorageService.java
inespot: Can we actually add the reset statement now with a small comment? To avoid any oversight when we do the decommission PR

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1627,6 +1646,12 @@ private boolean bootstrap(final Collection<Token> tokens)
         }
     }
 
+    @VisibleForTesting
+    void checkGossiperSeeds() {
```
inespot: Do we still need those VisibleForTesting changes?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -124,6 +125,8 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     private final List<ProgressListener> bootstrapListeners = new CopyOnWriteArrayList<>();
 
     private final Condition startBootstrapCondition = new SimpleCondition(DISABLE_WAIT_TO_BOOTSTRAP);
+    // we expect to use startRequestStreamsCondition to gate decommissions as well with `.reset()`
```
inespot: Nit: `// TODO(dguo): use startRequestStreamsCondition to gate ...`
for easier search

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1695,6 +1720,12 @@ public void startBootstrap()
         startBootstrapCondition.signalAll();
     }
 
+    @Override
+    public void startRequestingStreams()
+    {
+        startRequestStreamsCondition.signalAll();
+    }
```
inespot, line 1726: Call reset() before returning?

src/java/org/apache/cassandra/utils/concurrent/SimpleCondition.java
```diff
@@ -101,4 +105,9 @@ public boolean awaitUntil(Date deadline) throws InterruptedException
     {
         throw new UnsupportedOperationException();
     }
+
+    public void reset() {
+        signaled = disabled;
+        waitingUpdater.set(this, null);
```
inespot: `waitingUpdater.compareAndSet(this, null, new WaitQueue());` ?
+ throw if `waiting.hasWaiters()`

src/java/org/apache/cassandra/utils/concurrent/SimpleCondition.java
```diff
@@ -30,12 +30,16 @@ public class SimpleCondition implements Condition
     private static final AtomicReferenceFieldUpdater<SimpleCondition, WaitQueue> waitingUpdater = AtomicReferenceFieldUpdater.newUpdater(SimpleCondition.class, WaitQueue.class, "waiting");
 
     private volatile WaitQueue waiting;
+    private final boolean disabled;
```
inespot: Nit: unused value, we can probably remove as well for now

inespot (dismissed):

inespot (approved):

PR #609, opened 2025-01-29
Fix effective ownership in Nodetool Status and Ring

rhuffy:
Nodetool Status and Ring can display effective ownership information if a keyspace is provided, or if the non-system keyspaces all use the same replication strategy.

In our case, we can assume that the `system_palantir` keyspace is always present, and use its replication strategy by default, if present.

### DIFF
--- a/src/java/com/palantir/cassandra/db/SystemPalantir.java
+++ b/src/java/com/palantir/cassandra/db/SystemPalantir.java
@@ -1,0 +1,28 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.cassandra.db;
+
+final public class SystemPalantir
+{
+    public static final String NAME = "system_palantir";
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4900,8 +4901,11 @@ public LinkedHashMap<InetAddress, Float> effectiveOwnership(String keyspace) thr
         else
         {
             List<String> userKeyspaces = Schema.instance.getUserKeyspaces();
-
-            if (userKeyspaces.size() > 0)
+            if (Schema.instance.getKeyspaceInstance(SystemPalantir.NAME) != null)
+            {
+                keyspace = SystemPalantir.NAME;
+            }
```
inespot, line 4907: Instead of modifying it here (I feel it might be confusing if we use this functions in other pieces of the code), wanna add to  the`Status` class instead: `ownerships = probe.effectiveOwnership(keyspace != null ? keyspace: SystemPalantir.NAME);` and add a small comment on why we are doing this.

src/java/com/palantir/cassandra/db/SystemPalantir.java
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.cassandra.db;
+
+final public class SystemPalantir
```
inespot, line 21: Nice!

src/java/org/apache/cassandra/service/StorageService.java
rhuffy: I think this is the correct place for this logic because this method already contains logic to auto resolve a keyspace if none is provided

rhuffy: Also, adding the system_palantir keyspace name as the argument to `effectiveOwnership` isn't great because then the method would throw in the case that system_palantir is not present, whereas in my patch it would instead fall back to the existing behavior.

inespot (approved):
Provided sanity tested on dev stack :)
`effectiveOwnership` is also only used for nodetool ring and nodetool status so the scope of this change is quite safe

PR #611, opened 2025-02-03
Safe logging for CassandraDaemon

fsxfreak:
adding `[un]safearg` annotations on every `logger.` call that I saw in the file. Followed the "what's safe to log" guidelines to figure out which ones are unsafe/safe, let me know if any of them should be changed.

### DIFF
--- a/src/java/org/apache/cassandra/service/CassandraDaemon.java
+++ b/src/java/org/apache/cassandra/service/CassandraDaemon.java
@@ -324,7 +324,7 @@ public class CassandraDaemon
         catch (Throwable t)
         {
             JVMStabilityInspector.inspectThrowable(t);
-            logger.warn("Error loading key or row cache", UnsafeArg.of("loadRowAndKeyCacheError", t));
+            logger.warn("Error loading key or row cache", t);
         }
 
         try
@@ -352,7 +352,7 @@ public class CassandraDaemon
                 && StorageService.instance.hasNonTransientError(StorageServiceMBean.NonTransientError.COMMIT_LOG_CORRUPTION))
             {
                 logger.error("Failed to recover from commitlog corruption due to some non transient errors: {}",
-                             UnsafeArg.of("commitLogRecoverError", StorageService.instance.getNonTransientErrors()));
+                             UnsafeArg.of("commitLogRecoverErrors", StorageService.instance.getNonTransientErrors()));
                 return;
             }
             throw new RuntimeException(e);
@@ -370,12 +370,12 @@ public class CassandraDaemon
         }
         catch (BootstrappingSafetyException e)
         {
-            logger.error("Non-fatal bootstrap error. Server will continue but is disabled and without metrics.", UnsafeArg.of("storageServiceInitError", e));
+            logger.error("Non-fatal bootstrap error. Server will continue but is disabled and without metrics.", e);
             return;
         }
         catch (ConfigurationException e)
         {
-            logger.error("Fatal configuration error; unable to start server.  See log for stacktrace.", UnsafeArg.of("storageServiceInitError", e));
+            logger.error("Fatal configuration error; unable to start server.  See log for stacktrace.", e);
             exitOrFail(1, "Fatal configuration error", e);
         }
 
@@ -399,7 +399,7 @@ public class CassandraDaemon
             }
             catch (Exception e)
             {
-                logger.warn("Failed to load metrics-reporter-config, metric sinks will not be activated", UnsafeArg.of("metricsReporterConfigException", e));
+                logger.warn("Failed to load metrics-reporter-config, metric sinks will not be activated", e);
             }
         }
         new HiccupMeter().start();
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -323,7 +324,7 @@ private void completeSetupMayThrowSstableException() {
         catch (Throwable t)
         {
             JVMStabilityInspector.inspectThrowable(t);
-            logger.warn("Error loading key or row cache", t);
+            logger.warn("Error loading key or row cache", UnsafeArg.of("loadRowAndKeyCacheError", t));
```
andybradshaw: I _believe_ we don't need to do anything for `Throwable`s... they should already correctly log the `stacktrace` part of the sls logging spec.

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -351,7 +352,7 @@ private void recoverCommitlogAndCompleteSetup() {
                 && StorageService.instance.hasNonTransientError(StorageServiceMBean.NonTransientError.COMMIT_LOG_CORRUPTION))
             {
                 logger.error("Failed to recover from commitlog corruption due to some non transient errors: {}",
-                             StorageService.instance.getNonTransientErrors());
+                             UnsafeArg.of("commitLogRecoverError", StorageService.instance.getNonTransientErrors()));
```
andybradshaw: nit: should maybe be an optional plural, since I think this can contain multiple errors

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -369,12 +370,12 @@ private void recoverCommitlogAndCompleteSetup() {
         }
         catch (BootstrappingSafetyException e)
         {
-            logger.error("Non-fatal bootstrap error. Server will continue but is disabled and without metrics.", e);
+            logger.error("Non-fatal bootstrap error. Server will continue but is disabled and without metrics.", UnsafeArg.of("storageServiceInitError", e));
```
andybradshaw: same as other throwable

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -369,12 +370,12 @@ private void recoverCommitlogAndCompleteSetup() {
...
             return;
         }
         catch (ConfigurationException e)
         {
-            logger.error("Fatal configuration error; unable to start server.  See log for stacktrace.", e);
+            logger.error("Fatal configuration error; unable to start server.  See log for stacktrace.", UnsafeArg.of("storageServiceInitError", e));
```
andybradshaw: same as other throwable

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -398,7 +399,7 @@ private void recoverCommitlogAndCompleteSetup() {
             }
             catch (Exception e)
             {
-                logger.warn("Failed to load metrics-reporter-config, metric sinks will not be activated", e);
+                logger.warn("Failed to load metrics-reporter-config, metric sinks will not be activated", UnsafeArg.of("metricsReporterConfigException", e));
```
andybradshaw: same as other throwable

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -453,12 +454,14 @@ public void startNativeTransport()
             throw new IllegalStateException("native transport should be set up before it can be started");
 
         nativeServer.start();
-        logger.info("Native server running on {}", new InetSocketAddress(DatabaseDescriptor.getRpcAddress(), DatabaseDescriptor.getNativeTransportPort()));
+        InetSocketAddress nativeAddress = new InetSocketAddress(DatabaseDescriptor.getRpcAddress(), DatabaseDescriptor.getNativeTransportPort());
+        logger.info("Native server running on {}", SafeArg.of("nativeAddress", nativeAddress));
```
andybradshaw: Think it might be better to split the parameters here actually. It doesn't matter much here, but both this and the log line below can both use `address` and `port` for param name consistency, as the message already differentiates native vs. thrift.

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -816,7 +827,7 @@ private void exitOrFail(int code, String message, Throwable cause) {
                 throw t;
             }
             else {
-                logger.error(message, cause);
+                logger.error("{} {}", UnsafeArg.of("message", message), UnsafeArg.of("cause", cause));
```
andybradshaw: Hmm... I'm not sure if these are actually unsafe. I think this is probably fine to leave as-is.

fsxfreak: gotcha, yeah I was a bit unsure about this one since I generally regarded exception messages as unsafe, but if the throwable gets correctly handled and this helper stays private, the message params so far seem safe.

andybradshaw (approved):

PR #612, opened 2025-02-04
Do not apply no-op schema updates

rhuffy:
If a client submits a schema update via `system_update_column_family`, changes that are actually no-ops will still be announced to the cluster.

This PR skips this expensive schema update process if there is no difference in CFMetaData.

### DIFF
--- a/src/java/org/apache/cassandra/service/MigrationManager.java
+++ b/src/java/org/apache/cassandra/service/MigrationManager.java
@@ -465,11 +465,7 @@ public class MigrationManager
         CFMetaData oldCfm = Schema.instance.getCFMetaData(cfm.ksName, cfm.cfName);
         if (oldCfm == null)
             throw new ConfigurationException(String.format("Cannot update non existing table '%s' in keyspace '%s'.", cfm.cfName, cfm.ksName));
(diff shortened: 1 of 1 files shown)


wi11dey:
is there a risk here of skipping the update and the cluster permanently not achieving schema agreement on different uuids ? or has the uuid already been updated by this point

andybradshaw:
> is there a risk here of skipping the update and the cluster permanently not achieving schema agreement on different uuids ? or has the uuid already been updated by this point

I don't think this would affect schema agreement, as the schema doesn't actually change. If it did, this wouldn't be safe to skip. The intent of this is to prevent us from even sending the gossip messages of the no-op, as the receiving node would not actually make any writes to the CF metadata, since the mutation in this message has no modifications. I was also concerned about this, but after reading through how these messages get applied, I _think_ this should be safe.

AFAICT, it might be safe to _never_ announce any mutations with empty modifications, although I'm not sure how often that happens (or how expensive the commitlog activity is when receiving them). But that's more risk than we need to take to resolve the immediate issue.

andybradshaw (approved):

wi11dey:
ah ack. i think would be good to test on an rc and try upgrading/downgrading the old atlas to double check

PR #613, opened 2025-02-13
Safe logging apache.cassandra.repair

fsxfreak:
Added safearg/unsafearg wherever I saw any `logger` usages in this package. Unrolled some `String.format` usages and referred to the arguments directly. Also worth looking at is the `RepairMessageVerbHandler::logErrorAndSendFailureResponse` function which I edited to take variable amount of safe/unsafe args.

Tested by deploying my local version of cassandra to sls form of the service to view logs.
```
# cqlsh
CREATE KEYSPACE my_keyspace WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 3};
USE my_keyspace;
CREATE TABLE users (   user_id UUID PRIMARY KEY,   first_name text,   last_name text,   email text );
CONSISTENCY LOCAL_ONE;
INSERT INTO users (user_id, first_name, last_name, email) VALUES (uuid(), 'John', 'Doe', 'john.doe@example.com');
```
then
```
./nodetool repair my_keyspace;
```
then view logs by tailing `/var/log/service.log`

### DIFF
--- a/src/java/org/apache/cassandra/repair/RepairMessageVerbHandler.java
+++ b/src/java/org/apache/cassandra/repair/RepairMessageVerbHandler.java
@@ -195,7 +195,7 @@ public class RepairMessageVerbHandler implements IVerbHandler<RepairMessage>
     }
 
     private void logErrorAndSendFailureResponse(InetAddress to, int id, String errorMessageFmt, Arg<?>... args) {
-        logger.error(errorMessageFmt, (Object[]) args);
+        logger.error(errorMessageFmt, (Arg[]) args);
 
         MessageOut reply = new MessageOut(MessagingService.Verb.INTERNAL_RESPONSE)
                 .withParameter(MessagingService.FAILURE_RESPONSE_PARAM, MessagingService.ONE_BYTE);
--- a/src/java/org/apache/cassandra/repair/RepairMessageVerbHandler.java
+++ b/src/java/org/apache/cassandra/repair/RepairMessageVerbHandler.java
@@ -24,6 +24,10 @@ import com.google.common.base.Predicate;
 import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
+
+import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeArg;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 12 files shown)


src/java/org/apache/cassandra/repair/RepairMessageVerbHandler.java
```diff
@@ -184,11 +194,11 @@ public void run()
         }
     }
 
-    private void logErrorAndSendFailureResponse(String errorMessage, InetAddress to, int id)
-    {
-        logger.error(errorMessage);
+    private void logErrorAndSendFailureResponse(InetAddress to, int id, String errorMessageFmt, Arg<?>... args) {
+        logger.error(errorMessageFmt, (Object[]) args);
```
d-guo, line 198: does `Arg[]` not work here?

fsxfreak: I believe Arg[] works but the compiler throws this warning
```
build-project:
     [echo] palantir-cassandra: /Volumes/git/cassandra/build.xml
    [javac] Compiling 1 source file to /Volumes/git/cassandra/build/classes/main
    [javac] warning: Supported source version 'RELEASE_6' from annotation processor 'org.openjdk.jmh.generators.BenchmarkProcessor' less than -source '1.8'
    [javac] /Volumes/git/cassandra/src/java/org/apache/cassandra/repair/RepairMessageVerbHandler.java:198: warning: non-varargs call of varargs method with inexact argument type for last parameter;
    [javac]         logger.error(errorMessageFmt, (Arg[]) args);
    [javac]                                       ^
    [javac]   cast to Object for a varargs call
    [javac]   cast to Object[] for a non-varargs call and to suppress this warning
    [javac] 2 warnings
```

let me change it and see what circle-ci says

d-guo: yeah i'm just not sure how casting args to object interacts with the logger. looks like the logger expects an `Object` type so should be fine? can you test locally to make sure the log line output is what we expect? ie has a `safe: true` tag and non-empty `params`

d-guo: https://stackoverflow.com/questions/17667573/why-do-i-get-a-compilation-warning-here-var-args-method-call-in-java

ah probably need to keep `Objects[]`

d-guo (dismissed):

d-guo (dismissed):
pending https://github.com/palantir/cassandra/pull/613#discussion_r1958824596

d-guo (approved):

fsxfreak:
also tested the varargs function by pasting it into a place that would be called, then pushed it onto a test cluster to view logs

PR #614, opened 2025-02-14
Skip replacement node for schema agreement check

inespot:
[#pds-654418](https://palantir.enterprise.slack.com/archives/C08DUCNKV88)

Also increase finish bootstrap gate timeout and later propagate the configuration back to sls-cassandra

### DIFF
--- a/src/java/com/palantir/cassandra/utils/SchemaAgreementCheck.java
+++ b/src/java/com/palantir/cassandra/utils/SchemaAgreementCheck.java
@@ -26,6 +26,7 @@ import java.util.UUID;
 import java.util.function.Supplier;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -54,6 +55,10 @@ public class SchemaAgreementCheck
     {
         this.localSchemaVersionSupplier = localSchemaVersionSupplier;
         this.endpointStatesSupplier = endpointStatesSupplier;
+    }
+
+    public boolean isSchemaInAgreement() {
+        return isSchemaInAgreement(ImmutableList.of());
     }
 
     public boolean isSchemaInAgreement(List<InetAddress> ignoredEndpoints)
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -958,7 +958,8 @@ private void joinTokenRing(int delay, boolean autoBootstrap, Collection<String>
             // if our schema hasn't matched yet, keep sleeping until it does
             // (post CASSANDRA-1391 we don't expect this to be necessary very often, but it doesn't hurt to be careful)
             SchemaAgreementCheck schemaAgreementCheck = new SchemaAgreementCheck();
-            while (!MigrationManager.isReadyForBootstrap() || !schemaAgreementCheck.isSchemaInAgreement())
+            List<InetAddress> ignoredEndpoints = replacing ? ImmutableList.of(DatabaseDescriptor.getReplaceAddress()) : ImmutableList.of();
```
d-guo: should this be `replacing && !isReplacingSameAddress()`?

src/java/com/palantir/cassandra/utils/SchemaAgreementCheck.java
```diff
@@ -49,23 +56,36 @@ public SchemaAgreementCheck()
         this.endpointStatesSupplier = endpointStatesSupplier;
     }
 
-    public boolean isSchemaInAgreement()
+    public boolean isSchemaInAgreement(List<InetAddress> ignoredEndpoints)
```
d-guo, line 59: i personally find this a bit nicer to add the `ignoredEndpoints` as part of the constructor to encapsulate which endpoints this "checker" considers but don't have a strong opinion on it

src/java/com/palantir/cassandra/utils/SchemaAgreementCheck.java
```diff
@@ -49,23 +56,36 @@ public SchemaAgreementCheck()
         this.endpointStatesSupplier = endpointStatesSupplier;
     }
 
-    public boolean isSchemaInAgreement()
```
d-guo, line 52: can we add a `isSchemaInAgreement()` that just calls `isSchemaInAgreement(ImmutableList.of())`

it'll be useful for making the tests look nicer and in https://github.com/palantir/cassandra/pull/617

src/java/org/apache/cassandra/service/StorageService.java
inespot: Ah, very nice catch!

d-guo (dismissed):

d-guo (approved):

PR #616, opened 2025-02-17
Add metric for table size relative to storage load

DaaanJan:
This PR adds a metric for a columnfamily's storage load relative to global storage load. This will help establish tables that are so large that we might no longer be able to compact them if we are also running short on disk space.

### DIFF
--- a/src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
+++ b/src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
@@ -455,9 +455,14 @@ public class ColumnFamilyMetrics
             }
         });
         liveDiskSpaceUsed = createColumnFamilyCounter("LiveDiskSpaceUsed");
-        liveDiskSpaceUsedRatio = Metrics.register(
-            factory.createMetricName("LiveDiskSpaceUsedRatio"),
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
```diff
@@ -453,6 +455,9 @@ public Long getValue()
             }
         });
         liveDiskSpaceUsed = createColumnFamilyCounter("LiveDiskSpaceUsed");
+        liveDiskSpaceUsedRatio = Metrics.register(
+            factory.createMetricName("LiveDiskSpaceUsedRatio"),
+            () -> (double) liveDiskSpaceUsed.getCount() / StorageMetrics.load.getCount());
```
inespot: I think this might be the supported way of doing this: https://github.com/palantir/cassandra/blob/8896a70b015102c212d0a27ed1f4e1f0fabe85c4/src/java/org/apache/cassandra/metrics/CacheMetrics.java#L66-L73

DaaanJan: Nice find, I've implemented that same pattern!

PR #617, opened 2025-02-17
Add streams gate for decommissions

inespot:
Follows: https://github.com/palantir/cassandra/pull/607/files

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -20,7 +20,6 @@ package org.apache.cassandra.service;
 import java.io.*;
 import java.lang.management.ManagementFactory;
 import java.net.InetAddress;
-
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.time.Instant;
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -4205,6 +4205,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
             boolean timeoutExceeded = !startSendStreamsCondition.await(STREAMS_CHECK_GRACE_PERIOD_MINUTES, MINUTES);
             if (timeoutExceeded)
             {
+                recordTransientError(TransientError.DECOMMISSION_ERROR, ImmutableMap.of("reason", "send stream gate timeout"));
                 throw new RuntimeException("Start signal to request streams was not given within 30 minutes. Streams request safety check failed.");
             }
         }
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4186,6 +4197,21 @@ public void decommission() throws InterruptedException
...
+            {
+                throw new RuntimeException("Start signal to request streams was not given within 30 minutes. Streams request safety check failed.");
+            }
+        }
+        catch (InterruptedException e)
+        {
+            throw new AssertionError(e);
+        }
+        logger.info("Received signal to start sending streams.");
+
```
d-guo, line 4219: should all this go inside `unbootstrap`? we probably want to keep it tightly coupled with the `rangesToStream` calculation

d-guo, line 4219: also, can we add
```
setMode(Mode.LEAVING, "calculating ranges to stream", true);
```
to keep the `LEAVING` state

i'm also now wondering if there's any problems with going from `LEAVING` to `WAITING_TO_SEND_STREAMS` and back to `LEAVING`. how do other nodes know when to stop sending this node writes? i would've thought it's from `leaveRing()` but i see there's the comment `// Start with BatchLog replay, which may create hints but no writes since this is no longer a valid endpoint.` that leads me to think it might have to do with the `LEAVING` state

d-guo: ah i think the latter is not a problem because nodes learn that from `ApplicationState.STATUS` not the node's mode

d-guo: ah i think the latter is not a problem because nodes learn that from `ApplicationState.STATUS=VersionedValue.STATUS_LEAVING` not the node's mode

inespot: Yep operation mode should solely be used by client services (although I will make sure to double check)

inespot: I am not sure this really matters as long as the node gossips it is leaving before the check itself? Which it does above

inespot: I think as long as we do this check after the leaving gossip state is set (which it does in `startLeaving`) it should be fine?

inespot: And I feel it makes it more readable to have it in the main function

d-guo (dismissed):

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4188,6 +4199,24 @@ public void decommission() throws InterruptedException
...
         Thread.sleep(timeout);
 
+        try
+        {
+            PendingRangeCalculatorService.instance.blockUntilFinished();
+            setMode(Mode.WAITING_TO_SEND_STREAMS, "Awaiting call to proceed with sending streams during decommission", true);
+            boolean timeoutExceeded = !startSendStreamsCondition.await(STREAMS_CHECK_GRACE_PERIOD_MINUTES, MINUTES);
+            if (timeoutExceeded)
+            {
+                recordTransientError(TransientError.DECOMMISSION_ERROR, ImmutableMap.of("reason", "send stream gate timeout"));
```
d-guo, line 4209: what's the point of this?

inespot: Signaling that the bootstrap has failed. It disappears when the node bounces. Otherwise the node stays in waiting to send streams forever. Alternative is to put it back to normal but I feel it's a little less clear, especially if something is actually wrong. It also allows for operator to detect the decommission failure if/when we want to make retry automatic

d-guo (approved):

PR #623, opened 2025-02-26
Backport Intelligent Token Allocation

rhuffy:
This PR backports a chain of commits related to the intelligent token allocator first introduced in Cassandra 3.0 with [CASSANDRA-7032](https://issues.apache.org/jira/browse/CASSANDRA-7032)

See [this Quip](https://palantir.quip.com/rTzpAfGfHPeZ) for more details

This PR is safe to merge, as all new behavior is guarded behind feature flags...
- `cassandra.allocate_tokens_for_keyspace`
- `cassandra.allocate_tokens_for_local_replication_factor`
If those are not set, we fall back to existing behavior.

### DIFF
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -24,6 +24,17 @@ cluster_name: 'Test Cluster'
 # multiple tokens per node, see http://wiki.apache.org/cassandra/Operations
 num_tokens: 256
 
+# Triggers automatic allocation of num_tokens tokens for this node. The allocation
+# algorithm attempts to choose tokens in a way that optimizes replicated load over
+# the nodes in the datacenter for the replication strategy used by the specified
+# keyspace.
+#
+# The load assigned to each node will be close to proportional to its number of
+# vnodes.
+#
+# Only supported with the Murmur3Partitioner.
+# allocate_tokens_keyspace: KEYSPACE
+
 # initial_token allows you to specify tokens manually.  While you can use # it with
 # vnodes (num_tokens > 1, above) -- in which case you should provide a 
 # comma-separated list -- it's primarily used when adding nodes # to legacy clusters 
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -1,4 +1,4 @@
-# Cassandra storage config YAML 
+# Cassandra storage config YAML
 
 # NOTE:
 #   See http://wiki.apache.org/cassandra/StorageConfiguration for
@@ -26,18 +26,24 @@ num_tokens: 256
 
 # Triggers automatic allocation of num_tokens tokens for this node. The allocation
 # algorithm attempts to choose tokens in a way that optimizes replicated load over
-# the nodes in the datacenter for the replication strategy used by the specified
-# keyspace.
+# the nodes in the datacenter for the replica factor.
 #
 # The load assigned to each node will be close to proportional to its number of
 # vnodes.
 #
 # Only supported with the Murmur3Partitioner.
-# allocate_tokens_keyspace: KEYSPACE
 
-# initial_token allows you to specify tokens manually.  While you can use # it with
+# Replica factor is determined via the replication strategy used by the specified
+# keyspace.
+# allocate_tokens_for_keyspace: KEYSPACE
+
+# Replica factor is explicitly set, regardless of keyspace or datacenter.
+# This is the replica factor within the datacenter, like NTS.
+# allocate_tokens_for_local_replication_factor: 3
+
+# initial_token allows you to specify tokens manually.  While you can use it with
 # vnodes (num_tokens > 1, above) -- in which case you should provide a 
-# comma-separated list -- it's primarily used when adding nodes # to legacy clusters 
+# comma-separated list -- it's primarily used when adding nodes # to legacy clusters
 # that do not have vnodes enabled.
 # initial_token:
 
@@ -284,14 +290,14 @@ counter_cache_save_period: 7200
 #
 # the default option is "periodic" where writes may be acked immediately
 # and the CommitLog is simply synced every commitlog_sync_period_in_ms
-# milliseconds. 
+# milliseconds.
 commitlog_sync: periodic
 commitlog_sync_period_in_ms: 10000
 
 # The size of the individual commitlog file segments.  A commitlog
 # segment may be archived, deleted, or recycled once all the data
 # in it (potentially from each columnfamily in the system) has been
-# flushed to sstables.  
+# flushed to sstables.
 #
 # The default size is 32, which is almost always fine, but if you are
 # archiving commitlog segments (see commitlog_archiving.properties),
@@ -339,7 +345,7 @@ concurrent_counter_writes: 32
 # the smaller of 1/4 of heap or 512MB.
 # file_cache_size_in_mb: 512
 
-# Total permitted memory to use for memtables. Cassandra will stop 
+# Total permitted memory to use for memtables. Cassandra will stop
 # accepting writes when the limit is exceeded until a flush completes,
 # and will trigger a flush based on memtable_cleanup_threshold
 # If omitted, Cassandra will set both to 1/4 the size of the heap.
@@ -375,11 +381,11 @@ memtable_allocation_type: heap_buffers
 
 # This sets the amount of memtable flush writer threads.  These will
 # be blocked by disk io, and each one will hold a memtable in memory
-# while blocked. 
+# while blocked.
 #
 # memtable_flush_writers defaults to the smaller of (number of disks,
 # number of cores), with a minimum of 2 and a maximum of 8.
-# 
+#
 # If your data directories are backed by SSD, you should increase this
 # to the number of cores.
 #memtable_flush_writers: 8
@@ -805,7 +811,7 @@ request_scheduler: org.apache.cassandra.scheduler.NoScheduler
 # NoScheduler - Has no options
 # RoundRobin
 #  - throttle_limit -- The throttle_limit is the number of in-flight
-#                      requests per client.  Requests beyond 
+#                      requests per client.  Requests beyond
 #                      that limit are queued up until
 #                      running requests can complete.
 #                      The value of 80 here is twice the number of
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -41,6 +41,10 @@ num_tokens: 256
 # This is the replica factor within the datacenter, like NTS.
 # allocate_tokens_for_local_replication_factor: 3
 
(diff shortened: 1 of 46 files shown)


src/java/org/apache/cassandra/locator/TokenMetadata.java
```diff
@@ -1031,15 +1031,17 @@ public Token getPredecessor(Token token)
     {
         List tokens = sortedTokens();
         int index = Collections.binarySearch(tokens, token);
-        assert index >= 0 : token + " not found in " + StringUtils.join(tokenToEndpointMap.keySet(), ", ");
+//        assert index >= 0 : token + " not found in " + StringUtils.join(tokenToEndpointMap.keySet(), ", ");
```
fsxfreak, line 1034: commented out for testing?

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -85,6 +85,10 @@ public class Config
     /* initial token in the ring */
     public String initial_token;
     public Integer num_tokens = 1;
+    /** Triggers automatic allocation of tokens if set, using the replication strategy of the referenced keyspace */
+    public String allocate_tokens_for_keyspace = null;
+    /** Triggers automatic allocation of tokens if set, based on the provided replica count for a datacenter */
```
fsxfreak, line 90: nit, this should reference that setting `allocate_tokens_for_keyspace` will override this option

test/unit/org/apache/cassandra/dht/BootStrapperTest.java
```diff
@@ -112,14 +138,206 @@ public boolean isAlive(InetAddress ep)
...
+
+    private void generateFakeEndpoints(TokenMetadata tmd, int numOldNodes, int numVNodes, String dc, String rack) throws UnknownHostException
+    {
         IPartitioner p = StorageService.getPartitioner();
 
         for (int i = 1; i <= numOldNodes; i++)
         {
             // leave .1 for myEndpoint
-            tmd.updateNormalToken(p.getRandomToken(), InetAddress.getByName("127.0.0." + (i + 1)));
+            InetAddress addr = InetAddress.getByName("127." + dc + "." + rack + "." + (i + 1));
```
fsxfreak, line 159: nit: would make the code clearer if there's a helper method to generate these test ip addresses

test/unit/org/apache/cassandra/dht/BootStrapperTest.java
```diff
@@ -112,14 +138,206 @@ public boolean isAlive(InetAddress ep)
...
+            List<Token> tokens = Lists.newArrayListWithCapacity(numVNodes);
+            for (int j = 0; j < numVNodes; ++j)
+                tokens.add(p.getRandomToken(rand));
+            
+            tmd.updateNormalTokens(tokens, addr);
+        }
+    }
+
```
fsxfreak, line 167: would it be useful to add a test of `getBootstrapTokens`? considering the branching/configuration logic, wanna make sure that no matter the configuration, we get valid tokens

src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
```diff
@@ -189,9 +209,13 @@ public BytesToken getMinimumToken()
 
     public BytesToken getRandomToken()
     {
-        Random r = new Random();
+        return getRandomToken(ThreadLocalRandom.current());
```
d-guo, line 212: `ThreadLocalRandom.current()` did you change this from `Random r = new Random()`? why?

src/java/org/apache/cassandra/dht/RandomPartitioner.java
```diff
@@ -70,7 +94,15 @@ public BigIntegerToken getMinimumToken()
 
     public BigIntegerToken getRandomToken()
     {
-        BigInteger token = FBUtilities.hashToBigInteger(GuidGenerator.guidAsBytes());
```
d-guo: why? looks like the implementation is still exactly the same in `FBUtilities`

src/java/org/apache/cassandra/dht/OrderPreservingPartitioner.java
```diff
@@ -106,12 +114,14 @@ public StringToken getMinimumToken()
 
     public StringToken getRandomToken()
     {
-        String chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
-        Random r = new Random();
+        return getRandomToken(ThreadLocalRandom.current());
```
d-guo, line 117: same here

src/java/org/apache/cassandra/locator/TokenMetadata.java
d-guo: looks like the negative index case is allowed based on the next line `if (index < 0) index = -index-1;`

src/java/org/apache/cassandra/dht/BootStrapper.java
```diff
@@ -151,36 +154,83 @@ public void onFailure(Throwable throwable)
...
         int numTokens = DatabaseDescriptor.getNumTokens();
         if (numTokens < 1)
             throw new ConfigurationException("num_tokens must be >= 1");
 
+        if (allocationKeyspace != null)
+            return allocateTokens(metadata, address, allocationKeyspace, numTokens, schemaWaitDelay);
+
+        if (allocationLocalRf != null)
+            return allocateTokens(metadata, address, allocationLocalRf, numTokens, schemaWaitDelay);
+
```
d-guo, line 180: it's weird that `allocationLocalRf` can be non-null and be ignored. maybe worth throwing an error if both `allocationKeyspace ` and `allocationLocalRf` are set at the config level

conf/cassandra.yaml
```diff
@@ -24,9 +24,26 @@ cluster_name: 'Test Cluster'
...
+#
+# Only supported with the Murmur3Partitioner.
+
+# Replica factor is determined via the replication strategy used by the specified
+# keyspace.
+# allocate_tokens_for_keyspace: KEYSPACE
+
+# Replica factor is explicitly set, regardless of keyspace or datacenter.
+# This is the replica factor within the datacenter, like NTS.
+# allocate_tokens_for_local_replication_factor: 3
```
d-guo, line 42: can you clarify or link docs for how these two configs are supposed to work?
1. `allocate_tokens_for_keyspace` makes sense to me since it uses the replication strategy of the given keyspace
```
    static StrategyAdapter getStrategy(final TokenMetadata tokenMetadata, final AbstractReplicationStrategy rs, final InetAddress endpoint)
    {
        if (rs instanceof NetworkTopologyStrategy)
            return getStrategy(tokenMetadata, (NetworkTopologyStrategy) rs, rs.snitch, endpoint);
        if (rs instanceof SimpleStrategy)
            return getStrategy(tokenMetadata, (SimpleStrategy) rs, endpoint);
        throw new ConfigurationException("Token allocation does not support replication strategy " + rs.getClass().getSimpleName());
    }
```
2. `allocate_tokens_for_local_replication_factor` confuses me though because it's just an integer, and the algorithm would be different based on if the replica number is for `NetworkTopologyStrategy` or `SimpleStrategy`. my read of the code is that it just assumes `NetworkTopologyStrategy` but i couldn't find this documented anywhere

src/java/org/apache/cassandra/dht/BootStrapper.java
rhuffy: maybe, but I don't think we need to diverge our fork from existing behavior. I'm sure we could find other examples of config options that don't make sense to set together, but also don't throw in the event that they're both set.

conf/cassandra.yaml
rhuffy: So there is a distinction between the concrete "ReplicationStrategy" implementation, and what the allocator does. If you choose to configure `allocate_tokens_for_local_replication_factor`, it does not assume `NetworkTopologyStrategy` necessarily, but it does look at the EndpointSnitch and allocates tokens in a way that is "rack-aware". 

This effectively means that it assumes `NetworkTopologyStrategy`, but it does not couple the implementation here to the implementation of `NetworkTopologyStrategy`, which I presume was to avoid unnecessary dependencies. NTS also declares more information that the algorithm needs.

Also, consider the case where keyspaces are configured to have different RFs per datacenter. You might have one keyspace with `{DC1: 3, DC2: 1}`, and another keyspace with `{DC1:1, DC2:3}`, and no keyspaces with `{DC1:3, DC2:3}`. However, you might want to allocate your tokens assuming `{DC1:3, DC2:3}`, which previously would only be possible by creating a dummy keyspace, and using that for allocation.

I assume that `allocate_tokens_for_local_replication_factor` was added to address that case, and also to allow the allocator to run before keyspaces are created.

rhuffy: To restate:

`allocate_tokens_for_local_replication_factor` uses a rack-aware strategy, without requiring you to define a per-DC rf.

src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
rhuffy: This change comes from this commit https://github.com/apache/cassandra/commit/47d3b7e7a013b485a2906fc7f0f2fc90e1143966

It appears like they wanted to allow you to provide your own random seed for testing.

The swap for `ThreadLocalRandom` is just to match trunk, and what is done in Murmur3Partitioner.

test/unit/org/apache/cassandra/dht/BootStrapperTest.java
rhuffy: I don't want to address small nits since this is a backport, but I do agree with the sentiment here

src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
d-guo: ah it probably just looks weird in the diff for some reason it's showing you manually added this

rhuffy: Yeah, for this commit there were too many unrelated changes, so I decided to manually copy over what we needed.

test/unit/org/apache/cassandra/dht/BootStrapperTest.java
rhuffy: I don't think so. The meat of the token allocation code is tested heavily. There a probably a lot of places in the C* codebase that don't have good unit test coverage - it doesn't make sense for us to add tests for everything on our fork.

d-guo (approved):

PR #624, opened 2025-02-26
Support Intelligent Token Allocation with the ByteOrderedPartitioner

rhuffy:
Follows #623

Implements `size`, `increaseSlightly`, and `split` on the `ByteOrderedPartitioner` to support intelligent token allocation.

Also extends the existing test cases to run on the `ByteOrderedPartitioner`, with some additions.

### DIFF
--- a/src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
+++ b/src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
@@ -187,7 +187,7 @@ public class ByteOrderedPartitioner implements IPartitioner
         BigInteger right = bigForBytes(rToken.token, sigbytes);
 
         Pair<BigInteger,Boolean> midpair = FBUtilities.midpoint(left, right, 8*sigbytes);
-        return new BytesToken(bytesForBig(midpair.left, sigbytes, false));
+        return new BytesToken(bytesForBig(midpair.left, sigbytes, midpair.right));
     }
 
     public Token split(Token left, Token right, double ratioToLeft)
@@ -239,7 +239,7 @@ public class ByteOrderedPartitioner implements IPartitioner
         if (sigbytes != bytes.length)
         {
             b = new byte[sigbytes];
-            System.arraycopy(bytes, 0, b, 0, bytes.length);
+            System.arraycopy(bytes, 0, b, 0, Math.min(bytes.length, b.length));
         } else
             b = bytes;
         return new BigInteger(1, b);
--- a/src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
+++ b/src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
@@ -47,12 +47,13 @@ import java.util.concurrent.ThreadLocalRandom;
 
 public class ByteOrderedPartitioner implements IPartitioner
 {
+    private static final int GENERATED_TOKEN_LENGTH = 16;
     public static final BytesToken MINIMUM = new BytesToken(ArrayUtils.EMPTY_BYTE_ARRAY);
-    public static final BytesToken MAXIMUM = new BytesToken(new byte[16]);
+    public static final BytesToken MAXIMUM = new BytesToken(new byte[GENERATED_TOKEN_LENGTH]);
     public static final BigDecimal MAX_TOKEN_MAGNITUDE;
     static {
         Arrays.fill(MAXIMUM.token, (byte) 0xFF);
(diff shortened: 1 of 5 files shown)


d-guo:
can we either add some unit tests or manually test on a dev stack the following scenarios?
1. spin up a new cluster with the token allocation algorithm and check its token ring looks reasonable
2. expand an existing skewed cluster with the token allocation algorithm and check that the new token ring is less skewed + looks as expected

d-guo:
also linking some testing in @rynorris's bakeoff PR https://github.com/palantir/cassandra/pull/621
```
// Score is the maximum ownership of any unit in the group.
Galleon Alta:
Cassandra 3 score: 0.14503257009888668
Genetic score: 0.13686754897763648

Rosewood Alta:
Cassandra 3 score: 0.10662885620903151
Genetic score: 0.09905060077862744

Taffeta
Cassandra 3 score: 0.065994680392078
Genetic score: 0.0608640797341693
```

src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
```diff
@@ -121,15 +129,39 @@ public Object getTokenValue()
         @Override
         public double size(Token next)
         {
-            throw new UnsupportedOperationException(String.format("Token type %s does not support token allocation.",
-                                                                  getClass().getSimpleName()));
+            BytesToken n = (BytesToken) next;
+            BigDecimal size = new BigDecimal(bigForBytes(n.token, 16).subtract(bigForBytes(token, 16)));
+            BigDecimal d = size.divide(MAX_TOKEN_MAGNITUDE, MathContext.DECIMAL64); // Scale so that the full range is 1.
+            return (d.compareTo(BigDecimal.ZERO) > 0 ? d : d.add(BigDecimal.ONE)).doubleValue(); // Adjust for signed long, also making sure t.size(t) == 1.
         }
```
d-guo, line 138: why is `t.size(t) == 1` instead of `0`?

rhuffy: I'm not too sure why this is necessary, but it matches the implementation of `size` in the Murmur3Partitioner.

`a.size(b)` returns the distance between tokens `a` and `b`, where `1.0` is the full ring. You could define the distance between the same point as either 0 or 1, and it appears they chose 1.

src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
```diff
@@ -160,20 +192,54 @@ public BytesToken midpoint(Token lt, Token rt)
 
     public Token split(Token left, Token right, double ratioToLeft)
     {
-        throw new UnsupportedOperationException();
+        BytesToken lToken = (BytesToken) left;
+        BytesToken rToken = (BytesToken) right;
+
+        int sigbytes = 16;
```
d-guo: nit: either stay consistent w using `16` or the constant. `SIG_BYTES` can probably also be a static var

src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
```diff
@@ -160,20 +192,54 @@ public BytesToken midpoint(Token lt, Token rt)
...
+
+        if (l.compareTo(r) < 0)
+        {
+            newToken = bytesForBig(r.subtract(l).multiply(ratio).add(l).toBigInteger(), sigbytes, false);
+        }
+        else
+        {
+            // wrapping case
+            // L + ((R - min) + (max - L)) * pct
+            BigInteger maxTokenInt = bigForBytes(MAXIMUM.token, MAXIMUM.token.length);
```
d-guo: nit: have a static constant for this `MAX_TOKEN_MAGNITUDE`

d-guo (dismissed):

rhuffy:
> 1. spin up a new cluster with the token allocation algorithm and check its token ring looks reasonable
> 2. expand an existing skewed cluster with the token allocation algorithm and check that the new token ring is less skewed + looks as expected

I think this should be covered by `ByteOrderedReplicationAwareTokenAllocatorTest#testNewCluster`, in terms of automated testing. You can find the implementation of those tests in `AbstractReplicationAwareTokenAllocatorTest`. It simulates multiple different replication and vnode configs, and my implementation here passes all the tests added by OSS for the Murmur3Partitioner.

In terms of practical testing - yes, I do intend to smoketest on some dev clusters before we consider using this feature. Setting up a dev cluster with an intentionally skewed distribution is a good idea.

d-guo (approved):

PR #626, opened 2025-03-04
require peer endpoint state in SchemaAgreementCheck

rhuffy:
This PR updates the SchemaAgreementCheck to require the EndpointStateMap to contain at least one other endpoint.

This will cause the check to fail if the local node has no state about other peers.

This check only runs when nodes bootstrap, so this will not impact new clusters, as seeds do not bootstrap.

### DIFF
--- a/src/java/com/palantir/cassandra/utils/SchemaAgreementCheck.java
+++ b/src/java/com/palantir/cassandra/utils/SchemaAgreementCheck.java
@@ -37,24 +37,28 @@ import org.apache.cassandra.gms.ApplicationState;
 import org.apache.cassandra.gms.EndpointState;
 import org.apache.cassandra.gms.Gossiper;
 import org.apache.cassandra.gms.VersionedValue;
+import org.apache.cassandra.locator.TokenMetadata;
 import org.apache.cassandra.service.StorageService;
+import org.apache.cassandra.utils.FBUtilities;
 
 public class SchemaAgreementCheck
 {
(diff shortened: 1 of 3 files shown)


src/java/com/palantir/cassandra/utils/SchemaAgreementCheck.java
```diff
@@ -89,8 +98,10 @@ private boolean schemaIsEqualToLocalVersion(UUID localSchemaVersion, InetAddress
         VersionedValue schema = endpointState.getApplicationState(ApplicationState.SCHEMA);
         boolean match = schema != null && localSchemaVersion.equals(UUID.fromString(schema.value));
         if(!match) {
-            logger.warn("Schema agreement check failure", SafeArg.of("localSchemaVersion", localSchemaVersion),
-                        SafeArg.of("endpoint", address), SafeArg.of("remoteSchemaVersion", schema.value));
+            logger.warn("Schema agreement check failure",
+                        SafeArg.of("localSchemaVersion", localSchemaVersion),
+                        SafeArg.of("endpoint", address),
+                        SafeArg.of("remoteSchemaVersion", schema));
```
rhuffy, line 104: drive by. `schema.value` can throw a NPE

inespot: TIL "drive by." idiom 🚗

inespot (approved):

PR #628, opened 2025-03-04
Expose endpoint shadow state map

nickbar01234:
Used internally within Palantir wrapper. I considered re-using `Gossiper#useShadowRound`, but that changes the existing steady-state behavior.

### DIFF
--- a/src/java/org/apache/cassandra/gms/Gossiper.java
+++ b/src/java/org/apache/cassandra/gms/Gossiper.java
@@ -879,6 +879,11 @@ public class Gossiper implements IFailureDetectionEventListener, GossiperMBean
         return endpointStateMap.entrySet();
     }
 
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -879,6 +879,11 @@ public Set<Entry<InetAddress, EndpointState>> getEndpointStates()
         return endpointStateMap.entrySet();
     }
 
+    public Map<InetAddress, EndpointState> getEndpointShadowStateMap()
+    {
+        return endpointShadowStateMap;
```
rhuffy: Might want to make this return an immutable copy, depending on how you intend to consume.

rhuffy (approved):

PR #629, opened 2025-03-05
Add metric for tables with throttled compactions due to high disk usage

lyubent:
Adds CF metric showing  number of tables being throttled when compacting as a follow up to:
- #508

The metric tracks the tables that were throttled due to high disk usage. Compactions are removed from the throttle list once the compaction completes.

### DIFF
--- a/src/java/com/palantir/cassandra/db/compaction/CompactionThroughputThrottler.java
+++ b/src/java/com/palantir/cassandra/db/compaction/CompactionThroughputThrottler.java
@@ -3,6 +3,7 @@ package com.palantir.cassandra.db.compaction;
 import com.google.common.util.concurrent.RateLimiter;
 import com.palantir.logsafe.SafeArg;
 
+import org.apache.cassandra.config.CFMetaData;
 import org.apache.cassandra.db.ColumnFamilyStore;
 import org.apache.cassandra.db.Directories;
 import org.apache.cassandra.db.compaction.CompactionManager;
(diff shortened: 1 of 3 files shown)


src/java/com/palantir/cassandra/db/compaction/CompactionThroughputThrottler.java
```diff
@@ -11,15 +12,39 @@
...
+    private final Set<UUID> throttledCfs = new HashSet<>();
+
+    private CompactionThroughputThrottler() {}
+
+    public int throttledTableCount()
+    {
+        return throttledCfs.size();
+    }
+
+    public void maybeRemoveThrottledCompaction(CFMetaData cfMetaData)
```
andybradshaw, line 38: Understanding that this is for metric reporting, not really correctness... but curious if adding/removing from this set from different code paths is best? A potential alternative would be to remove from the set when next attempting to get a rate limiter for a cf. I guess the difference would be that with this approach, we'd consider a cf compaction "throttled" between getting the rate limiter and finishing the compaction task, as opposed to considering it as throttled between attempts to get the rate limiter (e.g. until we've actually gone through the disk usage checks again). 

Definitely not as familiar with this code path/workflow, so would love to get your thoughts here.

lyubent: This was my original design but I went with removing this on compaction completion for the purpose of keeping the metric correct and up-to-date. I don't have a strong opinion on how we do this. The trade-off is roughly: relying on the removal during the check for whether throttling is necessary would mean a lot less redundant calls to remove tables from the set, that were never there to begin with **vs** having a very up-to-date metric that does have a lot of redundant calls to Set#remove.

andybradshaw (approved):

PR #630, opened 2025-03-06
Support coercing the read consistency level

rhuffy:
This PR introduces the config option `coerce_read_consistency_level`, which forces all read requests to a CL specified in `cassandra.yaml`, or live updated by `nodetool setcoercereadconsistencylevel`.

### DIFF
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -315,6 +315,7 @@ public class Config
     public int otc_coalescing_enough_coalesced_messages = 8;
 
     public volatile boolean coerce_read_consistency_all = false;
+    public Set<String> keyspaces_to_coerce_read_consistency = null;
 
     public int windows_timer_interval = 0;
 
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -315,7 +315,7 @@ public class Config
     public int otc_coalescing_window_us = otc_coalescing_window_us_default;
     public int otc_coalescing_enough_coalesced_messages = 8;
 
-    public volatile ConsistencyLevel coerce_read_consistency_level = ConsistencyLevel.ALL;
+    public volatile ConsistencyLevel coerce_read_consistency_level = null;
 
     public int windows_timer_interval = 0;
 
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -315,7 +315,6 @@ public class Config
     public int otc_coalescing_enough_coalesced_messages = 8;
 
     public volatile boolean coerce_read_consistency_all = false;
-    public Set<String> keyspaces_to_coerce_read_consistency = null;
 
     public int windows_timer_interval = 0;
 
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -37,6 +37,7 @@ import org.slf4j.LoggerFactory;
 
 import org.apache.cassandra.config.EncryptionOptions.ClientEncryptionOptions;
 import org.apache.cassandra.config.EncryptionOptions.ServerEncryptionOptions;
+import org.apache.cassandra.db.ConsistencyLevel;
 import org.apache.cassandra.exceptions.ConfigurationException;
 import org.supercsv.io.CsvListReader;
 import org.supercsv.prefs.CsvPreference;
@@ -313,6 +314,8 @@ public class Config
     public static final int otc_coalescing_window_us_default = 200;
     public int otc_coalescing_window_us = otc_coalescing_window_us_default;
     public int otc_coalescing_enough_coalesced_messages = 8;
+
+    public volatile ConsistencyLevel coerce_read_consistency_level = ConsistencyLevel.ALL;
 
     public int windows_timer_interval = 0;
 
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -37,7 +37,6 @@ import org.slf4j.LoggerFactory;
 
 import org.apache.cassandra.config.EncryptionOptions.ClientEncryptionOptions;
 import org.apache.cassandra.config.EncryptionOptions.ServerEncryptionOptions;
-import org.apache.cassandra.db.ConsistencyLevel;
 import org.apache.cassandra.exceptions.ConfigurationException;
 import org.supercsv.io.CsvListReader;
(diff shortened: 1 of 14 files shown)


src/java/org/apache/cassandra/tools/nodetool/SetCoerceReadConsistencyLevel.java
```diff
@@ -0,0 +1,39 @@
...
+public class SetCoerceReadConsistencyLevel extends NodeTool.NodeToolCmd
+{
+
+    @Arguments(title = "consistency_level", usage = "<CL name>", description = "Cassandra CONSISTENCY_LEVEL in UPPER_SNAKE_CASE", required = true)
+    private String consistencyLevel = null;
+
+    @Override
+    public void execute(NodeProbe probe)
+    {
+        probe.setCoerceReadConsistencyLevel(ConsistencyLevel.valueOf(consistencyLevel));
```
inespot: Would this handle setting it back to null?

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -226,9 +226,9 @@ public static ColumnFamily cas(String keyspaceName,
                 Tracing.trace("Reading existing values for CAS precondition");
                 long timestamp = System.currentTimeMillis();
                 ReadCommand readCommand = ReadCommand.create(keyspaceName, key, cfName, timestamp, request.readFilter());
-                ConsistencyLevel consistencyForPrecondition = consistencyForPaxos == ConsistencyLevel.LOCAL_SERIAL
+                ConsistencyLevel consistencyForPrecondition = maybeCoerceReadConsistencyLevel(consistencyForPaxos == ConsistencyLevel.LOCAL_SERIAL
```
inespot: Do we still need the `maybeCoerceReadConsistencyLevel` here then? Considering it is used in the read later on

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1274,6 +1275,7 @@ public static List<Row> read(List<ReadCommand> commands, ConsistencyLevel consis
     public static List<Row> read(List<ReadCommand> commands, ConsistencyLevel consistencyLevel, ClientState state)
     throws UnavailableException, IsBootstrappingException, ReadFailureException, ReadTimeoutException, InvalidRequestException
     {
+        consistencyLevel = maybeCoerceReadConsistencyLevel(consistencyLevel);
```
inespot, line 1279: If you have serial read here, your later `readWithPaxos` will fail. I think we need to exclude `SERIAL_*` consistency from `maybeCoerceReadConsistencyLevel` and only set it once it falls back to actual QUORUM?

inespot: I'll also check the cas path a little closer

d-guo: yeah i think we need to add a coerce in `fetchRows` a lot of the reads end up going through there

d-guo: actually do *all* reads got through there? edit: as @rhuffy mentioned on slack, no there's also `getRangeSlice`

rhuffy: doing the coerce in fetchRows is nice, _but_ the problem is that you then don't have access the the correctly coerced consistency level for reporting in metrics

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -314,6 +314,8 @@ public class Config
     public int otc_coalescing_window_us = otc_coalescing_window_us_default;
     public int otc_coalescing_enough_coalesced_messages = 8;
 
+    public volatile boolean coerce_read_consistency_all = false;
```
inespot, line 317: Could you add a comment describing the type of situations where we would need to set this?

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1738,13 +1745,16 @@ private static float calculateResultRowsUsingEstimatedKeys(ColumnFamilyStore cfs
     public static List<Row> getRangeSlice(AbstractRangeCommand command, ConsistencyLevel consistency_level)
     throws UnavailableException, ReadFailureException, ReadTimeoutException
     {
+        consistency_level = maybeCoerceReadConsistencyLevel(consistency_level);
```
inespot, line 1748: This feature is for now heavily relying on a manual audit of the code that confirmed that we hit all the read path we wanted. I am wondering if there is a way we could add some tests to assert this remains true in case of any modifications to those code path in the future

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -89,11 +89,13 @@ public AtomicInteger load(InetAddress inetAddress)
             return new AtomicInteger(0);
         }
     };
-    private static final ClientRequestMetrics readMetrics = new ClientRequestMetrics("Read");
-    private static final ClientRequestMetrics rangeMetrics = new ClientRequestMetrics("RangeSlice");
+    private static final ClientRequestMetrics topReadMetrics = new ClientRequestMetrics("Read");
```
inespot, line 92: "top"?

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
```diff
@@ -416,6 +421,10 @@ public Double getValue()
         coordinatorReadLatency = new LatencyMetrics(factory, "CoordinatorRead",  new LatencyMetrics(globalNameFactory, "CoordinatorRead"));
         coordinatorReadScanLatency = new LatencyMetrics(factory, "CoordinatorReadScan",  new LatencyMetrics(globalNameFactory, "CoordinatorReadScan"));
         coordinatorScanLatency = new LatencyMetrics(factory, "CoordinatorScan", new LatencyMetrics(globalNameFactory, "CoordinatorScan"));
+        coordinatorReadLatencyByCL = MetricUtils.byConsistencyLevel(cl -> new LatencyMetrics(factory, "CoordinatorRead", ",consistency=" + cl, coordinatorReadLatency));
+        coordinatorReadScanLatencyByCL = MetricUtils.byConsistencyLevel(cl -> new LatencyMetrics(factory, "CoordinatorReadScan", ",consistency=" + cl, coordinatorReadScanLatency));
+        coordinatorScanLatencyByCL = MetricUtils.byConsistencyLevel(cl -> new LatencyMetrics(factory, "CoordinatorScan", ",consistency=" + cl, coordinatorScanLatency));
```
inespot, line 426: We verified during the soaking/testing that parent metrics remained consistent?

PR #631, opened 2025-03-07
Add Metrics for Read Repair

d-guo:
Add Metrics for Avoided Read Repairs and Digest + Data Reads

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageProxy.java
+++ b/src/java/org/apache/cassandra/service/StorageProxy.java
@@ -1896,7 +1896,6 @@ public class StorageProxy implements StorageProxyMBean
                     }
                     scanHandlers.add(Pair.create(nodeCmd, handler));
                     numRequestQueries++;
-
                 }
                 Tracing.trace("Submitted {} concurrent range requests covering {} ranges", concurrentRequests, i - concurrentFetchStartingIndex);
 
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1443,6 +1443,7 @@ private static List<Row> fetchRows(List<ReadCommand> initialCommands, Consistenc
                 try
                 {
                     Row row = exec.get();
+                    Keyspace.open(exec.command.ksName).getColumnFamilyStore(exec.command.cfName).metric.avoidedReadRepairs.mark();
```
fsxfreak, line 1446: why does this line mean an avoided read repair? is `exec.get()` meant to throw `DigestMismatchException` here

d-guo: yep! `exec.get()` will throw a `DigestMismatchException` if the resolver sees the received digests don't match. if they do match we can avoid doing all the read repair logic under `catch (DigestMismatchException ex)`

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1895,6 +1896,7 @@ public static List<Row> getRangeSlice(AbstractRangeCommand command, ConsistencyL
                     }
                     scanHandlers.add(Pair.create(nodeCmd, handler));
                     numRequestQueries++;
+
```
rhuffy: nit whitespace

rhuffy (approved):

PR #632, opened 2025-03-07
Range Scan with Page Token (Local)

d-guo:
This PR implements _Limited and Resumable Range Scans_ at the local level. The next two PRs https://github.com/palantir/cassandra/pull/639 annd https://github.com/palantir/cassandra/pull/651 add the coordinator and thrift level logic. Together, these expose three new thrift endpoints that allow clients to limit and resume range scans by returning a page token: `multiget_multislice_paging`, `multiget_slice_paging`, and `get_range_slices_paging`.

Cassandra struggles with large range scans, and that's expected. What is not expected is that issuing a small `multiget_multislice`, `multiget_slice`, and `get_range_slices` request can time out or bring down a Cassandra node. The issue is that a range scan over the interval `[a, b]` with `LIMIT 1` may yield only one results `{b}` but still have to scan through any number of cells covered by range tombstones such as `{a1, a11, ..., a2}` with a range tombstone covering `[a1, a2]`. This makes it impossible to effectively protect the Cassandra node as a client via regular paging.

The solution implemented here adds pagination within Cassandra with any type of threshold defined.

This initial implementation adds a simple threshold: the number of cells read, including those covered by tombstones. This prevents timing out and still guarantees the client can make progress by returning a page token of the scanned cell, even if it's dead and not included in the results set. The page token is calculated by each replica node and returned along with the partial results set to the coordinator.

*Palantir Internal Docs*
[[RFC] Resumable Range Scans](https://palantir.quip.com/PGwbABP19ae2)
[Resumable Range Scans Divergences and Implementation Details](https://palantir.quip.com/lqdmAaLVFxI3#XRDABAkdap3)


d-guo:
i just followed the existing testing pattern in the added unit tests, but it does look a bit messy. i'll look at how i can clean them up but didn't want to block review on it. let me know if you have any ideas here

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1312,6 +1312,11 @@ public static int getTombstoneFailureThreshold()
         return conf.tombstone_failure_threshold;
     }
 
+    public static int getRangeScanCellsReadThreshold()
+    {
+        return Integer.getInteger("cassandra.range_scan_cells_read_threshold", conf.range_scan_cells_read_threshold);
+    }
```
rhuffy: can we prefix this `palantir_cassandra` to make it clear it's our addition?

src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -60,6 +61,9 @@ public abstract class ColumnFamily implements Iterable<Cell>, IRowCacheEntry
 
     protected final CFMetaData metadata;
 
+    private PageToken pageToken;
+    private boolean isPageTokenSet = false;
```
rhuffy: why do we need this boolean flag? Can you accomplish the same with a null check on pageToken?

src/java/org/apache/cassandra/db/commitlog/CommitLogDescriptor.java
```diff
@@ -57,12 +57,13 @@ public class CommitLogDescriptor
...
     public static final int VERSION_21 = 4;
     public static final int VERSION_22 = 5;
+    public static final int VERSION_23 = 6;
     /**
      * Increment this number if there is a changes in the commit log disc layout or MessagingVersion changes.
      * Note: make sure to handle {@link #getMessagingVersion()}
      */
     @VisibleForTesting
-    public static final int current_version = VERSION_22;
+    public static final int current_version = VERSION_23;
```
rhuffy: what are the implications of incrementing this version?

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -104,15 +116,26 @@ public SliceQueryFilter(ColumnSlice[] slices, boolean reversed, int count)
...
         this.reversed = reversed;
+        this.usePageToken = usePageToken;
         this.count = count;
         this.compositesToGroup = compositesToGroup;
         this.highMemoryCollectionThreshold = LOG_HIGH_MEMORY_COLLECTION
-                                             ? Long.parseLong(System.getProperty("palantir_cassandra.high_memory_collection_threshold_in_mb")) * FileUtils.ONE_MB
-                                             : null;
+                ? Long.parseLong(System.getProperty("palantir_cassandra.high_memory_collection_threshold_in_mb")) * FileUtils.ONE_MB
+                : null;
     }
```
rhuffy: nit, probably don't want this since it's just a whitespace change

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -302,9 +325,32 @@ public void collectReducedColumns(ColumnFamily container, Iterator<Cell> reduced
         long dataSizeCollected = 0;
         long metadataSizeCollected = 0;
 
+        // only set page token if usePageToken is true
+        // if the range scan completes, set the pageToken to a "end of row" value
+        // otherwise set it to a cell name if we hit on of the defensive guards, except in the case we hit the guard on the last column of the row, in which
+        // case return the "end of row" value again
```
rhuffy: type "if we hit on -> one of the  defensive guards...

(nit) Also can you fix the line length?

test/unit/org/apache/cassandra/AbstractSerializationsTester.java
```diff
@@ -42,7 +42,7 @@ public class AbstractSerializationsTester
         put("1.0", 3);
         put("1.2", MessagingService.VERSION_12);
         put("2.0", MessagingService.VERSION_20);
-        put("2.1", MessagingService.VERSION_21);
+        put("2.1", MessagingService.VERSION_23);
```
rhuffy: Appreciate this is just testing, but we should be on `2.2`, not `2.1`. Perhaps the existing test is out of date

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -83,7 +83,8 @@ public final class MessagingService implements MessagingServiceMBean
     public static final int VERSION_20 = 7;
     public static final int VERSION_21 = 8;
     public static final int VERSION_22 = 9;
-    public static final int current_version = VERSION_22;
+    public static final int VERSION_23 = 10;
+    public static final int current_version = VERSION_23;
```
rhuffy: and, what are the implications of incrementing this version?

test/conf/cassandra.yaml
```diff
@@ -40,3 +40,4 @@ compaction_throughput_mb_per_sec: 0
 row_cache_class_name: org.apache.cassandra.cache.OHCProvider
 row_cache_size_in_mb: 16
 enable_user_defined_functions: true
+range_scan_cells_read_threshold: 5
```
rhuffy: nit, should end with newline

src/java/org/apache/cassandra/db/ColumnFamilySerializer.java
```diff
@@ -131,7 +170,8 @@ public long serializedSize(ColumnFamily cf, TypeSizes typeSizes, int version)
         {
             return typeSizes.sizeof(true)  /* nullness bool */
                  + cfIdSerializedSize(cf.id(), typeSizes, version)  /* id */
-                 + contentSerializedSize(cf, typeSizes, version);
+                    + contentSerializedSize(cf, typeSizes, version)
+                   + pageTokenSerializedSize(cf, typeSizes, version);
```
rhuffy: could you fix the formatting here to minimize the diff?

rhuffy (commented):
Gave a first pass. Will need to check out the code locally and review again more closely.

Could you add a PR description here for posterity and also link the relevant internal docs?

src/java/org/apache/cassandra/db/ColumnFamily.java
d-guo: for deserialization, we need something to indicate `pageToken` is null

src/java/org/apache/cassandra/net/MessagingService.java
d-guo: the version in `MessagingService` is the "protocol version" that the node uses for message serializing/deserializing
```
private final ConcurrentMap<InetAddress, Integer> versions = new NonBlockingHashMap<InetAddress, Integer>();
```
this map tracks the version each node is on. the current node uses this info to be backward compatible when sending/receiving messages from other nodes that may be on other versions due to a rolling upgrade or rollback. for example in this PR the node won't serialize the `pageToken` field unless `version >= MessagingService.VERSION_23`

src/java/org/apache/cassandra/db/commitlog/CommitLogDescriptor.java
d-guo: similar reasoning as below for `MessagingService`, except this one is for serializing/deserializing the mutations in the commitlog since the node may be on a different version after an upgrade
```
            mutation = Mutation.serializer.deserialize(new DataInputStream(bufIn),
                                                       desc.getMessagingVersion(),
                                                       ColumnSerializer.Flag.LOCAL);
```

src/java/org/apache/cassandra/db/ColumnFamilySerializer.java
```diff
@@ -71,6 +72,14 @@ public void serialize(ColumnFamily cf, DataOutputPlus out, int version)
...
             }
             assert count == written: "Table had " + count + " columns, but " + written + " written";
+            if (version >= MessagingService.VERSION_22_18)
+            {
+                out.writeBoolean(cf.isPageTokenSet());
+                if (cf.isPageTokenSet())
+                {
+                    new PageToken.Serializer(columnSerializer).serialize(cf.pageToken(), out, version);
+                }
+            }
```
rhuffy, line 82: could you update the comment at the top of this method with the additions to the serialization format?

rhuffy: Could you update the comment at the top of this method with the changes to the serialization format?

rhuffy: Is it necessary to write a boolean for `isPageTokenSet`? Could we instead choose a magic number to mean "page token not set"?

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -83,7 +83,8 @@ public final class MessagingService implements MessagingServiceMBean
     public static final int VERSION_20 = 7;
     public static final int VERSION_21 = 8;
     public static final int VERSION_22 = 9;
-    public static final int current_version = VERSION_22;
+    public static final int VERSION_22_18 = 10;
+    public static final int current_version = VERSION_22_18;
```
rhuffy: For this version (and the commitlog version), I have two concerns.

1. I don't love `22_18` as a name since it implies OSS `2.2.18`. What do you think about `22_PLTR`, or similar to call out that this is unique to the fork?
2. The choice of `int 10` here conflicts with [OSS's choice](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/net/MessagingService.java#L224) for `VERSION_30`. I know we do not intent at this time to attempt a live upgrade from `palantir_2.2` to `3.0`, but I do think we shouldn't needlessly leave a mine like this. What do you think about picking a version number substantially higher than any of the OSS - defined versions, just so that there's no conflict. They say the version is 8 bits, so maybe something like `50` is high enough, while still leaving us plenty of headroom to increment the version in the future.

src/java/org/apache/cassandra/db/filter/PageToken.java
```diff
@@ -0,0 +1,141 @@
...
+
+        public PageToken deserialize(DataInput in, ColumnSerializer.Flag flag, int version) throws IOException
+        {
+            boolean reachedEnd = in.readBoolean();
+            if (reachedEnd)
+            {
+                return PageToken.createPageTokenReachedEnd();
+            }
+            return PageToken.createPageToken(columnSerializer.deserialize(in, flag));
+        }
```
rhuffy, line 94: `version` is unused here. Is that intentional?

src/java/org/apache/cassandra/db/filter/PageToken.java
```diff
@@ -0,0 +1,141 @@
...
+    private PageToken(Cell pageToken, boolean reachedEnd)
+    {
+        this.token = pageToken;
+        this.reachedEnd = reachedEnd;
+    }
+
+    public Cell getToken()
+    {
+        return token;
+    }
```
rhuffy: token is a bit of an overloaded work (think tokens wrt the partitioner). Is there any other naming we could use?

rhuffy: maybe "column", since we are using the ColumnSerializer to handle this value.

src/java/org/apache/cassandra/db/filter/QueryFilter.java
```diff
@@ -433,6 +433,17 @@ public static QueryFilter getSliceFilter(DecoratedKey key,
         return new QueryFilter(key, cfName, new SliceQueryFilter(start, finish, reversed, limit), timestamp);
     }
 
+    public static QueryFilter getSliceFilterUsingPageToken(DecoratedKey key,
+                                                           String cfName,
```
rhuffy: can you add a docstring to match the other methods of this class?

src/java/org/apache/cassandra/thrift/ThriftValidation.java
```diff
@@ -668,4 +668,25 @@ public static IDiskAtomFilter asIFilter(SlicePredicate sp, CFMetaData metadata,
             filter = SuperColumns.fromSCFilter(metadata.comparator, superColumn, filter);
         return filter;
     }
+
+    public static IDiskAtomFilter asIFilterUsingPageToken(SlicePredicate sp, CFMetaData metadata, ByteBuffer superColumn)
+    {
+        SliceRange sr = sp.slice_range;
+        IDiskAtomFilter filter;
```
rhuffy: This is only used in tests. Will it be used in production by the coordinator when that is implemented?

test/unit/org/apache/cassandra/db/SerializationsTest.java
```diff
@@ -118,6 +118,7 @@ public void testRangeSliceCommandRead() throws IOException
         DataInputStream in = getInput("db.RangeSliceCommand.bin");
         for (int i = 0; i < 6; i++)
             MessageIn.read(in, getVersion(), -1);
+        // reading is failing here because it doesn't detect the correct version during deserialization
```
rhuffy: is this comment still accurate?

test/unit/org/apache/cassandra/db/ColumnFamilyStoreTest.java
```diff
@@ -1394,6 +1408,469 @@ public void testRangeSlicePaging() throws Throwable
         assertColumnNames(row2, "c1");
     }
 
+    @Test
+    public void testRangeSlicePageTokenVariousSlices()
+    {
+        String keyspaceName = KEYSPACE1;
```
rhuffy: is there any way you could make these tests more readable? Perhaps more, shorter test cases, and some util methods for constructing the initial state.

rhuffy: You can also experiment with inner classes for test suites, so that a set of tests can all share the same @BeforeEach method. Or just put the tests for page token range scans in a separate test class.

test/unit/org/apache/cassandra/db/ColumnFamilyTest.java
```diff
@@ -109,6 +109,98 @@ public void testManyColumns() throws IOException
         assert Iterables.size(cf.getColumnNames()) == map.size();
     }
 
+    @Test
+    public void testValuedPageTokenWithEmptyColumns() throws IOException
+    {
+        ColumnFamily cf = ArrayBackedSortedColumns.factory.create(KEYSPACE1, CF_STANDARD1);
+        DataOutputBuffer bufOut = new DataOutputBuffer();
+        Cell pageToken = column("1", "Daniel Guo is a good man: 1", 314);
```
rhuffy, line 117: It might be better to save these easter eggs for private repos 🤷‍♂️  😅

test/unit/org/apache/cassandra/db/ColumnFamilyStoreTest.java
rhuffy: Specifically, I think each assertion where you've commented a brief description of the test case should really be a separate `@Test` method.

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -376,6 +426,11 @@ public void collectReducedColumns(ColumnFamily container, Iterator<Cell> reduced
                       warnTombstones ? " (see tombstone_warn_threshold)" : "");
     }
 
+    private boolean hitRangeScanThreshold(long cellsRead)
+    {
+        return cellsRead >= DatabaseDescriptor.getRangeScanCellsReadThreshold();
+    }
```
rhuffy, line 434: say in the future we decide we want the client to be able to pass a `CellsReadThreshold` along with every request. How would that work?

src/java/org/apache/cassandra/db/commitlog/CommitLogDescriptor.java
```diff
@@ -195,6 +196,8 @@ public int getMessagingVersion()
                 return MessagingService.VERSION_21;
             case VERSION_22:
                 return MessagingService.VERSION_22;
+            case VERSION_22_18:
+                return MessagingService.VERSION_22_18;
```
rhuffy: How does this PR impact the serialized format of the commitlog? Are we resilient to upgrades/downgrades? Do we have test coverage for this? I didn't see any test commitlog files as part of this PR.

src/java/org/apache/cassandra/db/ColumnFamilySerializer.java
d-guo: i think it's simplest to do it this way. several of the other serializers follow a similar pattern

src/java/org/apache/cassandra/db/filter/PageToken.java
d-guo: yeah, there's no need to distinguish the version within the page token deserialization. most other serializers similarly don't use the version. i believe you only need it if your serializer behaves differently depending on the version received, which is not the case here

src/java/org/apache/cassandra/db/filter/PageToken.java
d-guo: renamed to `cell`

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
d-guo: this was something we explicitly did not want to support from the RFC discussion, but if we decide to do that in the future it's a small change to just include the custom threshold from the client in the thrift request and filter

src/java/org/apache/cassandra/db/commitlog/CommitLogDescriptor.java
d-guo: yep, those already exist in `CommitLogUpgradeTest`

test/unit/org/apache/cassandra/db/SerializationsTest.java
d-guo: oops, no. that was meant to be a note for myself

src/java/org/apache/cassandra/thrift/ThriftValidation.java
d-guo: yep, this is for getting range slices which is implemented in the next PR

d-guo: now that you mention that i'm realizing i only have test coverage through `cfs.getRangeSlice` which is nearly identical but flexes different logic from `getTopLevelColumns` which is used by the normal `getRow`. i'll add tests for that too after your requested refactor

test/unit/org/apache/cassandra/db/ColumnFamilyTest.java
d-guo: i took this from the test above. i'd like to keep it if it's okay with you :)

src/java/org/apache/cassandra/thrift/ThriftValidation.java
d-guo: oh whoops just kidding — i added them in `KeyspaceTest` instead of `ColumnFamilyStoreTest` bc of where the methods are invoked but the tests are there

src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -21,12 +21,7 @@
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.security.MessageDigest;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
rhuffy, line 24: let's avoid .* imports

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -282,6 +282,7 @@ public class Config
 
     public volatile int tombstone_warn_threshold = 1000;
     public volatile int tombstone_failure_threshold = 100000;
+    public volatile int range_scan_cells_read_threshold = 5000000;
```
inespot, line 285: How was this number defined?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1322,6 +1322,11 @@ public static int getTombstoneFailureThreshold()
         return conf.tombstone_failure_threshold;
     }
 
+    public static int getRangeScanCellsReadThreshold()
+    {
+        return Integer.getInteger("palantir_cassandra.range_scan_cells_read_threshold", conf.range_scan_cells_read_threshold);
+    }
```
inespot, line 1328: We intentionally don't want to make this configurable live? Solely to simplify the first version of this feature or are we worried about other interactions?

src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -60,6 +56,8 @@ public abstract class ColumnFamily implements Iterable<Cell>, IRowCacheEntry
 
     protected final CFMetaData metadata;
 
+    private PageToken pageToken;
```
inespot, line 61: I think we discussed this in one of our 1:1s, but this structure relies on the fact that you never reuse the same columnfamily twice, correct? And consequently that the row cache remains disabled? Are there other places in the code where a columnfamily could be re-used?

src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -96,6 +94,31 @@ public int liveCQL3RowCount(long now)
...
+
+    public PageToken pageToken()
+    {
+        return pageToken;
+    }
+
+    public boolean isPageTokenSet()
+    {
+        return pageToken != null;
+    }
```
inespot, line 122: Nit: you could probably inline this

src/java/org/apache/cassandra/net/MessagingService.java
inespot: Onboard with not conflicting with existing OSS versions. I was wondering whether going _above_ all versions would make a potential OSS upgrade harder, but with this feature and us modifying messaging versions we'll probably still want to stay on the PLTR increment "track" and merge both the changes from OSS and Palantir

src/java/org/apache/cassandra/db/filter/PageToken.java
```diff
@@ -0,0 +1,141 @@
...
+
+    public Cell getCell()
+    {
+        return cell;
+    }
+
+    public boolean isReachedEnd()
+    {
+        return reachedEnd;
+    }
```
inespot, line 50: Nitty nit: `cell()` and `reachedEnd()` to match your `pageToken()` in ColumnFamily

inespot: Also wanna rename `reachedEnd()` to `endOfRow()` also for consistency?

test/unit/org/apache/cassandra/db/SerializationsTest.java
```diff
@@ -318,6 +318,33 @@ public void testTruncateRead() throws IOException
...
+
+        DataInputStream in = getInput("db.PageToken.bin");
+        PageToken deserializedPageToken1 = pageTokenSerializer.deserialize(in, getVersion());
+        PageToken deserializedPageToken2 = pageTokenSerializer.deserialize(in, getVersion());
+
+        assert deserializedPageToken1.isReachedEnd();
+        assert deserializedPageToken1.getCell() == null;
+        assert !deserializedPageToken2.isReachedEnd();
+        assert deserializedPageToken2.getCell().equals(pageTokenHasValue.getCell());
+    }
```
inespot, line 346: Not sure if I missed this somewhere but could add one or two small unit tests for columnfamily serialization when the target is at a lower version and when the target is at a higher version?

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -302,9 +324,33 @@ public void collectReducedColumns(ColumnFamily container, Iterator<Cell> reduced
...
+        // only set page token if usePageToken is true
+        // if the range scan completes, set the pageToken to "end of row" value
+        // otherwise set it to a cell name if we hit one of the defensive guards
+
+        CellName firstCell = null;
+        CellName lastScannedCellInContainer = null;
         while (!columnCounter.hasSeenAtLeast(count) && reducedCells.hasNext())
         {
             Cell cell = reducedCells.next();
+            assert cell != null;
```
inespot, line 336: Do we need this?

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -302,9 +324,33 @@ public void collectReducedColumns(ColumnFamily container, Iterator<Cell> reduced
         long dataSizeCollected = 0;
         long metadataSizeCollected = 0;
 
+        // only set page token if usePageToken is true
+        // if the range scan completes, set the pageToken to "end of row" value
+        // otherwise set it to a cell name if we hit one of the defensive guards
+
+        CellName firstCell = null;
+        CellName lastScannedCellInContainer = null;
```
inespot: Nitty nit: to be consistent with the terminology below: `lastSeenCell`

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -302,9 +324,33 @@ public void collectReducedColumns(ColumnFamily container, Iterator<Cell> reduced
...
+            {
+                assert cell.name() != firstCell :
+                        "Hit the range scan threshold on the first cell. Either the configured threshold is too low or there are unexpected duplicate cells.";
+                assert lastScannedCellInContainer == null || cell.name() != lastScannedCellInContainer :
+                        "Hit the range scan threshold on a cell that is included in the results set. This should never happen.";
+
+                container.setPageToken(cell);
+                break;
+            }
+
```
inespot, line 354: Partially unrelated to this PR but do you understand the reasoning behind:
```
if (cell.getLocalDeletionTime() < gcBefore)
    continue;
```
Is the guarantee that reducedCells will already ensure overwritten column names are reduced to a single cell?

src/java/org/apache/cassandra/config/Config.java
d-guo: i got this from my experiments where i saw the normal endpoint would regularly time out around 5M cells. that cluster had no load and was over-provisioned so not a very practical estimate. realistically i think we will need to better tune this according to our metrics, and we do have a dashboard for this. we should do that in the sls cassandra.yaml

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
d-guo: mostly just to simplify the feature. i think this makes more sense to keep as a config (as opposed to set by nodetool) since the value used by the node will always be the same as what's in the cassandra.yaml

src/java/org/apache/cassandra/db/ColumnFamily.java
d-guo: yes, you're correct that whenever you set the pageToken, you shouldn't "re-use" the columnfamily again unless you are somehow aware of the page token existing and handle that correctly. the row cache doesn't do this because i didn't implement it, and we don't use the row cache.

> Are there other places in the code where a columnfamily could be re-used?
i'll do another pass at the code here to make sure, but i don't think so

src/java/org/apache/cassandra/db/ColumnFamily.java
d-guo: the code style seems to prefer the brackets? see two methods down
```
    public UUID id()
    {
        return metadata.cfId;
    }
```

src/java/org/apache/cassandra/db/filter/PageToken.java
d-guo: you make a good point to be consistent about these names, and i agree `cell` and `endOfRow` are nicer.

can we do this after all PRs are combined into a single PR to make this easier? otherwise the merging is pretty annoying

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
d-guo: yeah looking at how the `reducedColumns` are gathered, this should never happen, but it is important for how we decide to update the page token here

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
d-guo: changed to `lastSeenCellInContainer`. it's distinguishing that this is the last cell added to the container (aka the results set), which is different from the last cell we've seen, which may not be added to the container

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
d-guo: > if (cell.getLocalDeletionTime() < gcBefore)

yeah exactly. see here https://palantir.quip.com/lqdmAaLVFxI3#temp:C:XRDb3b2086d978b49318f4ccb34c

d-guo: > Is the guarantee that reducedCells will already ensure overwritten column names are reduced to a single cell?

yeah exactly. i also thought this through here: https://palantir.quip.com/lqdmAaLVFxI3#temp:C:XRDb3b2086d978b49318f4ccb34c

test/unit/org/apache/cassandra/db/SerializationsTest.java
d-guo: Need to think about this more, potentially needs actual code changes as well. BLUF form in-person discussion is that upgrading the message version seems to disallow connections with lower version nodes? This means you can't live upgrade the message version, which is unexpected.

src/java/org/apache/cassandra/db/ColumnFamily.java
d-guo: Checked more of the codepaths that we touch and confirmed we don't re-use the column family object anywhere. See here for the details analysis of this code: https://palantir.quip.com/lqdmAaLVFxI3#temp:C:XRD386f4c9309c740d9a62350790

test/unit/org/apache/cassandra/db/SerializationsTest.java
d-guo: while looking at the code with Raymond on this, we figured out how you safely live-upgrade / downgrade. the connection with a lower version node does fail initially, but the handshake gives you info on that node's version. then on the second connection attempt, you know that nodes correct version so the connection succeeds

src/java/org/apache/cassandra/db/ColumnFamilySerializer.java
```diff
@@ -22,6 +22,7 @@
 import java.util.UUID;
 
 import org.apache.cassandra.config.Schema;
+import org.apache.cassandra.db.filter.PageToken;
```
rhuffy, line 25: nit - looks like an unused import - and you probably could revert the fomatting changes (I assume the existing code didn't conform to the format)

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
```diff
@@ -302,9 +324,33 @@ public void collectReducedColumns(ColumnFamily container, Iterator<Cell> reduced
...
+            {
+                firstCell = cell.name();
+            }
+
+            if (usePageToken && hitRangeScanThreshold(reducedCells.deadAndLiveCells()))
+            {
+                assert cell.name() != firstCell :
+                        "Hit the range scan threshold on the first cell. Either the configured threshold is too low or there are unexpected duplicate cells.";
+                assert lastSeenCellInContainer == null || cell.name() != lastSeenCellInContainer :
+                        "Hit the range scan threshold on a cell that is included in the results set. This should never happen.";
```
rhuffy, line 348: How do these error messages get surfaced? Do they make it to safe logs?

rhuffy (approved):
just left some small nits, otherwise LGTM

src/java/org/apache/cassandra/db/ColumnFamily.java
d-guo: i think this is just the oss cassandra style the codebase is already littered with them

src/java/org/apache/cassandra/db/filter/SliceQueryFilter.java
d-guo: probably not to safe logs, but they should show up during testing

PR #634, opened 2025-03-11
[DNM] do not purge tombstones on any cfs being repaired

rhuffy:
Based on #630 

This PR disables the purging of tombstones for any ColumnFamilies that are being repaired.

We assume that our internal cassandra-manager is used, where each repair command only targets a single keyspace+columnFamily. This means that each `ParentRepairSession` only tracks a single `ColumnFamilyStore`, so it is cheap to iterate over `columnFamilyStores.values()`.

### DIFF
--- a/src/java/org/apache/cassandra/service/ActiveRepairService.java
+++ b/src/java/org/apache/cassandra/service/ActiveRepairService.java
@@ -499,6 +499,18 @@ public class ActiveRepairService implements IEndpointStateChangeSubscriber, IFai
         }
     }
 
+    public boolean isRepairing(UUID cfId)
+    {
+        for (ParentRepairSession session : parentRepairSessions.values())
+        {
+            if (session.isRepairing(cfId))
+            {
+                return true;
+            }
+        }
+        return false;
+    }
+
     /**
      * We keep a ParentRepairSession around for the duration of the entire repair, for example, on a 256 token vnode rf=3 cluster
      * we would have 768 RepairSession but only one ParentRepairSession. We use the PRS to avoid anticompacting the sstables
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/service/ActiveRepairService.java
```diff
@@ -499,6 +499,18 @@ public void handleMessage(InetAddress endpoint, RepairMessage message)
         }
     }
 
+    public boolean isRepairing(UUID cfId)
+    {
+        for (ParentRepairSession session : parentRepairSessions.values())
```
inespot, line 504: The ParentRepairSession object is created when a coordinator starts a repair: https://github.com/palantir/cassandra/blob/31dd7de91aaa0bb4db8af82e46fd068817afd5e7/src/java/org/apache/cassandra/repair/RepairRunnable.java#L219 

and when a receiving node gets a repair message: https://github.com/palantir/cassandra/blob/31dd7de91aaa0bb4db8af82e46fd068817afd5e7/src/java/org/apache/cassandra/repair/RepairMessageVerbHandler.java#L84-L89. 

But while the session does seem to be cleaned up on the coordinator's end: [removeParentRepairSession(parentRepairSession)](https://github.com/palantir/cassandra/blob/df3fdf21a944ad1420375421cae24db3f250bbb5/src/java/org/apache/cassandra/service/ActiveRepairService.java#L450), 

I am not seeing it being cleaned up on the receiver end?

rhuffy: We said that Anticompaction does not run because we do not run incremental repairs. This is true, however, the anticompaction messages are still sent, and each node handles the message with the following logic

https://github.com/palantir/cassandra/blob/714d4f92cf4ccd17620d77b5ae885139398a8867/src/java/org/apache/cassandra/service/ActiveRepairService.java#L436-L452

rhuffy: The call stack is 
- `RepairRunnable`
- `ActiveRepairService.instance.finishParentSession`
- for-each neighbor: `AnticompactionTask#run` (sends the message)
- (jump to remote)
- `RepairMessageVerbHandler#doVerb() ANTICOMPACTION_REQUEST`
- `ActiveRepairService.instance.doAntiCompaction`
- removeParentRepairSession

inespot: Niice, sgtm!

PR #635, opened 2025-03-11
[Backport] Fix two issues with Assassinate

rhuffy:
Backports [CASSANDRA-9510](https://issues.apache.org/jira/browse/CASSANDRA-9510) and [CASSANDRA-16847](https://issues.apache.org/jira/browse/CASSANDRA-16847)

- [CASSANDRA-9510](https://issues.apache.org/jira/browse/CASSANDRA-9510) Fixes a NPE error when running `nodetool assassinate`
- [CASSANDRA-16847](https://issues.apache.org/jira/browse/CASSANDRA-16847) nodetool assassinate of a node that failed bootstrap may lead to IndexOutOfBoundsException

### DIFF
--- a/src/java/org/apache/cassandra/gms/Gossiper.java
+++ b/src/java/org/apache/cassandra/gms/Gossiper.java
@@ -602,7 +602,6 @@ public class Gossiper implements IFailureDetectionEventListener, GossiperMBean
     {
         InetAddress endpoint = InetAddress.getByName(address);
         EndpointState epState = endpointStateMap.get(endpoint);
-        Collection<Token> tokens = null;
         logger.warn("Assassinating {} via gossip", SafeArg.of("endpoint", endpoint));
 
         if (epState == null)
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -640,6 +628,21 @@ else if (newState.getHeartBeatState().getHeartBeatVersion() != heartbeat)
...
+                tokens = StorageService.instance.getTokenMetadata().getTokens(endpoint);
+            }
+            catch (Throwable th)
+            {
+                JVMStabilityInspector.inspectThrowable(th);
+            }
+            if (tokens == null || tokens.isEmpty())
+            {
+                logger.warn("Trying to assassinate an endpoint {} that does not have any tokens assigned. This should not have happened, trying to continue with a random token.", SafeArg.of("address", address));
+                tokens = Collections.singletonList(StorageService.getPartitioner().getRandomToken());
```
fsxfreak, line 643: should this make sure that `metadata.getEndpoint(token) != endpoint`? or this is impossible/we don't care either way?

src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -640,6 +628,21 @@ else if (newState.getHeartBeatState().getHeartBeatVersion() != heartbeat)
...
+            {
+                tokens = StorageService.instance.getTokenMetadata().getTokens(endpoint);
+            }
+            catch (Throwable th)
+            {
+                JVMStabilityInspector.inspectThrowable(th);
+            }
+            if (tokens == null || tokens.isEmpty())
+            {
+                logger.warn("Trying to assassinate an endpoint {} that does not have any tokens assigned. This should not have happened, trying to continue with a random token.", SafeArg.of("address", address));
```
fsxfreak, line 642: any way to set up a test to hit this case? or are there already tests that exist but need to be backported?

src/java/org/apache/cassandra/gms/Gossiper.java
rhuffy: Discussed offline. The token value assigned here is not actually used by anything. The logic in StorageService#excise removes state for the tokens in the ApplicationState.TOKENS field, not the tokens defined as part of the `LEFT` state. A token is inserted here just to prevent issues with serialization and deserialization, because in normal operation a Token is always present in this state.

nickbar01234 (approved):

PR #637, opened 2025-03-12
Migrate org.apache.cassandra.net to safe logging

nickbar01234:


src/java/org/apache/cassandra/net/IncomingTcpConnection.java
```diff
@@ -115,7 +117,7 @@ public void close()
         try
         {
             if (logger.isTraceEnabled())
-                logger.trace("Closing socket {} - isclosed: {}", socket, socket.isClosed());
+                logger.trace("Closing socket {} - isclosed: {}", SafeArg.of("socket", socket.getInetAddress()), SafeArg.of("closed", socket.isClosed()));
```
fsxfreak: seems like this should either be
`SafeArg.of("endpoint", socket.getInetAddress())`
or
`SafeArg.of("socket", socket)`
was there a typing issue with the second option?

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -535,8 +537,8 @@ else if (e.getMessage().contains("Cannot assign requested address"))
                 throw new RuntimeException(e);
             }
             String nic = FBUtilities.getNetworkInterface(localEp);
-            logger.info("Starting Messaging Service on {}:{}{}", localEp, DatabaseDescriptor.getStoragePort(),
-                        nic == null? "" : String.format(" (%s)", nic));
+            logger.info("Starting Messaging Service on {}:{}{}", SafeArg.of("localEp", localEp), SafeArg.of("storagePort", DatabaseDescriptor.getStoragePort()),
```
fsxfreak: nit, probably not useful to have a specific key name like `localEp`, I think `endpoint` works fine here

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -723,7 +725,8 @@ public void sendReply(MessageOut message, int id, InetAddress to)
     public void sendOneWay(MessageOut message, int id, InetAddress to)
     {
         if (logger.isTraceEnabled())
-            logger.trace("{} sending {} to {}@{}", FBUtilities.getBroadcastAddress(), message.verb, id, to);
+            logger.trace("{} sending {} to {}@{}", SafeArg.of("broadcastAddress", FBUtilities.getBroadcastAddress()), SafeArg.of("verb", message.verb),
+                         SafeArg.of("id", id), SafeArg.of("to", to));
```
fsxfreak, line 729: same nit about endpoint here

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -798,7 +801,7 @@ public void receive(MessageIn message, int id, long timestamp, boolean isCrossNo
     {
         TraceState state = Tracing.instance.initializeFromMessage(message);
         if (state != null)
-            state.trace("{} message received from {}", message.verb, message.from);
+            state.trace("{} message received from {}", SafeArg.of("verb", message.verb), SafeArg.of("from", message.from));
```
fsxfreak: same nit about endpoint here

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -945,11 +948,10 @@ private void incrementDroppedMessages(DroppedMessages droppedMessages, boolean i
     private void logDroppedMessages()
     {
         List<String> logs = getDroppedMessagesLogs();
-        for (String log : logs)
-            logger.info(log);
-
-        if (logs.size() > 0)
+        if (logs.size() > 0) {
+            logger.info("Dropped messages: {}", SafeArg.of("logs", logs));
```
fsxfreak, line 952: have you confirmed that all of these logs must be safe? if so might be worth making `getDroppedMessagesLogs` return `List<SafeArg<...>`

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -1052,12 +1054,12 @@ public void run()
                 }
                 catch (SSLHandshakeException e)
                 {
-                    logger.debug("SSL handshake error for inbound connection from " + socket, e);
+                    logger.debug("SSL handshake error for inbound connection {}", SafeArg.of("endpoint", socket), e);
```
fsxfreak: conflict here - key is `endpoint` but value is a socket obj

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -1052,12 +1054,12 @@ public void run()
...
                     FileUtils.closeQuietly(socket);
                 }
                 catch (Throwable t)
                 {
-                    logger.trace("Error reading the socket {}", socket, t);
+                    logger.trace("Error reading the socket {}", SafeArg.of("endpoint", socket), t);
```
fsxfreak: same conflict nit here

src/java/org/apache/cassandra/net/MessagingService.java
nickbar01234: Yeah `DroppedMessages` here - basically a counter for a given verb

```java
    private static final class DroppedMessages
    {
        final DroppedMessageMetrics metrics;
        final AtomicInteger droppedInternalTimeout;
        final AtomicInteger droppedCrossNodeTimeout;

        DroppedMessages(Verb verb)
        {
            this.metrics = new DroppedMessageMetrics(verb);
            this.droppedInternalTimeout = new AtomicInteger(0);
            this.droppedCrossNodeTimeout = new AtomicInteger(0);
        }

    }
```

src/java/org/apache/cassandra/net/IncomingTcpConnection.java
nickbar01234: I wasn't thinking too much about the key name -- will update to socket!

src/java/org/apache/cassandra/net/MessagingService.java
nickbar01234: I think it might be cleaner to print one list than logging line by line. Not sure in practice how many messages get dropped though

fsxfreak: agree will be much less spammy to print the whole list once

fsxfreak (approved):
thanks for the update

PR #639, opened 2025-03-12
Range Scan with Page Token (Coordinator)

d-guo:
Continuation of https://github.com/palantir/cassandra/pull/632

This PR implements a row resolver on the coordinator that can handle a page token and partial results set response from each replica. This PR intentionally does not modify the digest and read repair path beyond making it compatible with page tokens for the V0 implementation. Future work loosely scoped [here](https://palantir.quip.com/lqdmAaLVFxI3#temp:C:XRD942209ff1ce244d6b1bc29425) will potentially improve this path for resumable range scans.


src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -21,12 +21,7 @@
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.security.MessageDigest;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
rhuffy: nit: can we avoid the `*` import here? (This was probably done by intellij)

src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -99,6 +98,42 @@ public int liveCQL3RowCount(long now)
         return counter.countAll(this).live();
     }
 
+    public ColumnFamily cloneLimitByPageToken(PageToken pageToken)
+    {
+        ColumnFamily cf = cloneMeShallow();
+        cf.delete(this);
```
rhuffy, line 117: what does it mean to `delete(this)`? I see this in other places in this class.

src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -99,6 +98,42 @@ public int liveCQL3RowCount(long now)
...
+
+        CellNameType comparator = getComparator();
+        Collection<Cell> cells = getSortedColumns();
+
+        for (Iterator<Cell> iter = cells.iterator(); iter.hasNext(); )
+        {
+            Cell cell = iter.next();
+
+            if (pageToken == null || pageToken.isReachedEnd() || comparator.compare(cell.name(), pageToken.getToken().name()) < 0)
+            {
```
rhuffy: noting for my review that the page token here is an exclusive upper bound

src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -209,7 +244,8 @@ public DeletionInfo.InOrderTester inOrderDeletionTester()
 
     public abstract void delete(DeletionInfo info);
     public abstract void delete(DeletionTime deletionTime);
-    protected abstract void delete(RangeTombstone tombstone);
+
+    public abstract void delete(RangeTombstone tombstone);
```
rhuffy: nit whitespace

src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -395,7 +431,7 @@ public boolean equals(Object o)
 
         return metadata.equals(comparison.metadata)
                && deletionInfo().equals(comparison.deletionInfo())
-               && ByteBufferUtil.compareUnsigned(digest(this), digest(comparison)) == 0;
+                && ByteBufferUtil.compareUnsigned(digest(this), digest(comparison)) == 0 && Objects.equals(pageToken, comparison.pageToken);
```
rhuffy: nit: third expression should be on a new line to match formatting

src/java/org/apache/cassandra/db/Row.java
```diff
@@ -55,6 +56,23 @@ public String toString()
...
+        }
+        if (!(o instanceof Row))
+        {
+            return false;
+        }
+
+        Row otherRow = (Row) o;
+
+        return key.equals(otherRow.key) && Objects.equals(cf, otherRow.cf);
+    }
```
rhuffy: why was it necessary to add this method?

src/java/org/apache/cassandra/service/RowDigestResolver.java
```diff
@@ -72,30 +75,41 @@ public Row resolve() throws DigestMismatchException
...
+            {
                 digest = newDigest;
-            else if (!digest.equals(newDigest))
-                throw new DigestMismatchException(key, digest, newDigest);
+                pageTokenDigest = newPageTokenDigest;
+            }
+            else if (!digest.equals(newDigest) || (pageTokenDigest != null && !pageTokenDigest.equals(newPageTokenDigest)))
+            {
+                throw new DigestMismatchException(key, digest, pageTokenDigest, newDigest, newPageTokenDigest);
+            }
```
rhuffy: couldn't the pageTokenDigest be baked in to the existing digest? Inequal digests would still fail this check.

rhuffy (commented):
Much of the branching in this PR is to support a separate page token digest. Why store and send it separately? Why not just incorporate the page token into the existing digest?

At least in this PR, the only place I saw the page token being read is in `RowDataResolver`, and I didn't see where the PageTokenDigest needed to be a separate thing

src/java/org/apache/cassandra/db/ColumnFamily.java
d-guo: it takes the row-level and range tombstones in `this` and adds them to `cf`, reconciling duplicates and overlaps using the `DeletionInfo` data structure
```
    public void delete(DeletionInfo newInfo)
    {
        deletionInfo.add(newInfo);
    }
```
and
```
    public void delete(DeletionInfo info)
    {
        if (info.isLive())
            return;

        // Keeping deletion info for max markedForDeleteAt value
        while (true)
        {
            Holder current = ref;
            DeletionInfo curDelInfo = current.deletionInfo;
            DeletionInfo newDelInfo = info.mayModify(curDelInfo) ? curDelInfo.copy().add(info) : curDelInfo;
            if (refUpdater.compareAndSet(this, current, current.with(newDelInfo)))
                break;
        }
    }
```

src/java/org/apache/cassandra/db/Row.java
d-guo: i think i was using this for testing originally — will remove now it's not necessary

src/java/org/apache/cassandra/service/RowDigestResolver.java
d-guo: i couldn't think of a safe way to bake it in. it isn't enough to just include the page token with `pageToken.updateDigest(digest)` because it doesn't distinguish between `results1 = [c1,c2,c3], pageToken1 = null` and `results2 = [c1,c2], pageToken2 = c3`

src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -122,7 +167,17 @@ public boolean isPageTokenSet()
...
+    {
+        ColumnFamily cf = cloneMeInternal();
+        if (pageToken != null)
+        {
+            cf.setPageToken(pageToken);
+        }
+        return cf;
+    }
+
+    public abstract ColumnFamily cloneMeInternal();
```
rhuffy: Can we tighten visibility on `cloneMeInternal`? I see it's only being called by `cloneMe()` in this class

src/java/org/apache/cassandra/service/RowDigestResolver.java
rhuffy: We should probably talk about this synchronously, because I think the logic here can be greatly simplified if we can figure this out. 

In your example, you're saying that `digest(c1, c2, c3, PageToken(null)) == digest(c1, c2, PageToken(c3)`, which assumes that the digest representation of the PageToken is just the Column it wraps. So `digest(cx) == digest(PageToken(cx))`.

My thinking is that we just need to have `digest(cx) != digest(PageToken(cx))` which could be accomplished by adding some magic constant byte to the digest whenever digesting a page token.

What does `PageToken(null)` mean? Is this the same as `reachedEnd = true`?

src/java/org/apache/cassandra/db/ReadResponse.java
```diff
@@ -67,16 +69,15 @@ public ByteBuffer digest()
...
-            assert digest.equals(this.digest) :
-                String.format("Digest mismatch : %s vs %s",
-                              Arrays.toString(digest.array()),
-                              Arrays.toString(this.digest.array()));
-        }
+        return pageTokenDigest;
+    }
+
+    public synchronized void setDigest(ByteBuffer digest, PageTokenDigest pageTokenDigest)
+    {
```
rhuffy: Why does this need to by synchronized? Are there other similar places where existing code used synchronization?

src/java/org/apache/cassandra/thrift/CassandraServer.java
```diff
@@ -381,6 +381,8 @@ private List<ColumnOrSuperColumn> getSliceInternal(String keyspace,
         return multigetSliceInternal(keyspace, Collections.singletonList(key), column_parent, timestamp, predicate, consistency_level, cState).get(key);
     }
 
+    // implement multiget_slice_paging
+
     public Map<ByteBuffer, List<ColumnOrSuperColumn>> multiget_slice(List<ByteBuffer> keys, ColumnParent column_parent, SlicePredicate predicate, ConsistencyLevel consistency_level)
```
rhuffy, line 386: probably don't need these comments in this PR

rhuffy (commented):
As discussed briefly offline, I think we should consider a simpler implementation that does not use a separate page token digest

src/java/org/apache/cassandra/thrift/CassandraServer.java
d-guo: https://github.com/palantir/cassandra/pull/651 will remove them. they won't appear in the final PR that gets merged

src/java/org/apache/cassandra/db/ReadResponse.java
d-guo: `setDigest` is used in `RowDigestResolver` by `getData` and `resolve`. previously this used an atomic property on the digest field, though i don't fully understand why that was necessary. since we are adding a new digest for the page token, we now have two values that need to be updated atomically

d-guo: and yeah we use synchronized methods in multiple places in the code. searching for `synchronized` in my ide finds this partial list
Screenshot 2025-04-21 at 1 07 23 PM

d-guo:
responding to this comment here bc it was on outdated code so i can't reply anymore: https://github.com/palantir/cassandra/pull/639#discussion_r2050736322

> My thinking is that we just need to have digest(cx) != digest(PageToken(cx)) which could be accomplished by adding some magic constant byte to the digest whenever digesting a page token.

i agree the best approach is something like what you've described. the part i had trouble coming up with is how to guarantee our modification of the digest doesn't increase the chances of a hash collision. i think your proposal of `digest(cx) != digest(PageToken(cx))` is not sufficient for correctness. an easy example is if we add a "magic byte" like `0x01` then `digest(A, B, '0x1', PageToken=null) = digest(A, B, PageToken(reachedEnd=true)`, and this is true for any bytes we choose. i don't think we can say there is a value that will never be the column name.

we know with atlas the column name will always end with a long timestamp, but we shouldn't overly rely on assumptions outside of cassandra unless we make it explicit in cassandra itself ie: assert `columnName != MAGIC_BYTES`, but i'd like to avoid that. i also want to preserve the length of the digest, meaning not append any bytes to the end of it, to maintain backward compatibility. i'm not sure if cassandra makes any assumptions about the length of the digest (it doesn't like look like it does in serialization), but i'd be wary about making changes of that fashion

> What does PageToken(null) mean? Is this the same as reachedEnd = true?

`PageToken(null)` (which i assume you mean `PageToken=null`) means this request didn't use / compute a page token at all aka the behavior for existing endpoints. `PageToken(cell)` and `PageToken(reachedEnd=true)` means the request did use / compute a page token and it was either a cell bc the range query hit the threshold or a `reachedEnd=true` because the query didn't hit the threshold. this is the behavior for non-existing endpoints

src/java/org/apache/cassandra/service/RowDataResolver.java
```diff
@@ -156,7 +163,27 @@ static ColumnFamily resolveSuperset(Iterable<ColumnFamily> versions, long now)
             if (version != null)
                 iters.add(FBUtilities.closeableIterator(version.iterator()));
         filter.collateColumns(resolved, iters, Integer.MIN_VALUE);
-        return ColumnFamilyStore.removeDeleted(resolved, Integer.MIN_VALUE);
+        resolved = ColumnFamilyStore.removeDeleted(resolved, Integer.MIN_VALUE);
```
inespot, line 166: Trying to reason through whether resolving prior to filtering by min page token could create inconsistent results somehow.
Also am I correct in my understanding that no matter what the page token is set to, we return all range tombstones to the coordinator?

src/java/org/apache/cassandra/db/ReadResponse.java
```diff
@@ -34,27 +34,29 @@
 public class ReadResponse
 {
     public static final IVersionedSerializer<ReadResponse> serializer = new ReadResponseSerializer();
-    private static final AtomicReferenceFieldUpdater<ReadResponse, ByteBuffer> digestUpdater = AtomicReferenceFieldUpdater.newUpdater(ReadResponse.class, ByteBuffer.class, "digest");
 
     private final Row row;
     private volatile ByteBuffer digest;
+    private volatile PageTokenDigest pageTokenDigest;
+    // need to add page token here or somehow incorporate into the digest
```
inespot: Is there anything left to do for this comment?

src/java/org/apache/cassandra/db/ReadResponse.java
rhuffy: I don't think this is accomplishing what you intend, because there are non-synchronized accessors for the two digests

src/java/org/apache/cassandra/db/ColumnFamily.java
```diff
@@ -122,7 +167,17 @@ public boolean isPageTokenSet()
...
+    {
+        ColumnFamily cf = cloneMeInternal();
+        if (pageToken != null)
+        {
+            cf.setPageToken(pageToken);
+        }
+        return cf;
+    }
+
+    abstract ColumnFamily cloneMeInternal();
```
rhuffy, line 183: what do you think about instead just implementing the copy constructors to properly set the page token?

src/java/org/apache/cassandra/service/RowDataResolver.java
```diff
@@ -156,7 +163,27 @@ static ColumnFamily resolveSuperset(Iterable<ColumnFamily> versions, long now)
...
+        PageToken resolvedPageToken = resolvedPageToken(versions, resolved);
+        return resolved == null ? null : resolved.cloneMeLimitByPageToken(resolvedPageToken);
+    }
+
+    private static PageToken resolvedPageToken(Iterable<ColumnFamily> versions, ColumnFamily resolved)
+    {
+        List<ColumnFamily> allCfsWithPageTokens = StreamSupport.stream(Iterables.concat(
+                versions,
+                Collections.singleton(resolved)
+        ).spliterator(), false).filter(Objects::nonNull).filter(ColumnFamily::isPageTokenSet).collect(Collectors.toList());
```
rhuffy: I think we should avoid using the stream api in the hot path. We should not need to allocate a collection here.

src/java/org/apache/cassandra/thrift/CassandraServer.java
rhuffy: We should be merging this stack bottom to top though, no? Each PR is squashed to a single commit. Otherwise that would defeat the purpose of stacking PRs

src/java/org/apache/cassandra/db/ColumnFamily.java
d-guo: hm what's the benefit of that over this approach? and it feels a bit weird for the copy constructor to only set the page token but not the actual columns?

src/java/org/apache/cassandra/service/RowDataResolver.java
d-guo: removed the streams usage, but still allocating a collection. i think that's fine because the existing code in `resolveSuperset` and several other places in the hot path already do this

src/java/org/apache/cassandra/service/RowDataResolver.java
d-guo: yep, that's correct. and that's why (as far as i can tell) we are never at risk of returning inconsistent results

src/java/org/apache/cassandra/db/ReadResponse.java
d-guo: oops, it's leftover. i'll remove

src/java/org/apache/cassandra/service/RowDigestResolver.java
d-guo: > My thinking is that we just need to have digest(cx) != digest(PageToken(cx)) which could be accomplished by adding some magic constant byte to the digest whenever digesting a page token.

i agree the best approach is something like what you've described. the part i had trouble coming up with is how to guarantee our modification of the digest doesn't increase the chances of a hash collision. i think your proposal of `digest(cx) != digest(PageToken(cx))` is not sufficient for correctness. an easy example is if we add a "magic byte" like `0x01` then `digest(A, B, '0x1', PageToken=null) = digest(A, B, PageToken(reachedEnd=true)`, and this is true for any bytes we choose. i don't think we can say there is a value that will never be the column name.

we know with atlas the column name will always end with a long timestamp, but we shouldn't overly rely on assumptions outside of cassandra unless we make it explicit in cassandra itself ie: assert `columnName != MAGIC_BYTES`, but i'd like to avoid that. i also want to preserve the length of the digest, meaning not append any bytes to the end of it, to maintain backward compatibility. i'm not sure if cassandra makes any assumptions about the length of the digest (it doesn't like look like it does in serialization), but i'd be wary about making changes of that fashion

> What does PageToken(null) mean? Is this the same as reachedEnd = true?

`PageToken(null)` (which i assume you mean `PageToken=null`) means this request didn't use / compute a page token at all aka the behavior for existing endpoints. `PageToken(cell)` and `PageToken(reachedEnd=true)` means the request did use / compute a page token and it was either a cell bc the range query hit the threshold or a `reachedEnd=true` because the query didn't hit the threshold. this is the behavior for non-existing endpoints

src/java/org/apache/cassandra/db/ReadResponse.java
d-guo: good catch, made those synchronized. should be fine performance-wise given how cheap the methods are

src/java/org/apache/cassandra/thrift/CassandraServer.java
d-guo: i'll be merging these into a separate branch for testing and rebase appropriately. the purpose of stacking the PRs was for easier review, not rollout

src/java/org/apache/cassandra/db/ReadResponse.java
d-guo: changed back to using the atomic ref field updator instead of synchronizing as suggested

src/java/org/apache/cassandra/service/RowDigestResolver.java
d-guo: as discussed async, agreed to keep the current implementation of having the data digest and the page token digest separate

src/java/org/apache/cassandra/db/ReadVerbHandler.java
```diff
@@ -67,7 +68,7 @@ public static ReadResponse getResponse(ReadCommand command, Row row)
     {
         if (command.isDigestQuery())
         {
-            return new ReadResponse(ColumnFamily.digest(row.cf));
+            return new ReadResponse(ColumnFamily.digest(row.cf), (row.cf == null || !row.cf.isPageTokenSet()) ? null : row.cf.pageToken().digest());
         }
```
rhuffy: This null handling happens in at least one other place. Could you move this logic to a factory method like

`PageTokenDigest.of(ColumnFamily)` that cleanly handles the null case?

src/java/org/apache/cassandra/service/RowDigestResolver.java
```diff
@@ -43,7 +45,10 @@ public Row getData()
             if (!result.isDigestQuery())
             {
                 if (result.digest() == null)
-                    result.setDigest(ColumnFamily.digest(result.row().cf));
+                {
+                    result.setDigest(ColumnFamily.digest(result.row().cf), (result.row().cf == null || !result.row().cf.isPageTokenSet()) ? null :
+                            result.row().cf.pageToken().digest());
+                }
```
rhuffy: could use the `PageTokenDigest.of(ColumnFamily)` helper here

src/java/org/apache/cassandra/service/RowDigestResolver.java
```diff
@@ -72,30 +77,41 @@ public Row resolve() throws DigestMismatchException
...
             }
             else
             {
                 // note that this allows for multiple data replies, post-CASSANDRA-5932
                 data = response.row().cf;
                 if (response.digest() == null)
-                    message.payload.setDigest(ColumnFamily.digest(data));
+                {
+                    message.payload.setDigest(ColumnFamily.digest(data), (data == null || !data.isPageTokenSet()) ? null : data.pageToken().digest());
+                }
```
rhuffy: and here

src/java/org/apache/cassandra/service/RowDigestResolver.java
```diff
@@ -72,30 +77,41 @@ public Row resolve() throws DigestMismatchException
...
+            {
                 digest = newDigest;
-            else if (!digest.equals(newDigest))
-                throw new DigestMismatchException(key, digest, newDigest);
+                pageTokenDigest = newPageTokenDigest;
+            }
+            else if (!digest.equals(newDigest) || !Objects.equals(pageTokenDigest, newPageTokenDigest))
+            {
+                throw new DigestMismatchException(key, digest, pageTokenDigest, newDigest, newPageTokenDigest);
+            }
```
rhuffy: Is there a good way to easily surface if the mismatch was due to a digest mismatch or page token digest mismatch? Maybe we can create a `PageTokenDigestMismatchException` that extends `DigestMismatchException`. I'd assume the params here are unsafe.

src/java/org/apache/cassandra/service/RowDataResolver.java
```diff
@@ -114,7 +118,9 @@ public static List<AsyncOneResponse> scheduleRepairs(ColumnFamily resolved, Stri
 
         for (int i = 0; i < versions.size(); i++)
         {
-            ColumnFamily diffCf = ColumnFamily.diff(versions.get(i), resolved);
+            ColumnFamily version = versions.get(i);
+            version = version == null ? null : version.cloneMeLimitByPageToken(resolved == null ? null : resolved.pageToken());
+            ColumnFamily diffCf = ColumnFamily.diff(version, resolved);
```
rhuffy, line 123: What do you think about moving this logic to the implementation of `ColumnFamily.diff`. There's already existing logic to handle the null case there.

src/java/org/apache/cassandra/service/RowDataResolver.java
```diff
@@ -157,7 +163,39 @@ static ColumnFamily resolveSuperset(Iterable<ColumnFamily> versions, long now)
...
         filter.collateColumns(resolved, iters, Integer.MIN_VALUE);
-        return ColumnFamilyStore.removeDeleted(resolved, Integer.MIN_VALUE);
+        resolved = ColumnFamilyStore.removeDeleted(resolved, Integer.MIN_VALUE);
+        PageToken resolvedPageToken = resolvedPageToken(versions, resolved);
+        return resolved == null ? null : resolved.cloneMeLimitByPageToken(resolvedPageToken);
+    }
+
+    private static PageToken resolvedPageToken(Iterable<ColumnFamily> versions, ColumnFamily resolved)
+    {
+
```
rhuffy: nit whitespace

rhuffy (commented):
Left some more comments, mostly style. I'll give another pass on Monday.

src/java/org/apache/cassandra/service/RowDataResolver.java
rhuffy: I don't think I see why we need to allocate here since we're just looking for the minimum. Fair point that they do allocate ArrayLists elsewhere in this code and I'm not sure I fully appreciate why those allocations are necessary, but I do think you can easily avoid these allocations.

rhuffy: For example
```
    private static PageToken resolvedPageToken(Iterable<ColumnFamily> versions, ColumnFamily resolved)
    {
        PageToken resolvedPageToken = resolved.pageToken();
        PageToken.Comparator comparator = null;

        for (ColumnFamily version : versions)
        {
            if (version != null && version.isPageTokenSet())
            {
                if (comparator == null)
                {
                    comparator = new PageToken.Comparator(version.getComparator());
                }
                if (resolvedPageToken == null)
                {
                    resolvedPageToken = version.pageToken();
                }
                else if (comparator.compare(version.pageToken(), resolvedPageToken) < 0)
                {
                    resolvedPageToken = version.pageToken();
                }
            }
        }

        return resolvedPageToken;
    }
```

src/java/org/apache/cassandra/db/ColumnFamily.java
rhuffy: My thinking was that it would be a bit less error prone to set the page token on construction where possible. It looks like `cloneMe` calls `cloneMeInternal` which then does set the actual columns.
```
    private ArrayBackedSortedColumns(ArrayBackedSortedColumns original)
    {
        super(original.metadata);
        this.reversed = original.reversed;
        this.deletionInfo = DeletionInfo.live(); // this is INTENTIONALLY not set to original.deletionInfo.
        this.cells = Arrays.copyOf(original.cells, original.size);
        this.size = original.size;
        this.sortedSize = original.sortedSize;
        this.isSorted = original.isSorted;
    }
```

inlining the superclass constructor
```
    private ArrayBackedSortedColumns(ArrayBackedSortedColumns original)
    {
        assert metadata != null;
        this.metadata = metadata;
        this.reversed = original.reversed;
        this.deletionInfo = DeletionInfo.live(); // this is INTENTIONALLY not set to original.deletionInfo.
        this.cells = Arrays.copyOf(original.cells, original.size);
        this.size = original.size;
        this.sortedSize = original.sortedSize;
        this.isSorted = original.isSorted;
   }
```

From my perspective, it feels surprising that the page token would be set separately.

rhuffy: Looking at this a bit more also raises some questions about why we copy the page token only sometimes.

rhuffy: Actually I think a better question is - are there any instances where we clone a CF and would _not_ want to copy the page token?

src/java/org/apache/cassandra/service/RowDataResolver.java
d-guo: ah the "limit by page token" logic doesn't really belong in `ColumnFamily.diff` though

src/java/org/apache/cassandra/service/RowDigestResolver.java
d-guo: i'll just add a log line above

d-guo: double check me on this but allocation (un)safe args is fine here because it's contained within the bad (and rare) conditional branch

src/java/org/apache/cassandra/db/ReadVerbHandler.java
d-guo: nice, done! + added to the places listed below

src/java/org/apache/cassandra/service/RowDataResolver.java
d-guo: i think they do those allocations bc it makes the code cleaner and it's probably not that expensive. that said, your looping implementation is cleaner than what i have rn, so i'm going to take it

src/java/org/apache/cassandra/db/ColumnFamily.java
d-guo: not that i can think of. i believe all cloning methods in `ColumnFamily` do copy the page token. see `cloneMeShallow`, `cloneMeLimitByPageToken`, and `cloneMe`.

and i think it makes sense to set the page token where it's done now because the page token is a property of the abstract class. why would we leave it up to the implementing class to properly copy the page token? it makes sense for the cells because each implementing class stores the cells differently

src/java/org/apache/cassandra/service/RowDataResolver.java
rhuffy: I think you're right in principle, but `ColumnFamily.diff` is only used here, and in tests, so we could change it. This could also be a good spot to optimize, because you shouldn't need to copy the CF just to apply a page token limit. The implementation of `diff` could just be page token aware.

Maybe we just consider this in a flup PR.

src/java/org/apache/cassandra/service/RowDigestResolver.java
rhuffy: this is fine

src/java/org/apache/cassandra/db/ColumnFamily.java
rhuffy: It's not that we leave it up to the implementation to set the page token, but instead it's that it's forced to be passed in with the call to `super`. The current situation is that the consumer is responsible for setting this field, which would be very easy to miss. 

I think the situation is different if there are cases that you would want to clone the CF without also cloning the page token, but because you say that all cloning operations should copy the page token, then to me it makes more sense to follow exactly the same pattern as CFMetaData.

src/java/org/apache/cassandra/db/ArrayBackedSortedColumns.java
```diff
@@ -101,7 +101,7 @@ public ColumnFamily.Factory getFactory()
         return factory;
     }
 
-    public ColumnFamily cloneMe()
+    public ColumnFamily cloneMeInternal()
```
rhuffy, line 104: nit: package-private

src/java/org/apache/cassandra/db/AtomicBTreeColumns.java
```diff
@@ -135,7 +135,7 @@ public Factory getFactory()
         return factory;
     }
 
-    public ColumnFamily cloneMe()
+    public ColumnFamily cloneMeInternal()
```
rhuffy, line 138: nit: package-private

rhuffy (approved):
just left two last nits. I will not block on the other stylistic comments I had regarding the copy constructor

PR #640, opened 2025-03-14
Add Logging for Compactions Over 5GiB

Iacobacci:
Add logging for compactions where the table is over 5GiB in size. This will be used to help inform operational steps and reduce the need to run `nodetool compactionstats`.

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
@@ -128,8 +128,12 @@ public class CompactionTask extends AbstractCompactionTask
 
         final long expectedWriteSize = checkAvailableDiskSpaceAndGetWriteSize(checkAvailableDiskSpaceFunction);
 
-        if (expectedWriteSize > 5 * 1024 * 1024 * 1024) {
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -128,6 +128,10 @@ protected void runMayThrow() throws Exception
 
         final long expectedWriteSize = checkAvailableDiskSpaceAndGetWriteSize(checkAvailableDiskSpaceFunction);
 
+        if (expectedWriteSize > 5 * 1024 * 1024 * 1024) {
+            logger.info(String.format("Compaction for ks/cf %s/%s exceeds 5GiB with total size of %d", cfs.keyspace, cfs.name, expectedWriteSize));
```
fsxfreak: we want to make the logs safe loggable, something like
```suggestion
            logger.info("Compaction for ks/cf {}/{} exceeds 5GiB with total size of {}", SafeArg.of("keyspace", cfs.keyspace), SafeArg.of("columnFamily", cfs.name), SafeArg.of("size", expectedWriteSize)));
```

rhuffy: also please match the formatting style of cassandra, so `{` on new lines

fsxfreak (approved):
thanks!

Iacobacci:
Updated title and description to match the drop from 200 GiB to 5GiB

PR #641, opened 2025-03-14
[DNM] Flags for disabling read repairs

rhuffy:
Based on #634 and #630 

This PR adds two new config options to `cassandra.yaml`, and their respective `nodetool` commands for checking/setting at runtime.

- `disable_block_on_read_repair` - this coordinator will read-repair as normal, but will not block on ACKs for `READ_REPAIR` verbs before returning to the client. The intention is to reduce the latency of read-repairs, at the cost of read-repairing multiple times if the same cell is read repeatedly.
- `disable_read_repair_mutation`- this coordinator will still resolve a full read in the case of digest mismatches, but will not send `READ_REPAIR` mutation verbs to any replicas. The intention is to reduce the write load on replicas.

### DIFF
--- a/src/java/org/apache/cassandra/db/ReadVerbHandler.java
+++ b/src/java/org/apache/cassandra/db/ReadVerbHandler.java
@@ -49,6 +49,16 @@ public class ReadVerbHandler implements IVerbHandler<ReadCommand>
         MessageOut<ReadResponse> reply = new MessageOut<ReadResponse>(MessagingService.Verb.REQUEST_RESPONSE,
                                                                       getResponse(command, row),
                                                                       ReadResponse.serializer);
+
+        if (command.isDigestQuery())
+        {
+            Keyspace.open(command.ksName).getColumnFamilyStore(command.cfName).metric.digestReads.mark();
+        }
+        else
+        {
+            Keyspace.open(command.ksName).getColumnFamilyStore(command.cfName).metric.dataReads.mark();
+        }
+
         Tracing.trace("Enqueuing response to {}", message.from);
         MessagingService.instance().sendReply(reply, id, message.from);
     }
--- a/src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
+++ b/src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
@@ -190,9 +190,13 @@ public class ColumnFamilyMetrics
     public final Map<ConsistencyLevel, LatencyMetrics> coordinatorScanLatencyByCL;
 
     public final LatencyMetrics blockingReadRepairLatency;
+    public final Meter avoidedReadRepairs;
     public final Meter blockingReadRepairs;
     public final Meter attemptedReadRepairs;
     public final Meter backgroundReadRepairs;
+
+    public final Meter digestReads;
+    public final Meter dataReads;
 
     public final Counter largePartitionsCompacted;
 
@@ -426,9 +430,12 @@ public class ColumnFamilyMetrics
         coordinatorScanLatencyByCL = MetricUtils.byConsistencyLevel(cl -> new LatencyMetrics(factory, "CoordinatorScan", ",consistency=" + cl, coordinatorScanLatency));
 
         blockingReadRepairLatency = new LatencyMetrics(factory, "BlockingReadRepair", new LatencyMetrics(globalNameFactory, "BlockingReadRepair"));
+        avoidedReadRepairs = Metrics.meter(factory.createMetricName("AvoidedReadRepairs"));
         blockingReadRepairs = Metrics.meter(factory.createMetricName("BlockingReadRepairs"));
         backgroundReadRepairs = Metrics.meter(factory.createMetricName("BackgroundReadRepairs"));
         attemptedReadRepairs = Metrics.meter(factory.createMetricName("AttemptedReadRepairs"));
+        digestReads = Metrics.meter(factory.createMetricName("DigestReads"));
+        dataReads = Metrics.meter(factory.createMetricName("DataReads"));
         largePartitionsCompacted = Metrics.counter("LargePartitionsCompacted");
         pendingFlushes = createColumnFamilyCounter("PendingFlushes");
         bytesFlushed = createColumnFamilyCounter("BytesFlushed");
--- a/src/java/org/apache/cassandra/service/RowDataResolver.java
+++ b/src/java/org/apache/cassandra/service/RowDataResolver.java
@@ -26,6 +26,7 @@ import java.util.concurrent.TimeUnit;
 
 import com.google.common.collect.Iterables;
 
+import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.db.*;
 import org.apache.cassandra.db.columniterator.IdentityQueryFilter;
 import org.apache.cassandra.db.filter.IDiskAtomFilter;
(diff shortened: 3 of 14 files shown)


src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1556,9 +1556,12 @@ else if (logger.isDebugEnabled())
...
                         {
-                            // wait for the repair writes to be acknowledged, to minimize impact on any replica that's
-                            // behind on writes in case the out-of-sync row is read multiple times in quick succession
-                            FBUtilities.waitOnFutures(resolver.repairResults, DatabaseDescriptor.getWriteRpcTimeout());
+                            if (!DatabaseDescriptor.getDisableBlockOnReadRepair())
+                            {
+                                // wait for the repair writes to be acknowledged, to minimize impact on any replica that's
+                                // behind on writes in case the out-of-sync row is read multiple times in quick succession
+                                FBUtilities.waitOnFutures(resolver.repairResults, DatabaseDescriptor.getWriteRpcTimeout());
+                            }
```
inespot, line 1565: If that flag is set, do we need to disable waiting on repair results in `getRangeSlice` as well? 
https://github.com/palantir/cassandra/blob/d21f3feefde21b5b35535a2639f1a408eab947bb/src/java/org/apache/cassandra/service/StorageProxy.java#L1970

inespot: Probably worth updating the metrics marking as well `ReadRepairMetrics.repairedBlocking`

src/java/org/apache/cassandra/service/RowDataResolver.java
```diff
@@ -109,6 +110,11 @@ public Row resolve() throws DigestMismatchException
      */
     public static List<AsyncOneResponse> scheduleRepairs(ColumnFamily resolved, String keyspaceName, DecoratedKey key, List<ColumnFamily> versions, List<InetAddress> endpoints)
```
inespot, line 111: I think there are other parts of the code that do async read repairs along the write path: https://github.com/palantir/cassandra/blob/e301a8dfa26197f9585e11011b9a3649e5858f99/src/java/org/apache/cassandra/service/ReadCallback.java#L272. But those are async as far as I can tell and don't seem to be leveraged in the read path. So maybe not worth tackling for now

src/java/org/apache/cassandra/service/StorageProxy.java
rhuffy: You can trace this through `RangeSliceResponseResolver`

src/java/org/apache/cassandra/service/RowDataResolver.java
rhuffy: Yeah, I'm not convinced that disabling async read repairs is even desirable at this point

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1964,7 +1968,10 @@ else if (logger.isDebugEnabled())
 
                 try
                 {
-                    FBUtilities.waitOnFutures(repairResponses, DatabaseDescriptor.getWriteRpcTimeout());
+                    if (!DatabaseDescriptor.getDisableBlockOnReadRepair())
+                    {
+                        FBUtilities.waitOnFutures(repairResponses, DatabaseDescriptor.getWriteRpcTimeout());
+                    }
```
rhuffy, line 1974: here's the addition in `getRangeSlice`

src/java/org/apache/cassandra/tools/NodeTool.java
```diff
@@ -165,7 +165,11 @@ public int execute(String... args)
                 SetWriteDelay.class,
                 SetRangeScanTokenRangesWarnThreshold.class,
                 GetCoerceReadConsistencyAll.class,
-                SetCoerceReadConsistencyAll.class
+                SetCoerceReadConsistencyAll.class,
+                GetDisableReadRepairMutation.class,
+                SetDisableReadRepairMutation.class,
+                GetDisableBlockOnReadRepair.class,
+                SetDisableBlockOnReadRepair.class
```
rhuffy, line 172: supporting in nodetool

src/java/org/apache/cassandra/db/ReadVerbHandler.java
```diff
@@ -49,6 +49,16 @@ public void doVerb(MessageIn<ReadCommand> message, int id)
...
                                                                       ReadResponse.serializer);
+
+        if (command.isDigestQuery())
+        {
+            Keyspace.open(command.ksName).getColumnFamilyStore(command.cfName).metric.digestReads.mark();
+        }
+        else
+        {
+            Keyspace.open(command.ksName).getColumnFamilyStore(command.cfName).metric.dataReads.mark();
+        }
```
rhuffy, line 60: metrics on data vs. digest reads

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1450,6 +1450,7 @@ private static List<Row> fetchRows(List<ReadCommand> initialCommands, Consistenc
                 try
                 {
                     Row row = exec.get();
+                    Keyspace.open(exec.command.ksName).getColumnFamilyStore(exec.command.cfName).metric.avoidedReadRepairs.mark();
```
inespot, line 1453: For future ref, this metric is placed here because exec.get() will throw a `DigestMismatchException` if values do not agree

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
```diff
@@ -426,9 +430,12 @@ public Double getValue()
         coordinatorScanLatencyByCL = MetricUtils.byConsistencyLevel(cl -> new LatencyMetrics(factory, "CoordinatorScan", ",consistency=" + cl, coordinatorScanLatency));
 
         blockingReadRepairLatency = new LatencyMetrics(factory, "BlockingReadRepair", new LatencyMetrics(globalNameFactory, "BlockingReadRepair"));
+        avoidedReadRepairs = Metrics.meter(factory.createMetricName("AvoidedReadRepairs"));
         blockingReadRepairs = Metrics.meter(factory.createMetricName("BlockingReadRepairs"));
         backgroundReadRepairs = Metrics.meter(factory.createMetricName("BackgroundReadRepairs"));
         attemptedReadRepairs = Metrics.meter(factory.createMetricName("AttemptedReadRepairs"));
+        digestReads = Metrics.meter(factory.createMetricName("DigestReads"));
+        dataReads = Metrics.meter(factory.createMetricName("DataReads"));
```
inespot, line 438: Can you add those panels to: https://grafana.palantircloud.com/grafana/d/ceg4st0zdq41sc/23-pds-6618933a-read-repairs?orgId=1&from=now-6h&to=now&timezone=browser&var-apolloenvironmentid=oz-production&var-cluster=foundry&var-host=$__all?

src/java/org/apache/cassandra/service/RowDataResolver.java
inespot: From what I can tell, I think this actually goes back into `RowDataResolver::resolve` code path and from a quick lookup in the code `MessagingService.Verb.READ_REPAIR` is only leveraged in `scheduleRepairs`. So we would effectively be disabling all read repairs on reads or writes (which makes me lean towards you and Lyuben's suggestion to be incremental more, considering we seem to already be throwing around a bunch of async read repairs when we write)

PR #642, opened 2025-03-17
Add partition header when empty columns from stream

inespot:
### Summary
If a stream receives a sstable with a partition key and an empty set of column, eg:
```
{"key": "18668d64625541fd31808201",
 "cells": []},
```
the `appendFromStream` function will result in a corrupted sstable with the Data.db file solely containing an end of row `END_OF_ROW = 0x0000` for that key, while the index file will still refer its position. This PR ensures the resulting sstable is at parity with the input, in case we don't find a better solution to handle this on the sending side.

### Details of the issue
The appendFromStream function takes a key and an input stream and writes the content of the stream to a new sstable, deserializing each component of the row one after the other:
https://github.com/palantir/cassandra/blob/ee8619e8f2d5739e450a3928b06f2ed696765a0e/src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java#L268
When the key has no cells but the row itself is not deleted (deletion info is default), `iter.next()` will return null right away: 
https://github.com/palantir/cassandra/blob/ee8619e8f2d5739e450a3928b06f2ed696765a0e/src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java#L249
The function will then move on to adding the end of row flag, without adding any key header: https://github.com/palantir/cassandra/blob/ee8619e8f2d5739e450a3928b06f2ed696765a0e/src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java#L280 and the key will be added to the index file.

On the next compaction, the reader will see the key in the index file but fail to deserialize ` 0x0000` because it expects a header. Example result:
```
./service/bin/sstable2json lb-1-big-Data.db

[
{"key": "X",
 "metadata": {"deletionInfo": {"markedForDeleteAt":5138089516213600898,"localDeletionTime":829355}},
 "cells": [Exception in thread "main" org.apache.cassandra.io.sstable.CorruptSSTableException: Corrupted: table-e7e3b4385c1d3ddc9f1df12add9e5c8c/lb-1-big-Data.db
    at org.apache.cassandra.io.sstable.SSTableIdentityIterator.hasNext(SSTableIdentityIterator.java:178)
```

### Fix
The fix calls `maybeWriteRowHeader` in all cases once we have finished adding the atoms:
https://github.com/palantir/cassandra/blob/e2e1cb89e7cfb82ba021320becacd1da61f2cd6e/src/java/org/apache/cassandra/db/ColumnIndex.java#L222

The previous code would only do this if the row was entirely deleted
https://github.com/palantir/cassandra/blob/e2e1cb89e7cfb82ba021320becacd1da61f2cd6e/src/java/org/apache/cassandra/db/ColumnIndex.java#L257

which makes sense since we don't actually expect to see any keys with empty cells in practice (root cause for this is still TBD but it seems reads and compactions can handle this state)

`maybeWriteRowHeader` checks if `lastColumn` is null before adding the header. `lastColumn` is updated whenever we add a cell to the output so this function should only write the header if no cells were ever written.

Alternative would be to move the header write fully out of the `ColumnIndex::add` function:
https://github.com/palantir/cassandra/blob/e2e1cb89e7cfb82ba021320becacd1da61f2cd6e/src/java/org/apache/cassandra/db/ColumnIndex.java#L184 
but this is called in a lot of other places so this feels less invasive

### DIFF
--- a/src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
@@ -273,8 +273,7 @@ public class BigTableWriter extends SSTableWriter
                 logger.warn("No cells written for key {} in appendFromStream for keyspace {} and cf {}",
                             UnsafeArg.of("key", key),
                             SafeArg.of("keyspace", metadata.ksName),
-                            SafeArg.of("cf", metadata.cfName),
-                            SafeArg.of(""));
+                            SafeArg.of("cf", metadata.cfName));
             }
             // Adds row header if no cells was written for this key
             columnIndexer.maybeWriteRowHeader();
--- a/src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
@@ -264,9 +264,9 @@ public class BigTableWriter extends SSTableWriter
 
                 columnIndexer.add(atom); // This write the atom on disk too
             }
-            columnIndexer.finishAddingAtoms();
+            columnIndexer.maybeWriteRowHeader();
 
-            columnIndexer.maybeWriteEmptyRowHeader();
+            columnIndexer.finishAddingAtoms();
             dataFile.stream.writeShort(END_OF_ROW);
         }
         catch (IOException e)
--- a/src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
@@ -25,6 +25,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.UnsafeArg;
 import org.apache.cassandra.db.*;
 import org.apache.cassandra.io.sstable.*;
 import org.apache.cassandra.io.sstable.format.SSTableReader;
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
```diff
@@ -264,9 +267,16 @@ public long appendFromStream(DecoratedKey key, CFMetaData metadata, DataInput in
 
                 columnIndexer.add(atom); // This write the atom on disk too
             }
-            columnIndexer.finishAddingAtoms();
+            if(dataFile.getFilePointer() - currentPosition == 0)
+            {
+                logger.warn("No column cells written for key {} in appendFromStream for keyspace {} and cf {}",
```
d-guo: nit: `cells` or `columns` instead of `column cells`

src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
```diff
@@ -264,9 +267,16 @@ public long appendFromStream(DecoratedKey key, CFMetaData metadata, DataInput in
...
+                            UnsafeArg.of("key", key),
+                            SafeArg.of("keyspace", metadata.ksName),
+                            SafeArg.of("cf", metadata.cfName));
+            }
+            columnIndexer.maybeWriteRowHeader();
 
-            columnIndexer.maybeWriteEmptyRowHeader();
+            columnIndexer.finishAddingAtoms();
```
d-guo, line 281: nit: can we move this back to above when we write the row header (and warn)? to minimize the diff

d-guo (dismissed):

d-guo (dismissed):

d-guo (approved):

PR #645, opened 2025-03-18
Fix compaction logging conditional

Iacobacci:
Fix compaction logging conditional as the current version overflows integers and will log compactions exceeding ~1.2GB in size, rather than the expected 5. 

Also, get the keyspace name for formatting, rather than keyspaces currently being appended with `Keyspace(name=`

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
@@ -129,10 +129,10 @@ public class CompactionTask extends AbstractCompactionTask
 
         final long expectedWriteSize = checkAvailableDiskSpaceAndGetWriteSize(checkAvailableDiskSpaceFunction);
 
-        if (expectedWriteSize > 5 * 1024 * 1024 * 1024)
+        if (expectedWriteSize > 5 * 1024 * 1024 * 1024L)
         {
             logger.info("Compaction for ks/cf {}/{} exceeds 5GiB with total size of {}",
-                    SafeArg.of("keyspace", cfs.keyspace),
+                    SafeArg.of("keyspace", cfs.keyspace.getName()),
                     SafeArg.of("columnFamily", cfs.name),
                     SafeArg.of("size", expectedWriteSize));
         }
--- a/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
@@ -61,7 +61,7 @@ public class CompactionTask extends AbstractCompactionTask
     protected static long totalBytesCompacted = 0;
     private CompactionExecutorStatsCollector collector;
     private static final boolean CONSIDER_CONCURRENT_COMPACTIONS = Boolean.getBoolean("palantir_cassandra.consider_concurrent_compactions");
-    private static final fiveGibibytesInBytes = 5 * 1024 * 1024 * 1024L;
+    private static final long fiveGibibytesInBytes = 5 * 1024 * 1024 * 1024L;
 
     public CompactionTask(ColumnFamilyStore cfs, LifecycleTransaction txn, int gcBefore, boolean offline)
     {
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -129,10 +129,10 @@ protected void runMayThrow() throws Exception
 
         final long expectedWriteSize = checkAvailableDiskSpaceAndGetWriteSize(checkAvailableDiskSpaceFunction);
 
-        if (expectedWriteSize > 5 * 1024 * 1024 * 1024)
+        if (expectedWriteSize > 5 * 1024 * 1024 * 1024L)
         {
```
rhuffy: can you make this magic number a `private static final` constant, explicitly as a `long`

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -129,10 +130,10 @@ protected void runMayThrow() throws Exception
 
         final long expectedWriteSize = checkAvailableDiskSpaceAndGetWriteSize(checkAvailableDiskSpaceFunction);
 
-        if (expectedWriteSize > 5 * 1024 * 1024 * 1024)
+        if (expectedWriteSize > FIVE_GIBIBYTES_IN_BYTES)
         {
             logger.info("Compaction for ks/cf {}/{} exceeds 5GiB with total size of {}",
-                    SafeArg.of("keyspace", cfs.keyspace),
+                    SafeArg.of("keyspace", cfs.keyspace.getName()),
                     SafeArg.of("columnFamily", cfs.name),
```
nickbar01234, line 137: Can we also do `cfs.getColumnFamilyName()`?

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -129,10 +130,10 @@ protected void runMayThrow() throws Exception
 
         final long expectedWriteSize = checkAvailableDiskSpaceAndGetWriteSize(checkAvailableDiskSpaceFunction);
 
-        if (expectedWriteSize > 5 * 1024 * 1024 * 1024)
+        if (expectedWriteSize > FIVE_GIBIBYTES_IN_BYTES)
         {
             logger.info("Compaction for ks/cf {}/{} exceeds 5GiB with total size of {}",
```
nickbar01234, line 135: Should this be exposed as metrics instead of logs for better visibility?

Iacobacci: Currently we want this as a log that can be surfaced on dashboards. 

We have already have metrics for KS/CF size, but we want logs for when a particularly large (over 5 GiB) compaction is starting so we can include this as a check in operational workflows.

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
Iacobacci: For CF name is a property: https://javadoc.io/doc/org.apache.cassandra/cassandra-all/2.2.18/org/apache/cassandra/db/ColumnFamilyStore.html#name

For Keyspace, we need to do getName(): https://javadoc.io/doc/org.apache.cassandra/cassandra-all/2.2.18/org/apache/cassandra/db/Keyspace.html#getName--

nickbar01234: Yep, I'm referring to this method [here](https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/db/ColumnFamilyStore.java#L987-L990) for CF class

Iacobacci: Is there a benefit to doing getColumnFamilyName() instead of just cfs.name? In testing the previous version, cfs.name worked but happy to switch it if there's a benefit?

nickbar01234: One benefit I can see is that the implementor can change the underlying representation without changing the client code. In this case though, seems like later version of C* sticks to having it as a public field, so I don't have a strong opinion on this now.

nickbar01234 (approved):
Approved with nit

PR #646, opened 2025-03-24
Delay JMX Initialization

d-guo:
BLUF of the issue is that certain JMX endpoints, such as `/compact` currently can be triggered before the compaction product cleanup has completed via JMX. These can cause the node to initialize an incorrect view of sstables that we have only seen affect snapshots but potentially has other consequences too.

### DIFF
--- a/src/java/org/apache/cassandra/service/CassandraDaemon.java
+++ b/src/java/org/apache/cassandra/service/CassandraDaemon.java
@@ -228,6 +228,8 @@ public class CassandraDaemon
 
         doNotStartupClientInterfacesIfDisabled();
         beginSetupMayThrowSstableException();
+        // It is important to only initialize JMX at this point because `beginSetupMayThrowSstableException` does compaction product cleanup,
+        // which must happen after sstable views have been initialized. Some JMX endpoints could otherwise intialize the views prematurely.
         maybeInitJmx();
         recoverCommitlogAndCompleteSetup();
 
--- a/src/java/org/apache/cassandra/service/CassandraDaemon.java
+++ b/src/java/org/apache/cassandra/service/CassandraDaemon.java
@@ -224,12 +224,12 @@ public class CassandraDaemon
         // This should be the first write to SystemKeyspace (CASSANDRA-11742)
         SystemKeyspace.persistLocalMetadata();
 
-        maybeInitJmx();
-
         Directories.scheduleVerifyingDiskDoesNotExceedThresholdChecks();
 
         doNotStartupClientInterfacesIfDisabled();
-        completeSetupMayThrowSstableException();
+        beginSetupMayThrowSstableException();
+        maybeInitJmx();
+        recoverCommitlogAndCompleteSetup();
 
         logger.debug("Completed CassandraDaemon setup.");
     }
@@ -261,7 +261,8 @@ public class CassandraDaemon
     }
 
     /* This part of setup may throw a CorruptSSTableException. */
-    private void completeSetupMayThrowSstableException() {
+    private void beginSetupMayThrowSstableException()
+    {
         // load schema from disk
         Schema.instance.loadFromDisk();
 
@@ -335,8 +336,6 @@ public class CassandraDaemon
             JVMStabilityInspector.inspectThrowable(t);
             logger.warn("Unable to start GCInspector (currently only supported on the Sun JVM)");
         }
-
-        recoverCommitlogAndCompleteSetup();
     }
 
     private void recoverCommitlogAndCompleteSetup() {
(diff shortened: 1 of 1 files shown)


d-guo:
Successfully tested cluster roll and expansion with this change

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -261,7 +261,8 @@ private void doNotStartupClientInterfacesIfDisabled() {
     }
 
     /* This part of setup may throw a CorruptSSTableException. */
-    private void completeSetupMayThrowSstableException() {
+    private void beginSetupMayThrowSstableException()
```
inespot, line 266: Nit: Should we rename this `loadAndCleanupDiskStateMayThrowSstableException` ? Or sth along those lines?

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -224,12 +224,12 @@ protected void setup()
...
         SystemKeyspace.persistLocalMetadata();
 
-        maybeInitJmx();
-
         Directories.scheduleVerifyingDiskDoesNotExceedThresholdChecks();
 
         doNotStartupClientInterfacesIfDisabled();
-        completeSetupMayThrowSstableException();
+        beginSetupMayThrowSstableException();
+        maybeInitJmx();
```
inespot, line 233: Can you add a small comment on why it's important to init jmx after the previous function? Theoretically if we get rid of product cleanup this won't be an issue anymore but just in case

src/java/org/apache/cassandra/service/CassandraDaemon.java
d-guo: i prefer the `beginSetup` because this method does more work than just disk state like  `store.disableAutoCompaction();` + the commitlog work that comes later is also "disk state" stuff

inespot: Fair enough

inespot:
Aside from the testing Daniel did, on a theoretical level, considering we are only delaying the jmx initialization and that the `beginSetup` function operations do not depend on jmx being initialized, this should be a safe change

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -224,12 +224,14 @@ protected void setup()
...
-        maybeInitJmx();
-
         Directories.scheduleVerifyingDiskDoesNotExceedThresholdChecks();
 
         doNotStartupClientInterfacesIfDisabled();
-        completeSetupMayThrowSstableException();
+        beginSetupMayThrowSstableException();
+        // It is important to only initialize JMX at this point because `beginSetupMayThrowSstableException` does compaction product cleanup,
+        // which must happen after sstable views have been initialized. Some JMX endpoints could otherwise intialize the views prematurely.
+        maybeInitJmx();
```
inespot, line 233: Wondering if we actually wanna enable this even after commit log is recovered considering Cassandra disables autocompaction then?

d-guo: i think it's fine for us to enable it before the commitlog finishes flushing

summarizing the context given in https://issues.apache.org/jira/browse/CASSANDRA-4474 and https://issues.apache.org/jira/browse/CASSANDRA-3181

in CASSANDRA-3181, they added a five min delay before triggering compactions on everything to avoid contention during startup. however, at the end of commitlog replay, the flush triggers a background compaction (this goes through `memtable.cfs.data.permitCompactionOfFlushed` in `PostFlush`) even during the five min delay. disabling autocompactions before commitlog replay prevents this, which is the change in CASSANDRA-4474.

why i think it's fine the initialize JMX where it is rn:
1. disabling autocompaction was just meant to be an optimization during startup, not a correctness guarantee, and it is also no longer true because `enableAutoCompaction` was later refactored to immediately trigger a compaction anyway in CASSANDRA-5074
2. the five min delay starts at `ScheduledExecutors.optionalTasks.scheduleWithFixedDelay(ColumnFamilyStore.getBackgroundCompactionTaskSubmitter(), 5, 1, TimeUnit.MINUTES)`, which happens after `StorageService.instance.initServer()`. even if we wanted to preserve the optimization (which i don't believe we should), we'd need a different way to not hit `/compact` since we need JMX for the bootstrap checks

inespot: Got it! Thanks for the detailed investigation. Considering this, plus the fact that your change is strictly an improvement and is not actually changing prior state regarding commit log replay vs compaction, happy to keep as is for now and we can re-evaluate whether this optimization makes sense in the future

inespot (approved):

PR #647, opened 2025-03-24
Fix schema agreement for token allocator

rhuffy:
Merging https://github.com/palantir/cassandra/pull/626 and https://github.com/palantir/cassandra/pull/623 cause a conflict resulting in test failures.

The new token allocator requires that schema is in agreement before proceeding, but our modified schema agreement check requires the existence of at least one peer endpoint.

Previously this was not an issue because the schema agreement check was run for new non-seed nodes when they bootstrapped, but now the token allocator runs on every new node, including seeds.

This PR removes the requirement for a peer endpoint to exist in the schema agreement check when run by the allocator, but preserves that requirement during the existing bootstrap and decommission paths.

### DIFF
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -692,7 +692,7 @@ public class DatabaseDescriptor
             conf.server_encryption_options = conf.encryption_options;
         }
 
-        if (Boolean.getBoolean("palantir_cassandra.is_new_cluster"))
+        if (getIsNewCluster())
         {
             if (conf.data_file_directories.length == 0)
                 throw new ConfigurationException("At least one DataFileDirectory must be specified", false);
@@ -2105,5 +2105,10 @@ public class DatabaseDescriptor
 
     public static int getWriteDelay() {
         return conf.write_delay_in_s;
+    }
+
+    public static boolean getIsNewCluster()
+    {
+        return Boolean.getBoolean("palantir_cassandra.is_new_cluster");
     }
 }
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -902,14 +902,14 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
         }
     }
 
-    public void waitForSchema(int delay)
+    public void waitForSchema(int delay, boolean requirePeers)
     {
         // first sleep the delay to make sure we see all our peers
         Uninterruptibles.sleepUninterruptibly(delay, TimeUnit.MILLISECONDS);
 
         // if our schema hasn't matched yet, keep sleeping until it does
         // (post CASSANDRA-1391 we don't expect this to be necessary very often, but it doesn't hurt to be careful)
-        SchemaAgreementCheck schemaAgreementCheck = new SchemaAgreementCheck();
+        SchemaAgreementCheck schemaAgreementCheck = new SchemaAgreementCheck(requirePeers);
         List<InetAddress> ignoredEndpoints = replacing && !isReplacingSameAddress() ?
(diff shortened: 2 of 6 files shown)


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2106,4 +2106,9 @@ public static void setWriteDelay(int writeDelayInSeconds) {
     public static int getWriteDelay() {
         return conf.write_delay_in_s;
     }
+
+    public static boolean getIsNewCluster()
+    {
+        return Boolean.getBoolean("palantir_cassandra.is_new_cluster");
+    }
```
inespot, line 2113: StorageService already has an equivalent `isNewCluster` function. Can we somehow merge those two?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -902,14 +902,14 @@ else if (shouldBootstrap())
...
     {
         // first sleep the delay to make sure we see all our peers
         Uninterruptibles.sleepUninterruptibly(delay, TimeUnit.MILLISECONDS);
 
         // if our schema hasn't matched yet, keep sleeping until it does
         // (post CASSANDRA-1391 we don't expect this to be necessary very often, but it doesn't hurt to be careful)
-        SchemaAgreementCheck schemaAgreementCheck = new SchemaAgreementCheck();
+        SchemaAgreementCheck schemaAgreementCheck = new SchemaAgreementCheck(requirePeers);
         List<InetAddress> ignoredEndpoints = replacing && !isReplacingSameAddress() ?
                                              ImmutableList.of(DatabaseDescriptor.getReplaceAddress()) : ImmutableList.of();
```
inespot, line 915: Nit on line 918: indentation

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
rhuffy: done

rhuffy: My preference is to put all of these flags in DatabaseDescriptor, since that seems be most aligned with the existing code

test/unit/com/palantir/cassandra/utils/SchemaAgreementCheckTest.java
```diff
@@ -165,7 +172,8 @@ public void checkSchemaAgreement_failsIfEndpointsAreMissing()
 
         SchemaAgreementCheck schemaAgreementCheck = new SchemaAgreementCheck(() -> schema,
                                                                              () -> ImmutableMap.of(InetAddresses.forString("127.0.0.1"), state).entrySet(),
-                                                                             InetAddresses.forString("127.0.0.1"));
+                                                                             InetAddresses.forString("127.0.0.1"),
+                                                                             true);
```
inespot, line 176: Add a unit test for the requirePeer flag?

rhuffy: [1618109](https://github.com/palantir/cassandra/pull/647/commits/161810938ecb3e09d372d9442042654e2ef96226)

src/java/org/apache/cassandra/service/StorageService.java
rhuffy: [5dfac33](https://github.com/palantir/cassandra/pull/647/commits/5dfac3333b0266ad6b3827b4d3316dba2fe90bb6)

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
rhuffy: [ed21485](https://github.com/palantir/cassandra/pull/647/commits/ed2148501f0f03e64ec49da3d5c2acb6a3670111)

inespot (approved):

PR #648, opened 2025-03-29
Tracking improvement of estimated space necessary for inflight compactions

lyubent:
## Before
The estimation of bytes needed to compact is aggregated for all in-flight compactions but even as the compactions progress, the estimation never drops. This has negative impact on large compactions that build up a lot of temporary files since we end up reserving the space for the estimate but do not account for the temporary files on disk reducing the total available disk space with the outcome being that new compactions cannot be scheduled due to no disk space being available even though there is plenty of disk.

Worst case scenario, for example for a 2T compaction - is that it will be double counted near its completion where the compaction is at 99%. The estimated needed space which will be reserved will sit at 2T and the temporary files will sit at 2T thus reducing the available space for compacting by 4T.

## After
The estimated space is changed to subtract the current space used by in-flight compactions giving a more reliable estimate. Since the estimation isn't always correct we take whatever is **_larger_** of:
1. estimated space - actual live space used by inflight compactions
2. actual live space used by inflight compactions

The outcome is that the reserved disk space needed for compactions becomes closer to what is realistically needed.


src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -496,37 +497,58 @@ static void sortWriteableCandidates(List<DataDirectoryCandidate> candidates, lon
...
 
     public Boolean checkAvailableDiskSpaceConsideringConcurrentCompactions(long estimatedSSTables, long expectedTotalWriteSize)
     {
         synchronized (COMPACTION_LOCK)
         {
-            if (!checkAvailableDiskSpace(estimatedSSTables, expectedTotalWriteSize, expectedSpaceUsedByCompactions))
+            if (!checkAvailableDiskSpace(estimatedSSTables,
+                                         expectedTotalWriteSize,
+                                         expectedSpaceUsedByCompactions,
+                                         CompactionMetrics.getCompactions()
```
d-guo, line 510: Agree with the overall idea. Just two concerns from me:
1. It feels weird to make decisions within the Compaction Manager based on metrics — it removes isolation between the two. Though the compaction throttler does this too, that was our addition and this feels more invasive. I'd prefer if there was a "concurrent compactions tracker" that both the compaction manager and metrics emitter read from. Curious to hear your thoughts here.
2. How reliable is the value from `compactionHolder.getCompactionInfo().getCompleted()`? I don't think we emit this as a metric anywhere right now. Can we sanity check either via logging or (preferably) a dashboard and soak for a couple days to verify it looks reasonable?

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -496,37 +497,58 @@ static void sortWriteableCandidates(List<DataDirectoryCandidate> candidates, lon
...
         }
     }
 
-    private boolean checkAvailableDiskSpace(long estimatedSSTables, long expectedTotalWriteSize, long expectedSpaceUsedByCompactions) {
+    /**
+     * Determines if there is sufficient disk space available to perform a compaction.
+     *
+     * @param estimatedSSTables The estimated number of SSTables expected to be generated as a result of the compaction.
+     * @param expectedTotalWriteSize The total estimated disk space required for all ongoing and pending compactions, in bytes.
+     * @param expectedSpaceUsedByCompactions The estimated disk space, in bytes, needed specifically for the current compaction.
```
d-guo: i think you flipped this and the one before it

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -496,37 +497,58 @@ static void sortWriteableCandidates(List<DataDirectoryCandidate> candidates, lon
...
+     */
+    private boolean checkAvailableDiskSpace(long estimatedSSTables,
+                                            long expectedTotalWriteSize,
+                                            long expectedSpaceUsedByCompactions,
+                                            long liveSpaceUsedByInProgressCompactions)
+    {
         long writeSize = expectedTotalWriteSize / estimatedSSTables;
         long totalAvailable = 0L;
         long totalSpace = 0L;
+        long spaceNeededForInProgressCompactions = Math.max(liveSpaceUsedByInProgressCompactions,
```
d-guo: confused why this is a max and not just `expectedSpaceUsedByCompactions - liveSpaceUsedByInProgressCompactions`

lyubent: Because the estimation for how big the compaction will be is **_an estimation_**. In most cases space used by tmp tables takes up more space than the initial estimation, with this setup we will never end up with a negative number meaning we are underestimating the needed space.

src/java/org/apache/cassandra/db/Directories.java
lyubent: 1 - Sounds like we want to add an abstraction for the sake of abstracting which is fine. I was going for the most simple solution but if we want to add complexity to ensure isolation that makes sense. Building a singleton to track the compactions and then using it to update metrics should be straight forward.
2 - From testing observations, much more accurate than the estimation used to reserve space. Validating this by observing actual telemetry sounds perfectly valid, I'll set something up.

src/java/org/apache/cassandra/db/Directories.java
d-guo: hm, maybe i'm missing the obvious reason here, but why would it be `Math.max(liveSpaceUsedByInProgressCompactions, expectedSpaceUsedByCompactions - liveSpaceUsedByInProgressCompactions);`, instead of `Math.max(0, expectedSpaceUsedByCompactions - liveSpaceUsedByInProgressCompactions);`, or `Math.max(HARDECODED_BUFFER_LIKE_1GB, expectedSpaceUsedByCompactions - liveSpaceUsedByInProgressCompactions);`?

using `liveSpaceUsedByInProgressCompactions` seems like we'd frequently over-estimate the amount of space in-progress compactions are using. for ex when a compaction is `90GB/100GB` complete, this estimate `spaceNeededForInProgressCompactions ` should be `10GB`, not `90GB` like it will be now

d-guo (dismissed):
for RC in case no perms

src/java/org/apache/cassandra/db/compaction/AbstractCompactionTask.java
```diff
@@ -52,11 +52,11 @@ public AbstractCompactionTask(ColumnFamilyStore cfs, LifecycleTransaction transa
     /**
      * executes the task and unmarks sstables compacting
      */
-    public int execute(CompactionExecutorStatsCollector collector)
+    public int execute(CompactionExecutorStatsCollector collector, CompactionTracker tracker)
```
d-guo: adding a `CompactionTracker` here feels rather confusing, since `CompactionMetrics` (subclass of `CompactionExecutorStatsCollector`) also contains the same `CompactionTracker`. is there a way we can avoid introducing a new parameter here and modify `CompactionExecutorStatsCollector` to correctly update its `CompactionTracker`?

lyubent: Doing this in `CompactionExecutorStatsCollector `  would mean we are again hijacking the metrics to do things for tracking compactions.

Side-note: CompactionMetrics isn't a subclass of CompactionExecutorStatsCollector, it implements the CompactionExecutorStatsCollector interface thus it doesn't get access to the tracker field by doing so.

d-guo: ah good point. i didn't realize that. i confused myself on what the original goal here was. i definitely still prefer this separation compared to what was happening before

src/java/org/apache/cassandra/metrics/CompactionMetrics.java
```diff
@@ -78,26 +78,18 @@ public Long getValue()
...
+        activeCompactionsBytesProcessed = Metrics.register(factory.createMetricName("ActiveCompactionsBytesProcessed"), new Gauge<Long>()
+        {
+            public Long getValue()
+            {
+                return tracker.getTotalCompletedBytes();
+            }
+        });
     }
 
     public void finishCompaction(CompactionInfo.Holder ci)
```
d-guo: are we able to move the two metrics `bytesCompacted` and  `totalCompactionsCompleted ` into the `CompactionTracker`? if we can do that i think that gets us the full isolation between the two and allows us to not need to pass the `CompactionMetrics` object around. and we wouldn't need to repeat
```
            metrics.finishCompaction(ci);
            concurrentCompactionTracker.finishCompaction(ci);
```
everywhere, which feels a bit awkward right now

lyubent: Moving those is trivial yes, but isn't the whole point to isolate metrics and the tracking of compactions? This again goes in the other direction.

d-guo: yeah my thought is that `CompactionMetrics` does the metrics setup / reporting / registering and contains a `CompactionTracker` from which it gets all the info. the idea is that `CompactionTracker` "tracks compactions" and has all the relevant information about compactions. `CompactionMetrics` just reports those as relevant metrics, and other classes such as `CompactionManager` can use `CompactionTracker` for whatever info it wants to make decisions

d-guo (approved):

PR #649, opened 2025-03-29
Add verbose option for compactionstats that displays KC/CF:pending task mapping

lyubent:
Adds a verbose option to `nodetool compactionstats`. The verbose option displays the KS/CF : \<number of pending compactions\> mapping. An adjustment of the row width is included to accommodate long keyspace and table names.

There is also a change in behavior, prior to this pr the pending compaction count included running compactions, those are now not part of the pending compaction count. We can add the inflight tasks trivially if desired.

## Before

```shell
./service/bin/nodetool compactionstats -H

pending tasks: 4
                                     id   compaction type                         keyspace                      table   completed       total    unit  progress
   02e24600-0cdb-11f0-9b2f-d974ead2032d        Compaction             metric_aggregator_v1   metrics_by_date_timeline     1.12 KB     1.17 MB   bytes  0.09%
   02ef3e50-0cdb-11f0-9b2f-d974ead2032d        Compaction   transactional_queue_builder_v2              pipeline_lock     0 bytes   640 bytes   bytes  0.00%
Active compaction remaining time :   0h00m00s
```

## After

```shell
./service/bin/nodetool compactionstats -vH

pending tasks: 2
pending queued tasks by ks/cf: token_tracker/sessions_by_date=1, system/compactions_in_progress=1
                                     id   compaction type                         keyspace                      table   completed       total    unit  progress
   02e24600-0cdb-11f0-9b2f-d974ead2032d        Compaction             metric_aggregator_v1   metrics_by_date_timeline     1.12 KB     1.17 MB   bytes  0.09%
   02ef3e50-0cdb-11f0-9b2f-d974ead2032d        Compaction   transactional_queue_builder_v2              pipeline_lock     0 bytes   640 bytes   bytes  0.00%
Active compaction remaining time :   0h00m00s
```


src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -1865,6 +1865,23 @@ public void setMaximumValidatorThreads(int number)
...
+            {
+                int estimatedRemainingTasks = cfs.getCompactionStrategy().getEstimatedRemainingTasks();
+                if (estimatedRemainingTasks > 0)
+                {
+                    pendingCompactions.put(String.format("%s/%s", keyspaceName, cfs.getColumnFamilyName()), estimatedRemainingTasks);
+                }
+            }
+        }
+        return pendingCompactions;
+    }
```
lyubent: This can be a stream but it's ugly due to the nested for, pref to keep it as is

```java
public Map<String, Integer> getPendingTasksByKeyspaceAndTable() {
    return Schema.instance.getKeyspaces().stream()
        .flatMap(keyspaceName -> Keyspace.open(keyspaceName).getColumnFamilyStores().stream()
            .map(cfs -> new AbstractMap.SimpleEntry<>(
                String.format("%s/%s", keyspaceName, cfs.getColumnFamilyName()),
                cfs.getCompactionStrategy().getEstimatedRemainingTasks()
            ))
        )
        .filter(entry -> entry.getValue() > 0)
        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
}
```

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -1865,6 +1865,23 @@ public void setMaximumValidatorThreads(int number)
...
+    {
+        Map<String, Integer> pendingCompactions = new HashMap<>();
+        for (String keyspaceName : Schema.instance.getKeyspaces())
+        {
+            for (ColumnFamilyStore cfs : Keyspace.open(keyspaceName).getColumnFamilyStores())
+            {
+                int estimatedRemainingTasks = cfs.getCompactionStrategy().getEstimatedRemainingTasks();
+                if (estimatedRemainingTasks > 0)
+                {
+                    pendingCompactions.put(String.format("%s/%s", keyspaceName, cfs.getColumnFamilyName()), estimatedRemainingTasks);
```
inespot: It feels like that formatting should live in `CompactionStats` in this function should return a more generic Keyspace/Table key class.

inespot: Nit: maybe worth also de-duping the logic with `CompactionMetrics` ?

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -1865,6 +1865,23 @@ public void setMaximumValidatorThreads(int number)
         validationExecutor.setMaximumPoolSize(number);
     }
 
+    public Map<String, Integer> getPendingTasksByKeyspaceAndTable()
```
inespot: Nit: Cassandra 2.2 still uses ColumnFamily -> Let's make the naming consistent?

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
inespot: Not opposed to keeping the for loop

inespot (dismissed):
Lgtm, let's do another sanity check on a dev stack before merging!

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
lyubent: When the CFS object is sent over jmx a `NotSerializableException` is encountered. I'm looking into a workaround.

lyubent: The ideas:
- I could extend o.a.c.utils.Pair to implement Serializable, clean easy and beautiful but this opens up an object that is used in some of the most critical performance paths to abuse. There's probably a good reason why Pair isn't serializable.
- I could just return an array of strings of length 2 that represents the key and value, this is too unstructured.
- I can create a new wrapper class for the ks / cf pair and make it serializable. I'm going to go with this option, we can structure the ks/cf pairing in a generic object that will be re-usable but do so without modifying the existing Pair object.

conf/cassandra-env.sh
```diff
@@ -328,3 +328,5 @@ JVM_OPTS="$JVM_OPTS -Djava.library.path=$CASSANDRA_HOME/lib/sigar-bin"
 JVM_OPTS="$JVM_OPTS $MX4J_ADDRESS"
 JVM_OPTS="$JVM_OPTS $MX4J_PORT"
 JVM_OPTS="$JVM_OPTS $JVM_EXTRA_OPTS"
+
+JVM_OPTS="$JVM_OPTS -Dpalantir_cassandra.consider_concurrent_compactions=true"
```
inespot: Where does this come from?

conf/cassandra.yaml
```diff
@@ -951,3 +953,8 @@ windows_timer_interval: 1
 
 # Do not try to coalesce messages if we already got that many messages. This should be more than 2 and less than 128.
 # otc_coalescing_enough_coalesced_messages: 8
+data_file_directories:
+    - /Users/ltodorov/Desktop/Dev/palandra/data/data
+persistent_settings_directory: /Users/ltodorov/Desktop/Dev/palandra/data/psd
+commitlog_directory: /Users/ltodorov/Desktop/Dev/palandra/data/commitlog
+saved_caches_directory: /Users/ltodorov/Desktop/Dev/palandra/data/saved_caches
```
inespot: Do we need those? In general, can you explain the changes to the cassandra.yaml file?

src/java/com/palantir/cassandra/utils/SerializablePair.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public T2 getValue() {
+        return value;
+    }
+
+    public static <X, Y> SerializablePair<X, Y> create(X x, Y y) {
+        return new SerializablePair<>(x, y);
+    }
+}
```
inespot: Feels overkill, if ColumnFamilyStore is unfortunately not serializable, we can probably have two functions:
```
public Map<String, Integer> getCompactionsPendingTasksByCf()

public Map<ColumnFamilyStore, Integer> getCompactionsPendingTasks()
```
with the first one calling the second one and using the format you initially defined

conf/cassandra.yaml
lyubent: We don't, these are changes to get things running locally, it was on oversight pushing the configs. I'll remove with a force push.

src/java/com/palantir/cassandra/utils/SerializablePair.java
lyubent: Using two functions makes sense in the scope of the compaction manager but do we want to expose a function that will not (cannot due to serialization issues) be used to the JXM API?

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -1865,6 +1866,25 @@ public void setMaximumValidatorThreads(int number)
...
+    }
+
+    private Map<ColumnFamilyStore, Integer> getPendingCompactionTasks() {
+        return Schema.instance.getKeyspaces().stream()
+                              .flatMap(keyspaceName -> Keyspace.open(keyspaceName).getColumnFamilyStores().stream())
+                              .filter(cfs -> cfs.getCompactionStrategy().getEstimatedRemainingTasks() > 0)
+                              .collect(Collectors.toMap(
+                                  cfs -> cfs,
+                                  cfs -> cfs.getCompactionStrategy().getEstimatedRemainingTasks()
+                              ));
```
inespot: Did you not initially wanna keep the for loop here? I kind of agreed it was more readable

lyubent: There is no longer a need for a map to be nested within the flatmap and I wanted to be consistent in using streams in both getPendingCompactionTasksByKeyspaceColumnFamily and getPendingCompactionTasks.

lyubent: Valid point here mady by Ines is that calling cfs#getCS#getEstimatedRemaingingTasks twice might yield different results if a compaction gets scheduled at the same time. 

```java
private Map<ColumnFamilyStore, Integer> getPendingCompactionTasks() {
    return Schema.instance.getKeyspaces().stream()
                          .flatMap(keyspaceName -> Keyspace.open(keyspaceName).getColumnFamilyStores().stream())
                          .map(cfs -> new AbstractMap.SimpleEntry<>(cfs, cfs.getCompactionStrategy().getEstimatedRemainingTasks()))
                          .filter(entry -> entry.getValue() > 0)
                          .collect(Collectors.toMap(
                              Map.Entry::getKey,
                              Map.Entry::getValue
                          ));
}
```

Fixing this in the stream definitely makes it less readible so going with the nester for instead.

inespot (approved):

PR #650, opened 2025-03-31
Publish test classes as a jar

rhuffy:
This PR publishes test classes in a separate jar called `cassandra-test`, which allows us to run the token allocator tests on our internal partitioner implementations.


build.xml
```diff
@@ -696,7 +696,15 @@
         <scm connection="${scm.connection}" developerConnection="${scm.developerConnection}" url="${scm.url}"/>
   <dependency groupId="com.google.guava" artifactId="guava"/>
```
inespot: Nit: fix indenting, while we are here

build.xml
```diff
@@ -696,7 +696,15 @@
...
-
+      <artifact:pom id="test-pom"
+                    artifactId="cassandra-test"
+                    url="https://cassandra.apache.org"
+                    name="Apache Cassandra">
+      <parent groupId="org.apache.cassandra"
+              artifactId="cassandra-parent"
+              version="${version}"/>
+      <scm connection="${scm.connection}" developerConnection="${scm.developerConnection}" url="${scm.url}"/>
+      </artifact:pom>
```
inespot, line 707: Do we need any dependencies here?

rhuffy: I don't believe so. The dependency works when tested in the sls repo

inespot (dismissed):

inespot (approved):

PR #651, opened 2025-03-31
Range Scan with Page Token (Thrift)

d-guo:
Continuation of https://github.com/palantir/cassandra/pull/632 and https://github.com/palantir/cassandra/pull/639

This PR implements the Thrift endpoints and `CassandraServer` interface for supporting range scans with page tokens by adding three new thrift endpoints that allow clients to limit and resume range scans by returning a page token: `multiget_multislice_paging`, `multiget_slice_paging`, and `get_range_slices_paging`.


interface/cassandra.thrift
```diff
@@ -608,6 +613,23 @@ struct KeyPredicate {
...
 }
 
+/**
+ * A token returned to the client to indicate the starting column for the next query.
+ * If end_of_row is true, then the paging is complete.
+ */
+struct PageToken {
+    1: optional binary column_name,
+    2: required bool end_of_row,
+}
```
rhuffy, line 623: do we need to include end_of_row in the api? I know this matches the serialized format, but is there any additional information communicated here vs an empty column name?

src/java/org/apache/cassandra/thrift/CassandraServer.java
```diff
@@ -155,6 +155,29 @@ public List<ColumnOrSuperColumn> thriftifyColumns(Collection<Cell> cells, boolea
         return thriftColumns;
     }
 
+    public PageResult thriftifyColumnsPaging(Collection<Cell> cells, boolean reverseOrder, long now, org.apache.cassandra.db.filter.PageToken pageToken)
+    {
+        ArrayList<ColumnOrSuperColumn> thriftColumns = new ArrayList<ColumnOrSuperColumn>(cells.size());
+        for (Cell cell : cells)
```
rhuffy: can this method just call and wrap the output of `thriftifyColumns`?

rhuffy (commented):
no major comments here - logic is very straighforward

my only poke is that we might be able to deduplicate some of this logic, provided it doesn't introduce additional allocations or repeated iterations over the same collections, since this code is on the request hot path.

interface/cassandra.thrift
d-guo: having an explicit `end_of_row` feels much more clear to me, though you're probably right we could encode that info in the `column_name` field assuming cassandra doesn't normally allow empty column names

src/java/org/apache/cassandra/thrift/CassandraServer.java
d-guo: ah nice, yes i'll make that change

src/java/org/apache/cassandra/thrift/ThriftValidation.java
```diff
@@ -640,7 +640,32 @@ public static void validateKeyspaceNotSystem(String modifiedKeyspace) throws org
...
 
+    /**
+     * Resumable range scans make some assumptions that may not be strictly required but have not been tested without them holding true.
+     */
+    public static void validateResumableRangeScan(String keyspaceName, String columnFamilyName, List<SlicePredicate> predicates)
+    {
+        for (SlicePredicate predicate : predicates)
+        {
+            assert predicate.isSetSlice_range() && !predicate.isSetColumn_names() : "Resumable range scans only support slice queries";
+            assert !predicate.getSlice_range().isReversed() : "Resumable range scans do not support reversed queries";
```
rhuffy, line 651: same question as in #632 , how do these errors get surfaced and can we ensure they appear in safe logs?

rhuffy (commented):
Is there any good way to reduce the code duplication in `CassandraServer`?

src/java/org/apache/cassandra/thrift/ThriftValidation.java
d-guo: i'm first going to use the assertions for testing (after merging these three PRs into https://github.com/palantir/cassandra/pull/701). when releasing to prod, we'll likely want to change these assertions to throw safe exceptions

d-guo:
some stuff is duplicated to support the new "with page token" classes, but overall i think it's reasonable as it stands

rhuffy (dismissed):

rhuffy (approved):

PR #660, opened 2025-04-04
Add additional logging to segment creation, recycling and replay

inespot:


nickbar01234 (dismissed):
Wondering if we need to add a debug statement to determine the CommitLog global position too?

nickbar01234 (approved):

PR #661, opened 2025-04-04
ColumnFamilyStore uses Safe Logging

d-guo:


src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -630,7 +631,7 @@ public boolean accept(File pathname)
                 continue;
 
             // missing the DATA file! all components are orphaned
-            logger.warn("Removing orphans for {}: {}", desc, components);
+            logger.warn("Removing orphans for {}: {}", UnsafeArg.of("desc", desc), UnsafeArg.of("components", components));
```
andybradshaw: Relativizing these paths should make them safe, yeah? Wondering if it's worth doing that here...

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -647,7 +648,9 @@ public boolean accept(File pathname)
             for (File file : dir.listFiles())
                 if (tmpCacheFilePattern.matcher(file.getName()).matches())
                     if (!file.delete())
-                        logger.warn("could not delete {}", file.getAbsolutePath());
+                    {
+                        logger.warn("could not delete {}", UnsafeArg.of("file", file.getAbsolutePath()));
```
andybradshaw: Same comment as above, relativizing this should make this safe to log

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -1097,8 +1099,8 @@ private void logFlush(String flushReason)
             }
         }
 
-        logger.debug("Enqueuing flush of {} for cause {}: {}", name, flushReason, String.format("%d (%.0f%%) on-heap, %d (%.0f%%) off-heap",
-                                                                     onHeapTotal, onHeapRatio * 100, offHeapTotal, offHeapRatio * 100));
+        logger.debug("Enqueuing flush of {} for cause {}: {}", SafeArg.of("cfName", name), SafeArg.of("flushCause", flushReason), SafeArg.of("heapInfo", String.format("%d (%.0f%%) on-heap, %d (%.0f%%) off-heap",
+                onHeapTotal, onHeapRatio * 100, offHeapTotal, offHeapRatio * 100)));
```
andybradshaw: Should we update this to explicitly log these as params instead of having to parse out the components downstream?
```suggestion
        logger.debug("Enqueuing flush of {} for cause {}: {}", SafeArg.of("cfName", name), SafeArg.of("flushCause", flushReason), SafeArg.of("onHeapTotal", onHeapTotal), SafeArg.of("onHeapPercent", onHeapRatio * 100), SafeArg.of("offHeapTotal", offHeapTotal), SafeArg.of("offHeapPercent", offHeapRatio * 100));
```

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -1169,7 +1171,7 @@ public ReplayPosition call()
 
     public ReplayPosition forceBlockingFlush(String reason)
     {
-        logger.debug("Flushing memtables on cf {} due to {}", name, reason);
+        logger.debug("Flushing memtables on cf {} due to {}", SafeArg.of("cfName", name), UnsafeArg.of("reason", reason));
```
andybradshaw, line 1181: Is this actually unsafe? It looks like it'll only ever be `"Unknown"`, `"Snapshot"`, or `"Truncate"`

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2570,7 +2574,7 @@ public List<Row> search(ExtendedFilter filter)
 
     public List<Row> filter(AbstractScanIterator rowIterator, ExtendedFilter filter)
     {
-        logger.trace("Filtering {} for rows matching {}", rowIterator, filter);
+        logger.trace("Filtering {} for rows matching {}", SafeArg.of("rowIterator", rowIterator), SafeArg.of("filter", filter));
```
andybradshaw: Can this `filter` contain data?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2730,7 +2733,9 @@ public Set<SSTableReader> snapshotWithoutFlush(String snapshotName, Predicate<SS
                     filesJSONArr.add(ssTable.descriptor.relativeFilenameFor(Component.DATA));
 
                     if (logger.isTraceEnabled())
-                        logger.trace("Snapshot for {} keyspace data file {} created in {}", keyspace, ssTable.getFilename(), snapshotDirectory);
+                    {
+                        logger.trace("Snapshot for {} keyspace data file {} created in {}", SafeArg.of("keyspace", keyspace), UnsafeArg.of("dataFile", ssTable.getFilename()), SafeArg.of("snapshotDirectory", snapshotDirectory));
```
andybradshaw: Same thing for relative path

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2773,22 +2778,22 @@ private void createEphemeralSnapshotMarkerFile(final String snapshot)
                 ephemeralSnapshotMarker.getParentFile().mkdirs();
 
             Files.createFile(ephemeralSnapshotMarker.toPath());
-            logger.trace("Created ephemeral snapshot marker file on {}.", ephemeralSnapshotMarker.getAbsolutePath());
+            logger.trace("Created ephemeral snapshot marker file on {}.", UnsafeArg.of("path", ephemeralSnapshotMarker.getAbsolutePath()));
```
andybradshaw: Same thing for relative path

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2773,22 +2778,22 @@ private void createEphemeralSnapshotMarkerFile(final String snapshot)
...
         }
         catch (IOException e)
         {
-            logger.warn(String.format("Could not create marker file %s for ephemeral snapshot %s. " +
+            logger.warn("Could not create marker file {} for ephemeral snapshot {}. " +
                                       "In case there is a failure in the operation that created " +
                                       "this snapshot, you may need to clean it manually afterwards.",
-                                      ephemeralSnapshotMarker.getAbsolutePath(), snapshot), e);
+                    UnsafeArg.of("path", ephemeralSnapshotMarker.getAbsolutePath()), SafeArg.of("snapshot", snapshot), e);
```
andybradshaw: Same thing for relative path

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
d-guo: hm my intellij shows a few more usages + it feels safer for the reason to be unsafe since it's just an unchecked string. we could also consider making this an enum but the existing reasons seem too specific to make an enum

Screenshot 2025-04-06 at 4 09 02 PM

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
d-guo: ah good catch. yes the data range has the two endpoints of keys so this is unsafe

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
d-guo: yeah i think if these paths were relative that would be safe. we do set `SERVICE_HOME` as an environment variable [here](https://github.com/palantir/cassandra/blob/85d61d12453ddd5ca6e873c65ea35e10bc2bea34/bin/cassandra.in.sh#L17)
```
SERVICE_HOME=$(cd "$(dirname "$0")/../../" && pwd)
```
and it's part of the [sls-spec](https://github.palantir.build/deployability/sls-spec/blob/9a69e70866c61808374ede573db4ed3aeab90e42/docs/layout.md#installing-running-and-upgrading-an-sls-service)
> When deployment tooling runs an SLS service, it should set the SERVICE_HOME environment variable and the current working directory to the top-level directory of the service (parent of var) before invoking service/bin/init.sh.

but not sure if we want to rely on that with 100% confidence. instead here we can log the keyspace and columnfamily since that's enough for us to identify the specific sstable

andybradshaw (approved):

PR #663, opened 2025-04-07
SafeLogging for org.apache.cassandra.db.compaction

inespot:


src/java/org/apache/cassandra/db/compaction/CompactionController.java
```diff
@@ -182,7 +190,11 @@ public static Set<SSTableReader> getFullyExpiredSSTables(ColumnFamilyStore cfSto
             else
             {
                logger.trace("Dropping expired SSTable {} (maxLocalDeletionTime={}, gcBefore={})",
```
fsxfreak, line 192: did you mean to change the format string here to match the new params

src/java/org/apache/cassandra/db/compaction/LeveledManifest.java
```diff
@@ -353,12 +374,15 @@ public synchronized CompactionCandidate getCompactionCandidates()
                     int nextLevel = getNextLevel(candidates);
                     candidates = getOverlappingStarvedSSTables(nextLevel, candidates);
                     if (logger.isTraceEnabled())
-                        logger.trace("Compaction candidates for L{} are {}", i, toString(candidates));
+                        logger.trace("Compaction candidates for L{} are {}",
+                                     SafeArg.of("level", i),
+                                     UnsafeArg.of("candidates", toString(candidates)));
```
fsxfreak, line 379: nit: probably don't need `toString` here?

fsxfreak (dismissed):

src/java/org/apache/cassandra/db/compaction/LeveledManifest.java
inespot: Probably not but considering that's how they originally did it, I see no strong reason to change it ¯\_(ツ)_/¯

src/java/org/apache/cassandra/db/compaction/CompactionController.java
inespot: From offline: Not strongly opinionated on whether we should change the format. Pro: it keeps the original message so more searchable, cons: it's less readable I agree

fsxfreak (approved):

PR #665, opened 2025-04-09
Cleanup and scrub propagate replay stats

inespot:
### Context

Unlike compaction, the cleanup operation targets each SSTable individually: [code ref](https://github.com/palantir/cassandra/blob/36c51744810e357bf104c19b6ae2864d1a700a41/src/java/org/apache/cassandra/db/compaction/CompactionManager.java#L511). For each SSTable, Cassandra first checks if it contains any data that is no longer owned by the node. If it does, Cassandra rewrites the SSTable through a compaction process, which also removes all unowned rows from the old SSTable.

Due to what appears to be a bug in the Cassandra implementation, the cleanup code does not propagate the replay statistics from the old SSTable to the writer ([createWriter](https://github.com/palantir/cassandra/blob/36c51744810e357bf104c19b6ae2864d1a700a41/src/java/org/apache/cassandra/db/compaction/CompactionManager.java#L1016) call from cleanup, resulting [SSTableWriter create](https://github.com/palantir/cassandra/blob/297cddf39475a9423ae81a653aae09b97a350424/src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java#L98-L107)). Consequently, the new SSTable loses all replay position information from its predecessor:

```
Minimum replay position: ReplayPosition(segmentId=-1, position=0)
Maximum replay position: ReplayPosition(segmentId=-1, position=0)
```

It seems that Cassandra 5.0 still has the same behavior: https://github.com/apache/cassandra/blob/cassandra-5.0.3/src/java/org/apache/cassandra/db/compaction/CompactionManager.java#L1627.

As a result, after each horizontal expansion operation (typically followed by a cleanup), column families may have their replay position statistics completely erased. This can lead to resurrected data if, on the next restart, the commit log begins replaying data that has already been subject to gc_grace: https://issues.apache.org/jira/browse/CASSANDRA-8498.

### Fix
The codepath targeted by the fix is also used by the scrub operation. This should be the same principle as cleanup.
The reasoning here being that since cleanup solely removes data from SSTables that is no longer owned by the node, and the commit log writes data in continuously increasing positions, the replay statistics will remain valid even after cleanup. Replaying unowned data after the cleanup could actually be dangerous, as it might lead to data resurrection during the next decommissioning process.

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
@@ -45,6 +45,7 @@ import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.config.Schema;
 import org.apache.cassandra.db.*;
 import org.apache.cassandra.db.compaction.CompactionInfo.Holder;
+import org.apache.cassandra.db.composites.CellNameType;
 import org.apache.cassandra.db.index.SecondaryIndexBuilder;
 import org.apache.cassandra.db.lifecycle.LifecycleTransaction;
 import org.apache.cassandra.dht.Bounds;
@@ -1227,12 +1228,17 @@ public class CompactionManager implements CompactionManagerMBean
     {
         FileUtils.createDirectory(compactionFileLocation);
 
-        return SSTableWriter.create(cfs.metadata,
-                                    Descriptor.fromFilename(cfs.getTempSSTablePath(compactionFileLocation)),
+        MetadataCollector collector = new MetadataCollector(Collections.singletonList(sstable),
+                                                            cfs.metadata.comparator,
+                                                            sstable.getSSTableLevel(),
+                                                            false);
+
+        return SSTableWriter.create(Descriptor.fromFilename(cfs.getTempSSTablePath(compactionFileLocation)),
                                     expectedBloomFilterSize,
                                     repairedAt,
-                                    sstable.getSSTableLevel(),
-                                    cfs.partitioner);
+                                    cfs.metadata,
+                                    cfs.partitioner,
+                                    collector);
     }
 
     public static SSTableWriter createWriterForAntiCompaction(ColumnFamilyStore cfs,
--- a/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
@@ -1119,7 +1119,7 @@ public class CompactionManager implements CompactionManagerMBean
     @VisibleForTesting
     void doCleanupOne(final ColumnFamilyStore cfs, LifecycleTransaction txn, Collection<Range<Token>> ranges) throws IOException
     {
-        doCleanupOne(cfs, txn, new CleanupStrategy.Full(cfs, ranges), ranges,false);
+        doCleanupOne(cfs, txn, CleanupStrategy.get(cfs, ranges), ranges,false);
     }
 
     private static abstract class CleanupStrategy
--- a/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
@@ -1114,6 +1114,13 @@ public class CompactionManager implements CompactionManagerMBean
 
     }
 
+    /* Used in tests. */
+    @VisibleForTesting
+    void doCleanupOne(final ColumnFamilyStore cfs, LifecycleTransaction txn, Collection<Range<Token>> ranges) throws IOException
+    {
+        doCleanupOne(cfs, txn, new CleanupStrategy.Full(cfs, ranges), ranges,false);
+    }
+
     private static abstract class CleanupStrategy
     {
         public static CleanupStrategy get(ColumnFamilyStore cfs, Collection<Range<Token>> ranges)
--- a/test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
+++ b/test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
@@ -784,6 +784,7 @@ public class CompactionsTest
 
         ReplayPosition upperBound = sstable.getSSTableMetadata().commitLogUpperBound;
         ReplayPosition lowerBound = sstable.getSSTableMetadata().commitLogLowerBound;
+        assertTrue(lowerBound != ReplayPosition.NONE);
         assertTrue(upperBound != ReplayPosition.NONE);
 
         LifecycleTransaction txn = store.getTracker().tryModify(store.getSSTables(), OperationType.UNKNOWN);
@@ -794,8 +795,8 @@ public class CompactionsTest
         SSTableReader cleanedSstable = store.getSSTables().iterator().next();
 
         assertTrue(cleanedSstable.descriptor.generation != sstable.descriptor.generation);
-        assertTrue(cleanedSstable.getSSTableMetadata().commitLogLowerBound == lowerBound);
-        assertTrue(cleanedSstable.getSSTableMetadata().commitLogUpperBound == upperBound);
+        assertEquals(cleanedSstable.getSSTableMetadata().commitLogLowerBound, lowerBound);
+        assertEquals(cleanedSstable.getSSTableMetadata().commitLogUpperBound, upperBound);
 
     }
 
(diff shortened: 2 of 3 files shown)


test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
```diff
@@ -701,6 +704,102 @@ public void finishCompaction(CompactionInfo.Holder ci) {}
...
+        assertEquals(1, sstables.size());
+        SSTableReader compactedSstable = sstables.iterator().next();
+
+        assertEquals(maxGeneration + 1, compactedSstable.descriptor.generation);
+        assertEquals(minLowerBound, compactedSstable.getSSTableMetadata().commitLogLowerBound);
+        assertEquals(maxUpperBound, compactedSstable.getSSTableMetadata().commitLogUpperBound);
+    }
+
+    @Test
+    public void testCleanupPropagatesReplayPositions() throws IOException
```
inespot, line 765: Failed before the fix (see first commit) and no longer does after

test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
```diff
@@ -74,14 +75,15 @@
...
-    private static final String CF_STANDARD1 = "CF_STANDARD1";
+    private static final String CF_STANDARD1 = "Standard1";
     private static final String CF_STANDARD2 = "Standard2";
     private static final String CF_STANDARD3 = "Standard3";
     private static final String CF_STANDARD4 = "Standard4";
     private static final String CF_STANDARD5 = "Standard5";
     private static final String CF_STANDARD6 = "Standard6";
     private static final String CF_STANDARD7 = "Standard7";
     private static final String CF_STANDARD8 = "Standard8";
+    private static final String CF_STANDARD9 = "Standard9";
```
andybradshaw: Curious what the purpose of adding a new CF is? It looks like a few are already unused outside of adding them to the schema.

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -1220,12 +1228,17 @@ public static SSTableWriter createWriter(ColumnFamilyStore cfs,
     {
         FileUtils.createDirectory(compactionFileLocation);
 
-        return SSTableWriter.create(cfs.metadata,
```
andybradshaw, line 1223: Was trying to find other code paths where we use this method of creating `SSTableWriter`'s... This feels like the most targeted way of handling this (i.e. we know there's a problem with this specific case of not preserving replay timestamps), but should we audit the other code paths where this could happen?

andybradshaw (commented):
Couple questions, but overall makes sense

test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
inespot: Ah, good catch, did not realize that, I'll reuse one of the existing ones

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
inespot: Yep, audited the `SSTableWriter::create` paths as part of the investigation.
- Compactions use the [constructor](https://github.com/palantir/cassandra/blob/297cddf39475a9423ae81a653aae09b97a350424/src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java#L81) that takes the metadatacollector as a parameter and the metadata collector is created from the set of sstables involved in the compaction, which propagates the replay stats ([example](https://github.com/palantir/cassandra/blob/297cddf39475a9423ae81a653aae09b97a350424/src/java/org/apache/cassandra/db/compaction/writers/MaxSSTableSizeWriter.java#L58))
- Memtables also constructs the [metadatacollector](https://github.com/palantir/cassandra/blob/3a4a753cdbbd0e9dfc29d0ad22e808ff0934fb6b/src/java/org/apache/cassandra/db/Memtable.java#L416-L418) separately and set the replay stats manually.
- The stream reader does not set replay stats ([code ref](https://github.com/palantir/cassandra/blob/1ec21266da1e537dbf5455803d4cf92328f897d6/src/java/org/apache/cassandra/streaming/StreamReader.java#L169)) but that's expected since the data is directly streamed from a peer and does not pass through the commit log
- The [AbstractSSTableSimpleWriter](https://github.com/palantir/cassandra/blob/1602e606348959aead18531cb8027afb15f276e7/src/java/org/apache/cassandra/io/sstable/AbstractSSTableSimpleWriter.java#L66-L69) traces back to cql utilities which are not part of the client normal request path
- [SSTableImport](https://github.com/palantir/cassandra/blob/e052f23b23d8fdceece6df2eb9f0f6f873fa8e32/src/java/org/apache/cassandra/tools/SSTableImport.java#L315) imports sstables manually from a json input as far as I understand it and is also not part of the client normal request path

- The upgrader would be the only other sus path: [code ref](https://github.com/palantir/cassandra/blob/7d2fdfeb41eca9badaf10b906b6afe077d166348/src/java/org/apache/cassandra/db/compaction/Upgrader.java#L70) but we never use this and are not planning on using it in the future. Maybe worth adding a comment there though

andybradshaw (dismissed):

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -1114,6 +1115,13 @@ private void doCleanupOne(final ColumnFamilyStore cfs, LifecycleTransaction txn,
 
     }
 
+    /* Used in tests. */
+    @VisibleForTesting
+    void doCleanupOne(final ColumnFamilyStore cfs, LifecycleTransaction txn, Collection<Range<Token>> ranges) throws IOException
+    {
+        doCleanupOne(cfs, txn, new CleanupStrategy.Full(cfs, ranges), ranges,false);
```
nickbar01234: nit: Can probably use `CleanupStraget.get(cfs, ranges)` so that we're aligned with the actual usage

test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
```diff
@@ -701,6 +704,102 @@ public void finishCompaction(CompactionInfo.Holder ci) {}
...
+            insertRowWithKey(i + 100);
+        }
+        store.forceBlockingFlush();
+
+        assertEquals(1, store.getSSTables().size());
+        SSTableReader sstable = store.getSSTables().iterator().next();
+
+        ReplayPosition upperBound = sstable.getSSTableMetadata().commitLogUpperBound;
+        ReplayPosition lowerBound = sstable.getSSTableMetadata().commitLogLowerBound;
+        assertTrue(upperBound != ReplayPosition.NONE);
```
nickbar01234, line 787: Do we also want to assert the same thing for lower bound?

test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
```diff
@@ -701,6 +704,102 @@ public void finishCompaction(CompactionInfo.Holder ci) {}
...
+
+        LifecycleTransaction txn = store.getTracker().tryModify(store.getSSTables(), OperationType.UNKNOWN);
+        Collection<Range<Token>> ranges = makeRanges(100, 109);
+        CompactionManager.instance.doCleanupOne(store, txn, ranges);
+
+        assertEquals(1, store.getSSTables().size());
+        SSTableReader cleanedSstable = store.getSSTables().iterator().next();
+
+        assertTrue(cleanedSstable.descriptor.generation != sstable.descriptor.generation);
+        assertTrue(cleanedSstable.getSSTableMetadata().commitLogLowerBound == lowerBound);
```
nickbar01234: nit: `assertEquals`

nickbar01234:
Given the `MetadataCollector` can have unintended side-effect, curious if we should leave an internal note for Palantir

test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
inespot: I think conceptually we don't have to, cause lowerBound NONE seems valid for our testing but in practice in this case it should be so I'll add it 🤔

nickbar01234 (dismissed):

test/unit/org/apache/cassandra/db/compaction/CompactionsTest.java
inespot: Never mind we use this: https://github.com/palantir/cassandra/blob/a5c11ed74f538a66505035889d44eca9a7f3da4b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java#L3058

nickbar01234 (approved):
🥳

PR #670, opened 2025-04-15
Add tracing to thrift

andybradshaw:
This PR adds [tracing metadata](https://opentelemetry.io/docs/concepts/signals/traces/) structs and arguments to the thrift API, allowing clients to propagate tracing information to C* for use with distributed tracing tools.


interface/cassandra.thrift
```diff
@@ -54,7 +54,7 @@ namespace rb CassandraThrift
 # An effort should be made not to break forward-client-compatibility either
 # (e.g. one should avoid removing obsolete fields from the IDL), but no
 # guarantees in this respect are made by the Cassandra project.
-const string VERSION = "20.1.0-pt2"
+const string VERSION = "20.1.1"
```
andybradshaw, line 57: This change should technically be a "major" thrift version, due to how optionals in function args don't actually work for clients... it's not a wire break, but would require code changes from clients. The other approach would be to add new functions with this arg, but that's quite a bit of duplication. Since there's a finite and known set of clients, I don't think the break is too bad, but could be convinced to do it the other way.

src/java/org/apache/cassandra/tracing/PalantirTracing.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    private static Observability getObservabilityFromTracing(@Nullable TraceMetadata tracing)
+    {
+        return Optional.ofNullable(tracing)
+                       .map(TraceMetadata::isIs_sampled)
+                       .map(sampled -> sampled ? Observability.SAMPLE : Observability.DO_NOT_SAMPLE)
+                       .orElse(Observability.UNDECIDED);
+    }
+}
```
andybradshaw: nit: newline

src/java/org/apache/cassandra/tracing/PalantirTracing.java
```diff
@@ -0,0 +1,67 @@
...
+import com.palantir.tracing.api.SpanType;
+import org.apache.cassandra.thrift.TraceMetadata;
+import org.apache.thrift.annotation.Nullable;
+
+/**
+ * Ported over from UndertowTracing.
+ */
+public final class PalantirTracing
+{
+    private PalantirTracing() {}
```
andybradshaw, line 32: One thing the internal libraries do is to make spans `AutoCloseable`, so you can do try-with-resource statements and not have to worry about forgetting to finish your root span. Would be open to doing that here as well.

andybradshaw (commented):
Some thoughts to seed the discussion

src/java/org/apache/cassandra/tracing/PalantirTracing.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.palantir.tracing.Observability;
+import com.palantir.tracing.Tracer;
+import com.palantir.tracing.Tracers;
+import com.palantir.tracing.api.SpanType;
+import org.apache.cassandra.thrift.TraceMetadata;
+import org.apache.thrift.annotation.Nullable;
+
+/**
+ * Ported over from UndertowTracing.
```
d-guo: could you link the `UndertowTracing` reference class?

d-guo: ah is it this guy? https://github.com/palantir/tracing-java/blob/f6c9cfbf56ae7a485374e6bf8853e070b74ffeba/tracing-undertow/src/main/java/com/palantir/tracing/undertow/UndertowTracing.java

src/java/org/apache/cassandra/tracing/PalantirTracing.java
d-guo: i'll leave it up to you if you want to do that here or in a FLUP PR. agree we should do that in general though, especially once we start adding more granular spans

interface/cassandra.thrift
d-guo: what's the importance of bumping this VERSION? does it actually get checked anywhere or is it more of a bookkeeping thing?

> optionals in function args don't actually work for clients... it's not a wire break

could you expand on this a bit? what doesn't work in clients and why is it not a wire break? from our conversation, i was of the understanding we expect old clients using the new endpoint (upgrade) to essentially put a null `TraceMetadata`, which the code handles, and new clients using the old endpoint (downgrade) to drop the `TraceMetadata` once it reaches the cassandra server

d-guo:
changes look very reasonable and correct to me. happy to approve once we verify the roll on / roll back behaves as we expect

if you agree, let's use the `cassandra-client-lite` service which currently hits `dg-expansion` on imola to verify the upgrade and downgrade are clean. this is easier than relying on atlas/alta endpoints since then we'd need to upgrade thrift in the atlas library first
1. modify `cassandra-client-lite` to run an experiment which just hits one of the old thrift endpoints, like `multiget_slice`, without adding trace metadata
2. upgrade `dg-expansion`. verify `cassandra-client-lite` doesn't complain the new parameters in the thrift endpoints
3. modify `cassandra-client-lite` to add trace metadata on the endpoints
4. downgrade `dg-expansion`. verify `cassandra-client-lite` doesn't complain about the missing parameter in the thrift endpoints and thrift just drops them

interface/cassandra.thrift
andybradshaw: > what's the importance of bumping this VERSION? does it actually get checked anywhere or is it more of a bookkeeping thing?

I believe it's more of a bookkeeping thing, and part of the problem with this is likely that different thrift implementations will handle some of these cases differently. I didn't see any validation that was happening with this version, it just gets persisted to a local table and can be discovered with a client call.

> could you expand on this a bit? what doesn't work in clients and why is it not a wire break?

reading through the generated code, I think I was just expecting this to get typed a bit differently. There are null checks for the trace metadata, and the [expected validation code](https://github.com/palantir/cassandra/pull/670/files#diff-a7aa1b38446544cd8953a81f47a8fbdff5580ef92d8dcdceddfa9d3bd63781ecR465-R474) gets called any time that it's set.

src/java/org/apache/cassandra/tracing/PalantirTracing.java
andybradshaw: It's actually probably not that relevant to mention... will remove the comment.

andybradshaw:
Have tested this internally, verifying:
- new clients are able to provide the tracing information to an old server without issue (forward compatible)
- old clients hitting a new server also works without issue (backwards compatible)
- new clients hitting a new server correctly add traces for the calls instrumented here, currently just the top-level thrift methods

d-guo (approved):

PR #672, opened 2025-04-22
Propagate traces between threads

jackwickham:
Starts from https://github.com/palantir/cassandra/pull/670

If a new FutureTask thread is created, the trace information is lost. This change should propagate the trace information between them.

### DIFF
--- a/src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
+++ b/src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
@@ -17,6 +17,10 @@
  */
 package org.apache.cassandra.concurrent;
 
+import static org.apache.cassandra.tracing.Tracing.isTracing;
+
+import com.palantir.tracing.CloseableSpan;
+import com.palantir.tracing.DetachedSpan;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


wi11dey (dismissed):
rc

src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
```diff
@@ -147,10 +145,12 @@ class FutureTask<T> extends SimpleCondition implements Future<T>, Runnable
         private boolean failure;
         private Object result = this;
         private final Callable<T> callable;
+        private final DetachedSpan detachedSpan;
 
         public FutureTask(Callable<T> callable)
         {
             this.callable = callable;
+            this.detachedSpan = DetachedSpan.start("FutureTask");
```
rynorris, line 153: @jackwickham does this span not need to be completed at some point?

jackwickham: ah you might be right, this is an older version of the lib so I was trying to piece together the behaviour that we need, but actually digging further it looks like it used to be done differently

PR #673, opened 2025-04-23
Simplify and restrict loadNewSstable

inespot:
**Changes to the original loadSstable implementation**

- Stops rewriting sstables when loading them, which messes up with generation/ancestors mapping and can cause Cassandra to erroneously delete sstables as part of compaction product cleanup on startup. See more details in section below.

Since we only use this function in the specific case of loading new sstables on a cluster being restored, this PR also refactors `loadNewSstable` to:
- Fail if `is_new_cluster` is false (this flag is only used when creating a fresh cluster and during restores)
- Assert that the cf view is actually empty before loading the sstable + compareAndSet the fileGeneration atomic int -> prevents generations conflict and bad interactions with existing sstables.

Those two constraints are meant to prevent this command from being run on active production clusters, due potential risks and race conditions that may occur. Indeed, it appears that Cassandra does not synchronize the data tracker for all view changes. Additionally, in versions 4-5, Cassandra has refactored this function to physically separate new SSTables from existing ones.

**OSS Cassandra relevant tickets**

- https://issues.apache.org/jira/browse/CASSANDRA-6719 -> since we enforce emptyCf and throw if existing view, we should not need the new directory/file deconflict
- https://issues.apache.org/jira/browse/CASSANDRA-14310 -> added. We synchronize on data instead of having the global `synchronized`. No other method was using the full object lock so the synchronized statement was only preventing loadNewSstable from being called twice. By synchronizing on `data`, we achieve the same result but also protect against flushes at the same time.

**Reloading sstables in place**

The expectation when using this endpoint is that the sstables being loaded into view come from a consistent snapshot and are not conflicting with each other -> in terms of level overlap, generations conflicts and data (which is the case in Rescue restores).

- Generation is kept -> but that should not cause any conflicts since we assert that there are no sstables in the view and the fileGenerator is 0.
- Ancestors and replay stats are kept -> fixes the compaction cleanup bugs from generations being reshuffled + replay stats should not also not be an issue since the commit log generation is based on current time and therefore always strictly increasing.
- stats metadata are not mutated: https://github.com/palantir/cassandra/blob/849a438690aa97a361227781108cc90355dcbcd9/src/java/org/apache/cassandra/io/sstable/metadata/StatsMetadata.java#L44-L57 includes the sstable level. Since the keyspace/cf is empty and that we expect the loaded sstables to come from the same snapshot, we should not need to modify the sstable level (ie push back all sstables to L0)

**Reasoning on potential races with other sstable/view modifications**

Note that the the current implementation imitates what the ColumnFamilyStore does on initialization, with the couple additional checks mentioned above, considering the node is already up:
https://github.com/palantir/cassandra/blob/a8bf468405d8264ea734b02789d77d57377ab0a9/src/java/org/apache/cassandra/db/ColumnFamilyStore.java#L439-L445

**Testing**

- [x] Verify loadNewSstable on an empty keyspace with empty new sstables is a no-op.
- [x] Verify loadNewSstable on a non-empty cluster with sstable in view fails.
- [x] Verify loadNewSstable on a non-empty cluster with data in commit log fails.
- [x] Verify loadNewSstable on a cluster with is_new_cluster set to false fails.
- [x] Verify loadNewSstable on a empty cluster with file generation number > 0 fails.
- [x] Verify loadNewSstable loads all data into view and does not re-write sstables stats or generation
- [x] Verify loadNewSstable failure followed by node bounce loads all data into view
- [x] Verify loadNewSstable followed by new writes does not overwrite existing data
- [x] Full Rescue restore using this api


src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -803,87 +801,101 @@ public static void removeUnusedSstables(CFMetaData metadata, Map<Integer, UUID>
...
-     * @param cfName The columnFamily name
-     */
-    public static synchronized void loadNewSSTables(String ksName, String cfName)
-    {
-        loadNewSSTables(ksName, cfName, false);
-    }
-
-    /**
-     * See #{@code StorageService.loadNewSSTables(String, String, boolean)} for more info
+     * See #{@code StorageService.loadNewSSTablesWithCount(String, String, boolean)} for more info
```
d-guo: i think you still mean `loadNewSSTables`

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -803,87 +801,101 @@ public static void removeUnusedSstables(CFMetaData metadata, Map<Integer, UUID>
...
      *
      * @param ksName        The keyspace name
      * @param cfName        The columnFamily name
-     * @param assumeCfIsEmpty   Whether or not we can assume the column family is empty before and while loading the new SSTables
+     * @param emptyCf   Whether or not we expect the column family is empty before and while loading the new SSTables
+     *
      */
-    public static synchronized void loadNewSSTables(String ksName, String cfName, boolean assumeCfIsEmpty)
+    public static synchronized void loadNewSSTables(String ksName, String cfName, boolean emptyCf)
```
d-guo: i'm a bit confused about the `emptyCf`/`assumeCfIsEmpty` params. my understanding is they are meant to say whether this column family has already been initialized by the current cassandra process, but tracing this back to where it's called it's always `false` since it comes only from `NodeProbe`. is there something actually validating that no column families have been initialized?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5201,40 +5201,17 @@ public void rescheduleFailedDeletions()
...
-     */
-    public int loadNewSSTablesWithCount(String ksName, String cfName, boolean assumeCfIsEmpty)
-    {
-        //return ColumnFamilyStore.loadNewSSTablesWithCount(ksName, cfName, assumeCfIsEmpty);
-        throw new UnsupportedOperationException("Cannot load SSTables on version 2.2.18-1.165.0+.");
+        if(!isNewCluster())
+            throw new RuntimeException("Failure: is_new_cluster config is false. "
+                                        + "Using loadnewsstable on active production cluster could lead to sstable/data conflicts.");
+        if (!isInitialized())
+            throw new RuntimeException("StorageService not yet initialized, can't load new sstables");
```
d-guo, line 5213: why is this necessary?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -3514,7 +3518,7 @@ public long trueSnapshotsSize()
     }
 
     @VisibleForTesting
-    void resetFileIndexGenerator()
+    public void resetFileIndexGenerator()
```
d-guo: nit: package-private is enough here

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -950,20 +960,14 @@ public synchronized int loadNewSSTablesWithCount(boolean assumeCfIsEmpty)
...
-        logger.info("Loading new SSTables and building secondary indexes for {}/{}: {}", keyspace.getName(), name, newSSTables);
-
         try (Refs<SSTableReader> refs = Refs.ref(newSSTables))
         {
             data.addSSTables(newSSTables);
             indexManager.maybeBuildSecondaryIndexes(newSSTables, indexManager.allIndexesNames());
         }
-
-        logger.info("Done loading load new SSTables for {}/{}", SafeArg.of("keyspace", keyspace.getName()), SafeArg.of("cfName", name));
-        return newSSTables.size();
```
d-guo, line 966: why'd you remove these logs? most of them seem pretty useful

d-guo: i guess it doesn't matter in this method though

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -803,87 +801,101 @@ public static void removeUnusedSstables(CFMetaData metadata, Map<Integer, UUID>
...
-     *
-     * @return the number of new sstables loaded
-     */
-    public static synchronized int loadNewSSTablesWithCount(String ksName, String cfName)
+    public void loadNewSSTables()
     {
-        return loadNewSSTablesWithCount(ksName, cfName, false);
-    }
+        // Prevents exterior flush or memtable from changing underneath us.
+        synchronized (data)
```
d-guo, line 819: i see a few uses and accesses of `data` which do not synchronize on it, such as `data.addInitialSSTables(sstables);` in the constructor. are these expected and okay?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -803,87 +801,101 @@ public static void removeUnusedSstables(CFMetaData metadata, Map<Integer, UUID>
...
-    /**
-     * See #{@code StorageService.loadNewSSTablesWithCount(String, String)} for more info
-     *
-     * @param ksName The keyspace name
-     * @param cfName The columnFamily name
-     *
-     * @return the number of new sstables loaded
-     */
-    public static synchronized int loadNewSSTablesWithCount(String ksName, String cfName)
+    public void loadNewSSTables()
```
d-guo, line 816: from the description

> Note that the the current implementation imitates what the ColumnFamilyStore does on initialization, with the couple additional checks mentioned above, considering the node is already up

can you point to where the equivalent logic happens on initialization? (i'm assuming you were referring to what happens in the current SOP when we bounce every cassandra node)

src/java/org/apache/cassandra/service/StorageService.java
inespot: https://issues.apache.org/jira/browse/CASSANDRA-14310

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
inespot: Indeed, this method should only be used in testing

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
inespot: Nice catch!

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
inespot: Discussed offline. 
The original purpose of this parameter was to indicate whether the `loadNewSstable` endpoint was invoked on a columnFamily with no existing sstables. I initially intended to keep the MBean api as is, but since the function implementation currently throws an error however we used and the sidecar actually calls loadNewSSTablesWithCount -- which was removed --- I now think retaining this boolean adds unnecessary complexity.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
inespot: Yep, referring to this constructor code:
https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/db/ColumnFamilyStore.java#L439-L446 which gets called on startup: https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/service/CassandraDaemon.java#L313.

The constructor code will list all sstable in directory and add them to the view

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
inespot: The constructor code should get called when initializing keyspaces in CassandraDaemon before the `initServer` is triggered (and the StorageService is initialized). 
More generally though, we consider the existing `data` synchronization across the ColumnFamilyStore to be imperfect, which is why we are both restricting the usage of this function to "new clusters" (ie non production ones) and using a `compareAndSet` for the fileGenerator. 
I think we actually could further strengthen this implementation by extending the Tracker api to have a `addSsTablesToEmptyView` which would itself do a compareAndSet on the view. 
Crashing after setting the file generator should be ok, as it will just increment sstable generation to a higher number (which could happen anyway if you compact everything away)

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -697,23 +697,23 @@ else if (conf.num_tokens > MAX_NUM_TOKENS)
...
+//                    && StorageService.joinRing)
+//                {
+//                    throw new ConfigurationException("is_new_cluster flag is still set to true at least 4 days after cluster creation."
+//                        + " You must remove this flag from configuration as it could cause SEVERE DATA CORRUPTION.", false);
+//                }
+//            }
+//            catch (IOException e)
+//            {
+//                logger.info("Unable to verify validity of palantir_cassandra.is_new_cluster flag. Will retry at next restart.");
+//            }
```
d-guo: is this just for testing?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -805,85 +803,85 @@ public static void removeUnusedSstables(CFMetaData metadata, Map<Integer, UUID>
...
     {
-        loadNewSSTables(false);
-    }
+        // Prevents exterior flush or memtable from changing underneath us.
+        synchronized (data)
+        {
+            logger.info("Loading new SSTables for {}/{}...",
+                        SafeArg.of("keyspace", keyspace.getName()),
+                        SafeArg.of("cfName", name));
+            forceBlockingFlush("Flush pre-load new sstable");
```
d-guo, line 824: this will be an issue if we ever do commitlog archival + restore but should be fine now

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -805,85 +803,85 @@ public static void removeUnusedSstables(CFMetaData metadata, Map<Integer, UUID>
...
-     */
-    public synchronized int loadNewSSTablesWithCount()
-    {
-        return loadNewSSTablesWithCount(false);
+            logger.info("Loading new SSTables {} and building secondary indexes for {}/{}",
+                        SafeArg.of("keyspace", keyspace.getName()),
+                        SafeArg.of("cf", name),
+                        SafeArg.of("generations", generations));
+
+            if(!fileIndexGenerator.compareAndSet(0, generations.isEmpty() ? 0 : generations.get(generations.size() - 1))) {
```
d-guo: `generations.isEmpty()` is impossible here, right?

src/java/org/apache/cassandra/db/lifecycle/Tracker.java
```diff
@@ -202,6 +202,21 @@ public void addSSTables(Iterable<SSTableReader> sstables)
         }
     }
 
+    public void addSSTablesToEmptyView(Iterable<SSTableReader> sstables)
```
d-guo, line 205: nit: `addInitialSSTablesToEmptyView`

src/java/org/apache/cassandra/db/lifecycle/Tracker.java
```diff
@@ -202,6 +202,21 @@ public void addSSTables(Iterable<SSTableReader> sstables)
...
+            setupKeycache(sstables);
+        if (apply(view -> view.sstables.isEmpty(), updateLiveSet(emptySet(), sstables)) == null)
+        {
+            throw new RuntimeException("addSSTablesToEmptyView called on data tracker with non-empty view");
+        }
+        maybeFail(updateSizeTracking(emptySet(), sstables, null));
+        for (SSTableReader sstable : sstables)
+        {
+            notifyAdded(sstable);
+        }
```
d-guo, line 217: why are notifications necessary here vs in `addInitialSSTables`?

d-guo (dismissed):
approving, since only minor comments remaining. would like the detailed testing to be complete before merging, especially

> Full Rescue restore using this api

the PR seems to be correct to me as long as all the assumptions that are asserted in code hold. i think an important part of testing for this PR is that these assumptions do hold in our DR SOP so that we don't end up struggling to restore a cluster because we keep hitting exceptions that we're assuming hold in this PR

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
inespot: Yes, see commit message: https://github.com/palantir/cassandra/pull/673/commits/6aea99d59d1b39653ec71230f160fe438eb4e9d8

src/java/org/apache/cassandra/db/lifecycle/Tracker.java
inespot: Your ColumnFamilyStore has already been constructed at that point and `compactionStrategyWrapper` subscribes to the view changes.
https://github.com/palantir/cassandra/blob/05bf1753ce324769903d38b37b30411ab1c85c89/src/java/org/apache/cassandra/db/compaction/WrappingCompactionStrategy.java#L278

src/java/org/apache/cassandra/db/lifecycle/Tracker.java
inespot: Any reason for this? Since this is not used in the column family constructor the `Initial` does not seem appropriate?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
inespot: Can you detail a little bit more what you mean?

d-guo: by commitlog archival + restore, i mean if we ever want to backup + restore the commitlog. since if we restore the commitlog, this flush will create sstables, our assumptions about not having any sstables in the view will be invalid

src/java/org/apache/cassandra/db/lifecycle/Tracker.java
d-guo: good point. my reasoning was that this method is essentially the same as `addInitialSSTables` except validating that the view is empty, but i see your point given the comment below about notifications

d-guo (dismissed):

inespot:
Run a test restore on IL which went through the `ReloadSstablesTask` successfully. Verified the cassandra data + logs, which showed a bunch of:
Screenshot 2025-05-30 at 12 05 44 AM

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
inespot: I guess it will depend on how we order the sstable load and the commit log. If we get to this, I would expect it to be quite a refactor though, so we'll probably reconsider this function at the same time

d-guo (dismissed):

d-guo (approved):

PR #674, opened 2025-04-23
Add config to always submit read/writes to the SEP asynchronously

rynorris:
Status of changes in this branch:

1. **Propagate trace information through thrift** -> Working well, should move forward with merging to trunk, but needs a way to disable so we can confidently roll it out without enabling everywhere initially to gauge perf impact. Also we should fix the fact that alta downsampled traces still trace in C*, this is too spammy.
2. **Propagate trace information through inter-node RPCs** -> Working well-ish, but parent spans get severed, needs a little more work, then proceed to get it into trunk.
3. **Propagate trace information through async tasks** -> Yes we need this.
4. **Add a ton of spans everywhere** -> We want this in trunk, these are incredibly useful.
5. **Always submit async writes** -> This didn't show a huge impact on mojito, but I think we do want it, it seems obviously better. Perhaps more testing required.
6. **Always submit async reads** -> This is even murkier than the writes. It's not even clear that reads get blocked anyway, so lower priority.
7. **Sync commit log waiters asynchronously** -> I didn't observe an obvious improvement here, although it feels to me like it should be better. More testing required. We should expose the metric for how long writers spend waiting for sync.
8. **ProgressWaitQueue** -> Again, I think this should be strictly better than the existing WaitQueue, but didn't observe an obvious gain. Need to test it again with improved metrics.
9. **Group writes by row** -> Didn't observe any change here. None of my spans showed any actual grouping happening, I think these mutations are already getting groups by row? Probably skip this one.
10. **Add more commitlog metrics** -> Yes, want to merge to trunk.


wi11dey (dismissed):
rc

wi11dey (dismissed):
rc again

wi11dey (dismissed):
rc

wi11dey (dismissed):
rc

fsamuel-bs (dismissed):

fsamuel-bs (dismissed):

PR #675, opened 2025-04-23
batched writes

fsamuel-bs:


wi11dey (dismissed):
rc

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -572,19 +572,46 @@ public static void mutate(Collection<? extends IMutation> mutations, Consistency
...
                 {
-                    WriteType wt = mutations.size() <= 1 ? WriteType.SIMPLE : WriteType.UNLOGGED_BATCH;
-                    responseHandlers.add(performWrite(mutation, consistency_level, localDataCenter, standardWritePerformer, null, wt));
+                    if (nonCounterMutations == null)
+                    {
+                        nonCounterMutations = mutation;
+                    }
+                    else
+                    {
+                        nonCounterMutations.addAll(mutation);
```
fsamuel-bs, line 600: This doesn't work - writes need to be for the same key

PR #679, opened 2025-04-30
Coerce consistency level only for specified keyspaces

andybradshaw:
Effectively the same change as in #678, this PR is based on a slightly later commit in the branch with all of these features (notably, the blocking read repair flags), and adds the ability to modify these settings with some additional nodetool commands, so restarts aren't necessary to update which keyspaces this applies to.

### DIFF
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -2114,7 +2114,9 @@ public class DatabaseDescriptor
     }
 
     public static boolean getCoerceReadConsistencyAllForKeyspace(String keyspace) {
-        return conf.coerce_read_consistency_all && conf.coerce_read_consistency_all_keyspaces.contains(keyspace);
+        return conf.coerce_read_consistency_all &&
+               (conf.coerce_read_consistency_all_keyspaces.isEmpty() ||
+                conf.coerce_read_consistency_all_keyspaces.contains(keyspace));
     }
 
     public static void setCoerceReadConsistencyAllForKeyspace(Set<String> keyspaces) {
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -60,6 +60,7 @@ import org.apache.cassandra.service.StorageService;
 import org.apache.cassandra.thrift.ThriftServer;
 import org.apache.cassandra.utils.FBUtilities;
 import org.apache.cassandra.utils.memory.*;
+import org.assertj.core.util.Sets;
 
 import static org.apache.cassandra.io.util.FileUtils.ONE_GB;
 import static org.apache.cassandra.io.util.FileUtils.ONE_MB;
@@ -111,6 +112,7 @@ public class DatabaseDescriptor
     private static boolean hasLoggedConfig;
 
     private static final String allHostsRaw = System.getProperty("palantir_cassandra.all_hosts", "");
+    private static final Set<String> coerceReadConsistencyIgnoredKeyspaces = Sets.newHashSet(Arrays.asList(conf.coerce_read_consistency_all_ignored_keyspaces));
 
     private static boolean daemonInitialized;
 
@@ -2105,6 +2107,14 @@ public class DatabaseDescriptor
     public static void setCoerceReadConsistencyAll(boolean value)
     {
         conf.coerce_read_consistency_all = value;
+    }
+
+    public static boolean getCoerceReadConsistencyAllForKeyspace(String keyspace) {
+        return conf.coerce_read_consistency_all && !coerceReadConsistencyIgnoredKeyspaces.contains(keyspace);
+    }
+
+    public static boolean setCoerceReadConsistencyAllForKeyspace(String keyspace) {
+        return coerceReadConsistencyIgnoredKeyspaces.add(keyspace);
     }
 
     public static boolean getDisableReadRepairMutation()
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -2103,6 +2103,11 @@ public class DatabaseDescriptor
         return conf.coerce_read_consistency_all;
     }
 
+    public static Set<String> getCoerceReadConsistencyAllKeyspaces()
+    {
+        return conf.coerce_read_consistency_all_keyspaces;
+    }
+
     public static void setCoerceReadConsistencyAll(boolean value)
     {
         conf.coerce_read_consistency_all = value;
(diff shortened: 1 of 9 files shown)


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2102,11 +2103,24 @@ public static boolean getCoerceReadConsistencyAll()
...
+        return conf.coerce_read_consistency_all_keyspaces;
+    }
+
     public static void setCoerceReadConsistencyAll(boolean value)
     {
         conf.coerce_read_consistency_all = value;
     }
 
+    public static boolean getCoerceReadConsistencyAllForKeyspace(String keyspace) {
+        return conf.coerce_read_consistency_all && conf.coerce_read_consistency_all_keyspaces.contains(keyspace);
```
inespot: Might be more usable to make `coerce_read_consistency_all_keyspaces` default to "every keyspaces" if empty

src/java/org/apache/cassandra/tools/nodetool/GetCoerceReadConsistencyAllKeyspaces.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.apache.cassandra.tools.nodetool;
+
+import io.airlift.command.Command;
+import org.apache.cassandra.tools.NodeProbe;
+import org.apache.cassandra.tools.NodeTool;
+
+@Command(name = "getcoercereadconsistencyallkeyspaces", description = "Get the coerce read consistency level, if present")
+public class GetCoerceReadConsistencyAllKeyspaces extends NodeTool.NodeToolCmd
```
inespot, line 29: Depending on how easy the plumbing is, wanna add an optional `--keyspace` fields that focuses on whether a single keyspace is coerced at all?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
inespot: Side note, if we really wanted to merge this, we would probably unify those two fields into one

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1300,16 +1299,15 @@ private static List<Row> readWithPaxos(List<ReadCommand> commands, ConsistencyLe
         final ConsistencyLevel consistencyForCommit = consistencyLevel == ConsistencyLevel.LOCAL_SERIAL
                                                                         ? ConsistencyLevel.LOCAL_QUORUM
                                                                         : ConsistencyLevel.QUORUM;
-        final ConsistencyLevel consistencyForFetch = maybeCoerceReadConsistencyLevel(consistencyForCommit);
-        ClientRequestMetrics readMetrics = consistencyLevelReadMetrics.get(consistencyForFetch);
+        ReadCommand command = commands.get(0);
+        consistencyLevel = maybeCoerceReadConsistencyLevel(command.ksName, consistencyForCommit);
```
inespot, line 1303: Do we have the guarantee that all read commands in the list use the same keyspace?

inespot: If not, and we wanna keep it simple/safe, you probably wanna default to the highest consistency levels for all read commands, ie do a "if any keyspace is coerced"

andybradshaw: This code path in particular is a bit weird, where we'll throw if we actually have [more than one command](https://github.com/palantir/cassandra/pull/679/files#diff-4f6615c703489b87cafd23c0dc5a4edc4af27c33aba172870b30e38a7d8faa1dR1309). Agreed that it feels a bit dangerous, so could drop a comment that this is only safe because of that check/if that ever changed, we'd need to update this as well.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
andybradshaw: Yep, makes sense

src/java/org/apache/cassandra/tools/nodetool/GetCoerceReadConsistencyAllKeyspaces.java
andybradshaw: Added, seems like a pretty minor change, so probably worth it.

src/java/org/apache/cassandra/tools/nodetool/GetCoerceReadConsistencyAllKeyspaces.java
```diff
@@ -0,0 +1,42 @@
...
+    private String keyspace;
+
+    @Override
+    public void execute(NodeProbe probe)
+    {
+        if (keyspace.isEmpty()) {
+            probe.output().out.println(probe.getCoerceReadConsistencyAllKeyspaces());
+            return;
+        }
+        probe.output().out.println(probe.getCoerceReadConsistencyAllKeyspaces().contains(keyspace));
```
inespot: or `getCoerceReadConsistencyAllKeyspaces().isEmpty()` based on the recent modification to the code

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2102,11 +2103,26 @@ public static boolean getCoerceReadConsistencyAll()
         return conf.coerce_read_consistency_all;
     }
 
+    public static Set<String> getCoerceReadConsistencyAllKeyspaces()
+    {
+        return conf.coerce_read_consistency_all_keyspaces;
+    }
```
inespot, line 2109: Nit: to simplify consuming logic, we could have this function return all keyspaces if the set is empty. Not strongly opinionated though considering this is just for a rc, so feel free to leave as is if you prefer

src/java/org/apache/cassandra/tools/nodetool/GetCoerceReadConsistencyAllKeyspaces.java
inespot: Or simpler, use `getCoerceReadConsistencyAllForKeyspace(keyspace)`

src/java/org/apache/cassandra/tools/nodetool/GetCoerceReadConsistencyAllKeyspaces.java
```diff
@@ -0,0 +1,42 @@
...
+{
+
+    @Option(name = {"-k", "--keyspace"}, description = "The keyspace to check for read consistency level coercion")
+    private String keyspace;
+
+    @Override
+    public void execute(NodeProbe probe)
+    {
+        if (keyspace.isEmpty()) {
+            probe.output().out.println(probe.getCoerceReadConsistencyAllKeyspaces());
```
inespot: Another nit: if this is empty you probably wanna print "All keyspaces" or sth along those lines

src/java/org/apache/cassandra/tools/nodetool/GetCoerceReadConsistencyAllKeyspaces.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.apache.cassandra.tools.nodetool;
+
+import io.airlift.command.Command;
+import io.airlift.command.Option;
+import org.apache.cassandra.tools.NodeProbe;
+import org.apache.cassandra.tools.NodeTool;
+
+@Command(name = "getcoercereadconsistencyallkeyspaces", description = "Get all the keyspace coercion read consistency levels, or a single keyspace level if using the 'keyspace' option")
```
inespot: Nitty nit: "Get list of keyspaces for which read consistency level is coerced to ALL, or if, a keyspace is provided, whether that keyspace is coerced or not"

PR #680, opened 2025-05-02
Add logging for invalid thrift requests

inespot:
So we don't have to cut rcs to figure this out. 

Not exactly sure why the message does not show up on the client side atm since the thrift conversion should include it. 
But this logging would also provide us the affected keyspace + the stacktrace + the thrift method so going for this speedy solution instead of a bigger refactor of the thrown thrift error.
With traceIds, this should automatically show up as part of the client stack trace. Setting to debug to avoid noise.


andybradshaw (approved):
lgtm!

PR #684, opened 2025-05-09
Add Large Compaction Metrics

Iacobacci:
Add metrics and improve logging for large compactions (compactions expected to be over 5GiB in size).

- Metrics: A gauge that increments at the start of a large compaction task, and decrements at the end of that task.
- Logs: Clearer logging at the start and end of a large compaction, with the expected value logged at the start of the compaction and the actual value logged at the end.

This will help diagnose performance issues potentially caused by large, ongoing compactions. Tested this change on IL.

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
@@ -134,7 +134,7 @@ public class CompactionTask extends AbstractCompactionTask
         if (expectedWriteSize > FIVE_GIBIBYTES_IN_BYTES)
         {
             cfs.metric.incrementOngoingLargeCompactionTasks();
-            logger.info("Started compaction for ks/cf {}/{} that is expected to exceed 5GiB with size of {}",
+            logger.info("Starting compaction for ks/cf {}/{} that is expected to exceed 5GiB with size of {}",
                     SafeArg.of("keyspace", cfs.keyspace.getName()),
                     SafeArg.of("columnFamily", cfs.name),
                     SafeArg.of("size", expectedWriteSize));
--- a/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
@@ -131,8 +131,9 @@ public class CompactionTask extends AbstractCompactionTask
 
         final long expectedWriteSize = checkAvailableDiskSpaceAndGetWriteSize(checkAvailableDiskSpaceFunction);
 
-        if (expectedWriteSize > FIVE_GIBIBYTES_IN_BYTES)
+        if (expectedWriteSize > 0)
         {
+            cfs.metric.ongoingLargeCompactionTasks.inc();
             logger.info("Compaction for ks/cf {}/{} exceeds 5GiB with total size of {}",
                     SafeArg.of("keyspace", cfs.keyspace.getName()),
                     SafeArg.of("columnFamily", cfs.name),
@@ -315,6 +316,10 @@ public class CompactionTask extends AbstractCompactionTask
             // update the metrics
             cfs.metric.compactionBytesWritten.inc(endsize);
             cfs.metric.compactionsCompleted.inc();
+            if (expectedWriteSize > 0)
+            {
+                cfs.metric.ongoingLargeCompactionTasks.dec();
+            }
             CompactionThroughputThrottler.instance.maybeRemoveThrottledCompaction(cfs.metadata);
         }
     }
--- a/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionTask.java
@@ -131,10 +131,10 @@ public class CompactionTask extends AbstractCompactionTask
 
         final long expectedWriteSize = checkAvailableDiskSpaceAndGetWriteSize(checkAvailableDiskSpaceFunction);
 
-        if (expectedWriteSize > 0)
+        if (expectedWriteSize > FIVE_GIBIBYTES_IN_BYTES)
         {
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -315,6 +316,14 @@ public boolean apply(SSTableReader sstable)
             // update the metrics
             cfs.metric.compactionBytesWritten.inc(endsize);
             cfs.metric.compactionsCompleted.inc();
+            if (expectedWriteSize > FIVE_GIBIBYTES_IN_BYTES)
+            {
+                cfs.metric.decrementOngoingLargeCompactionTasks();
```
fsxfreak, line 321: this decrement is quite far down - any reason to worry if a a large compaction gets interrupted before this decrement for some reason?

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
```diff
@@ -509,6 +512,23 @@ public Long getValue()
...
+                new Gauge<Integer>() {
+                    public Integer getValue() {
+                        return ongoingLargeCompactionTaskCount.get();
+                    }
+                },
+                new Gauge<Integer>() {
+                    public Integer getValue() {
+                        int total = 0;
+                        for (Metric cfGauge : allColumnFamilyMetrics.get("OngoingLargeCompactionTasks")) {
+                            total += ((Gauge<? extends Number>) cfGauge).getValue().intValue();
```
fsxfreak, line 526: what do you think about ignoring negative gauge values for this total? not sure if it'll happen, but it's logically possible

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -133,7 +133,8 @@ protected void runMayThrow() throws Exception
 
         if (expectedWriteSize > FIVE_GIBIBYTES_IN_BYTES)
         {
-            logger.info("Compaction for ks/cf {}/{} exceeds 5GiB with total size of {}",
+            cfs.metric.incrementOngoingLargeCompactionTasks();
+            logger.info("Starting compaction for ks/cf {}/{} that is expected to exceed 5GiB with size of {}",
                     SafeArg.of("keyspace", cfs.keyspace.getName()),
                     SafeArg.of("columnFamily", cfs.name),
                     SafeArg.of("size", expectedWriteSize));
```
fsxfreak: nit: I like to include the unit in the name, something like `sizeBytes`

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
```diff
@@ -509,6 +512,23 @@ public Long getValue()
                 return max;
             }
         });
+        ongoingLargeCompactionTasks = createColumnFamilyGauge(
+                "OngoingLargeCompactionTasks",
+                new Gauge<Integer>() {
+                    public Integer getValue() {
+                        return ongoingLargeCompactionTaskCount.get();
```
fsxfreak, line 519: how often does this `get` occur? incurring volatile reads on every metric update seems expensive. but I don't have good intuition on whether this will result in an actual perf penalty

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
Iacobacci: We'd need the decrement to be after the compaction completes. There is the concern that a compaction could be interrupted, but it would still be good signal if the counter is always high after that compaction is interrupted. The counter would be cleared out after the node rolls anyway (upgrades, periodic rolls, etc)

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
Iacobacci: I'm not sure there is a benefit to ignoring them -- if we do go negative, it might be good signal something went wrong with the compaction flow

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
Iacobacci: This would run based on the metric executor, which should be every 30 seconds, along with the refresh of the other metrics

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
Iacobacci: 👍

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
fsxfreak: makes sense, if this gauge is looking a bit fuzzy we at least have the individual gauges to dive deeper on

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
fsxfreak: just found that internally some other gauges use `AtomicLongFieldUpdater` which seems like a small optimization. I'll leave it up to you on whether this would be worth implementing though, every 30 secs doesn't seem too bad

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
fsxfreak: makes sense. will be interesting to see how fuzzy this gauge is.

fsxfreak (approved):
stamped with one question

src/java/org/apache/cassandra/metrics/ColumnFamilyMetrics.java
Iacobacci: Discussed IRL, going to keep the current implementation

PR #685, opened 2025-05-09
Remove two frequent allocations

rynorris:
Noticed these two suspicious methods accounting for a rather large portion of `byte[]` allocations in a JFR I took recently.
(note that in this image I'm selected to only `byte[]` allocations, which accounted for around 16% of all allocations recorded, so these methods are quite a large proportion of total allocations!)

image

  - **OwnershipVerificationUtils.java:** This one is silly because the log is only logged if the verification fails, but the argument is converted to a string pre-emptively. I've made it pass through the stringify method to use and only stringify when necessary.
  - **ColumnFamilyStore.java:** This one is in a trace log, so it can't be optimised in the same way as above. But we don't consume Cassandra trace logs anyway AFAIK, so just opting to remove this arg.

A JFR with a similar load profile after this fix:
image

### DIFF
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -2423,7 +2423,7 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean
         assert !(range.keyRange() instanceof Range) || !((Range<?>)range.keyRange()).isWrapAround() || range.keyRange().right.isMinimum() : range.keyRange();
 
         final ViewFragment view = select(viewFilter(range.keyRange()));
-        Tracing.trace("Executing seq scan across {} sstables for {}", view.sstables.size(), range.keyRange().getString(metadata.getKeyValidator()));
+        Tracing.trace("Executing seq scan across {} sstables", view.sstables.size());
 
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2423,7 +2423,7 @@ private AbstractScanIterator getSequentialIterator(final DataRange range, long n
         assert !(range.keyRange() instanceof Range) || !((Range<?>)range.keyRange()).isWrapAround() || range.keyRange().right.isMinimum() : range.keyRange();
 
         final ViewFragment view = select(viewFilter(range.keyRange()));
-        Tracing.trace("Executing seq scan across {} sstables for {}", view.sstables.size(), range.keyRange().getString(metadata.getKeyValidator()));
+        Tracing.trace("Executing seq scan across {} sstables", view.sstables.size());
```
rhuffy: could we instead wrap this call in `if(Tracing.isTracing())`?

rynorris: We certainly can, nice idea!

d-guo (approved):

PR #686, opened 2025-05-09
Remove PredictedSpeculativeRetryPerformanceMetrics

rynorris:
Loading these metric snapshots accounts for a large portion of all memory allocations on read codepaths (`StorageProxy.fetchRows()`). We call this method on **every read** request in order to compute a metric, and every `getSnapshot()` call performs memory allocations. This will be many thousands (or 10s, or 100s of thousands!) of times per second.

In my test trace, it accounts for 2.85% of all total allocs recorded. But my test workload is very write-heavy, with few reads.
image

Looking at a real JFR from a production environment, it's much, much higher (12% of all allocs!):
image

We are only creating these snapshots in order to compute these other metrics (which we added, I don't think these are in trunk at all). So I think it should be ok to cache these values for 1s.

JFR before:

image

JFR after (notice the above boxed codepaths are totally absent now):

image


nickbar01234:
Since we already disable speculative retry, I wonder if there's any benefits to keeping this metrics at all 🤔

andybradshaw:
I think given Nick's comment, we're maybe better off just removing this entirely, as I don't think caching the snapshot actually gives us what we expect for the values of the metrics themselves.

Can do a little bit more digging today about the intent of these metrics. I'm guessing this was to gather data before turning the feature on, and the metrics show that it wouldn't be meaningfully better.

rynorris:
Have updated the PR to remove these metrics entirely.

rynorris:
Closing in favor of #696

PR #687, opened 2025-05-09
Safe Logging for org.apache.cassandra.dht

d-guo:
Safe Logging for the `org.apache.cassandra.dht` package

### DIFF
--- a/src/java/org/apache/cassandra/dht/RangeStreamer.java
+++ b/src/java/org/apache/cassandra/dht/RangeStreamer.java
@@ -27,11 +27,11 @@ import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
+import com.palantir.logsafe.SafeArg;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.antlr.analysis.SemanticContext;
 import org.apache.cassandra.db.Keyspace;
 import org.apache.cassandra.db.SystemKeyspace;
 import org.apache.cassandra.gms.EndpointState;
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/dht/RangeStreamer.java
```diff
@@ -163,15 +163,19 @@ public void addRanges(String keyspaceName, Collection<Range<Token>> ranges)
         if (logger.isTraceEnabled())
         {
             for (Map.Entry<Range<Token>, InetAddress> entry : rangesForKeyspace.entries())
-                logger.trace(String.format("%s: range %s exists on %s", description, entry.getKey(), entry.getValue()));
+            {
+                logger.trace("{}: range {} exists on {}", SafeArg.of("description", description), SafeArg.of("range", entry.getKey()), SafeArg.of("endpoint", entry.getValue()));
```
d-guo, line 167: description is always either "Bootstrap" from `Bootstrapper` or "Rebuild" from `StorageService`

src/java/org/apache/cassandra/dht/RangeStreamer.java
```diff
@@ -395,7 +403,7 @@ Set<Range<Token>> removeAvailableRanges(Map.Entry<String, Map.Entry<InetAddress,
         entry.getValue().getValue().removeAll(availableRanges);
         if (!availableRanges.equals(unavailableRanges) && log)
         {
-            logger.info("Some ranges of {} are already available. Skipping streaming those ranges.", availableRanges);
+            logger.info("Some ranges of {} are already available. Skipping streaming those ranges.", SafeArg.of("availableRanges", availableRanges));
```
d-guo, line 406: i believe ranges/tokens in this class are safe to log because they are the ones owned by nodes and chosen to stream during rebuilds/bootstraps. this is different than tokens as keys from user data

rhuffy (approved):

PR #689, opened 2025-05-14
Add metrics for measuring commit log performance

rynorris:


src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
```diff
@@ -374,7 +374,7 @@ void waitForFinalSync()
 
     void waitForSync(int position, Timer waitingOnCommit)
     {
-        while (lastSyncedOffset < position)
+        try (Timer.Context ignored2 = commitLog.metrics.totalWaitingOnCommit.time())
         {
             WaitQueue.Signal signal = waitingOnCommit != null ?
```
andybradshaw, line 379: question: AFAICT, this is never actually null? Guessing this has to do with the fact that it gets called from `static class Allocation`...

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
```diff
@@ -374,7 +374,7 @@ void waitForFinalSync()
 
     void waitForSync(int position, Timer waitingOnCommit)
     {
-        while (lastSyncedOffset < position)
+        try (Timer.Context ignored2 = commitLog.metrics.totalWaitingOnCommit.time())
```
andybradshaw, line 377: super nit: why is this `ignored2`? shouldn't conflict with anything in this scope, could remove the `2` suffix

andybradshaw (approved):
Nothing blocking, one random java question, otherwise LGTM

PR #693, opened 2025-05-15
Add an optimized version of WaitQueue specifically for commit log syncs

rynorris:
The existing `WaitQueue` doesn't know anything about which waiters are actually "ready" to wake up on each loop of the commit log sync, so it just tries to wake them all up.

The waiting threads then rejoin the queue if their data isn't synced yet.

This creates a death loop whereby if there are enough waiting threads, they can be rejoining the back of the queue just as fast as we're popping them off and waking them, meaning the queue can't make progress.

The existing implementation papers over this by tracking a random item in the queue and bailing out if we see it loop around. But this is not very effective.

In practice, we've seen queues containing only ~100 waiting threads pop and wake over **500** threads before bailing. This blocks the commit log from continuing to the next sync iteration. We've seen it block for several milliseconds, which materially impacts write performance.

This new implementation (hidden behind a config flag) tracks the waiters in a skip list, and only wakes exactly the threads which are ready to be woken.

See results from a test environment, left is with the skip list, right is the default behaviour on trunk.
image

### DIFF
--- a/src/java/org/apache/cassandra/utils/concurrent/SkipListProgressWaitQueue.java
+++ b/src/java/org/apache/cassandra/utils/concurrent/SkipListProgressWaitQueue.java
@@ -58,7 +58,11 @@ public final class SkipListProgressWaitQueue implements ProgressWaitQueue
 
     public void signalUntil(long until) {
         ConcurrentNavigableMap<Long, RegisteredSignal> dueWaiters = waiters.headMap(until, true);
-        try (CloseableTracer ignored = CloseableTracer.startSpan("ProgressWaitQueue#signalAll", ImmutableMap.of("numWaiters", Integer.toString(waiters.size()), "numDueWaiters", Integer.toString(dueWaiters.size()))))
+        try (CloseableTracer ignored = CloseableTracer.startSpan(
+            "SkipListProgressWaitQueue#signalAll",
+                ImmutableMap.of(
+                    "numWaiters", Integer.toString(waiters.size()),
+                    "numDueWaiters", Integer.toString(dueWaiters.size()))))
         {
             Iterator<RegisteredSignal> iter = dueWaiters.values().iterator();
             while (iter.hasNext())
@@ -80,7 +84,7 @@ public final class SkipListProgressWaitQueue implements ProgressWaitQueue
      */
     private class RegisteredSignal extends WaitQueue.AbstractSignal
     {
-        private final DetachedSpan span = DetachedSpan.start("ProgressWaitQueue#parked");
+        private final DetachedSpan span = DetachedSpan.start("SkipListProgressWaitQueue#parked");
         private volatile Thread thread = Thread.currentThread();
         volatile int state;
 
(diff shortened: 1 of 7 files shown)


src/java/org/apache/cassandra/utils/concurrent/SkipListProgressWaitQueue.java
```diff
@@ -0,0 +1,183 @@
...
+                    "numWaiters", Integer.toString(waiters.size()),
+                    "numDueWaiters", Integer.toString(dueWaiters.size()))))
+        {
+            Iterator<RegisteredSignal> iter = dueWaiters.values().iterator();
+            while (iter.hasNext())
+            {
+                RegisteredSignal signal = iter.next();
+                signal.signal();
+                iter.remove();
+            }
```
andybradshaw, line 78: Is this roughly equivalent to the below?
```suggestion
            dueWaiters.values().iterator().forEachRemaining(RegisteredSignal::signal);
```

src/java/org/apache/cassandra/utils/concurrent/SkipListProgressWaitQueue.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private void cleanUpCancelled()
+    {
+        waiters.values().removeIf(RegisteredSignal::isCancelled);
+    }
+
+    /**
+     * A signal registered with this WaitQueue
+     */
+    private class RegisteredSignal extends WaitQueue.AbstractSignal
```
andybradshaw, line 90: A couple of thoughts/questions here:
- should we be closing these spans in the non-signaling case? and should we include that metadata in the span?
- it looks like the only differences between this and the `WaitQueue.RegisteredSignal` class is the tracing... should we just combine those? (looks like there may be a visibility issue with `signal()`, which could maybe be package private?)

andybradshaw: for the second bullet, sync'd offline... idea is that we should prefer small copies here out of an abundance of caution of any changes to shared things having unintended consequences

src/java/org/apache/cassandra/utils/concurrent/SkipListProgressWaitQueue.java
rynorris: I think the key thing is that you're missing the `iter.remove()` call in that version, which means you're not dropping them from the list.

src/java/org/apache/cassandra/utils/concurrent/SkipListProgressWaitQueue.java
rynorris: Handled the first point. 👍

src/java/org/apache/cassandra/utils/concurrent/SkipListProgressWaitQueue.java
andybradshaw: ahh... misread the bit about that being from the underlying collection.

andybradshaw (approved):

PR #695, opened 2025-05-15
Backport CASSANDRA-14660

nickbar01234:
Backport [CASSANDRA-14660](https://issues.apache.org/jira/browse/CASSANDRA-14660). Even though our #nodes is much smaller than the original ticket and we run a small number of vnodes, this change would still be beneficial given the rate at which we restart nodes.


inespot:
Could you add a quick set of unit tests for the `create` function for sanity checking?

inespot (approved):

PR #696, opened 2025-05-16
Remove speculative retry metrics

andybradshaw:
These metrics were added as part of an experiment to make single node disruptions less impactful to the cluster overall, but we ultimately never turned speculative retry on after seeing negligible improvements to performance. As outlined in more detail in https://github.com/palantir/cassandra/pull/686, these are actually fairly expensive to compute, and should just be removed.


nickbar01234:
Not super important, but curious if there's any reasons we wouldn't want to revert the [original PR](https://github.com/palantir/cassandra/commit/e301a8dfa26197f9585e11011b9a3649e5858f99#diff-1e7afcc93a8b993ec07c6256f8e0aaff89ac366d9d24a4c4614df9f13d8989b2L324-L325) that introduces the metrics too. From a quick skim, it looks like

* Changes to `AbstractReadExecutor`
* `IEndpointSnitch#getSnapshot`

Edit: Thinking about it a bit more, I think we should remove at the very least `IEndpointSnitch#getSnapshot` to avoid exposing an expensive API that could be a footgun.

nickbar01234:
small nit: Can probably remove these stale code / comment

1. [Annotation](https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/service/AbstractReadExecutor.java#L87-L90)
2. [Comment](https://github.com/palantir/cassandra/blob/palantir-cassandra-2.2.18/src/java/org/apache/cassandra/service/AbstractReadExecutor.java#L120)

nickbar01234 (approved):
For RC

nickbar01234:
Confirmed via JFR and metrics, looks good!

PR #697, opened 2025-05-20
Wrap AbstractLocalAwareExecutorService tasks with tracing information

andybradshaw:
Previously, tasks submitted to the `AbstractLocalAwareExecutorService` would lose tracing information, as tracing information is stored in thread local data and the task would be executed on a different thread. This PR wraps the lower-most `Callable` (all task-producing code paths end in this constructor) with a deferred tracer, allowing the task to be associated with the correct trace ID and parent span.

### DIFF
--- a/src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
+++ b/src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
@@ -30,8 +30,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.palantir.tracing.Tracers;
-import org.apache.cassandra.tracing.TraceState;
-import org.apache.cassandra.tracing.Tracing;
 import org.apache.cassandra.utils.concurrent.SimpleCondition;
 import org.apache.cassandra.utils.JVMStabilityInspector;
 
--- a/src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
+++ b/src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
@@ -29,6 +29,7 @@ import java.util.concurrent.TimeoutException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.palantir.tracing.Tracers;
 import org.apache.cassandra.tracing.TraceState;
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
```diff
@@ -150,7 +149,7 @@ class FutureTask<T> extends SimpleCondition implements Future<T>, Runnable
 
         public FutureTask(Callable<T> callable)
```
inespot, line 150: Seems like those get created for a wide variety of tasks in Cassandra. Did we get a chance to verify from a rc the log volume + perf from adding this additional wrapper? 
The Tracers metadata will be used for all traceIds coming from Alta correct? Or am I mixing up?

andybradshaw: So for any tasks generated that's not part of a thrift request, a random traceId will be used, which will be subjected to the configured trace rate (intentionally very low for us). Additionally, we have the ability to downsample forced client requests, so we can continue to reduce our trace volume if we need to. 

This on it's own won't add too many spans (just the one for the task executing), and in the cases where we're not sampling a trace (~99% of the time), the performance hit should be negligible (an `Unsampled` span has very cheap implementations of all trace operations).

andybradshaw:
An example of how the `FutureTask#run` spans end up getting emitted/rendered
Screenshot 2025-05-23 at 1 11 42 PM

src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
```diff
@@ -150,7 +149,7 @@ class FutureTask<T> extends SimpleCondition implements Future<T>, Runnable
...
         {
-            this.callable = callable;
+            this.callable = Tracers.wrap("FutureTask#run", callable);
```
inespot, line 152: For my own future ref: https://github.com/palantir/tracing-java/blob/4.23.0/tracing/src/main/java/com/palantir/tracing/Tracers.java#L179

inespot (approved):
From offline discussion:
When upgrading in sls-cassandra, drop the default downsampling mode to `FIRST_THIRTY_SECOND`, as we can override this to sample more on the stacks that we're actively evaluating

PR #698, opened 2025-05-21
Support propagating tracing metadata between nodes via message metadata

andybradshaw:
Previously, tracing metadata would only be recorded on the node receiving the initial thrift request. With this PR, if tracing information is present, it will be included in the `MessageOut` metadata and interpreted/re-injected when processing a `MessageIn`. This allows for cross-node tracing of requests, which will render like the following:

Screenshot 2025-05-23 at 2 33 04 PM


src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
```diff
@@ -150,7 +149,7 @@ class FutureTask<T> extends SimpleCondition implements Future<T>, Runnable
 
         public FutureTask(Callable<T> callable)
         {
-            this.callable = callable;
+            this.callable = Tracers.wrap("FutureTask#run", callable);
```
d-guo: i know this is just for testing but what does `Tracers.wrap` do? is this starting a span?

src/java/org/apache/cassandra/net/MessageOut.java
```diff
@@ -85,7 +86,14 @@ public MessageOut<T> withParameter(String key, byte[] value)
     {
         ImmutableMap.Builder<String, byte[]> builder = ImmutableMap.builder();
         builder.putAll(parameters).put(key, value);
-        return new MessageOut<T>(verb, payload, serializer, builder.build());
+        return new MessageOut<T>(verb, payload, serializer, builder.buildKeepingLast());
+    }
+
+    public MessageOut<T> withParameters(Map<String, byte[]> values)
```
d-guo: nit: ~~values~~ -> params

src/java/org/apache/cassandra/tracing/Tracing.java
```diff
@@ -240,6 +242,16 @@ public TraceState initializeFromMessage(final MessageIn<?> message)
         }
     }
 
+    public static Map<String, byte[]> serializeForMessage() {
+        if (isTracing()) {
+            return ImmutableMap.of(
+            TRACE_HEADER, UUIDGen.decompose(Tracing.instance.getSessionId()),
+            TRACE_TYPE, new byte[] { Tracing.TraceType.serialize(Tracing.instance.getTraceType()) });
```
d-guo: style nit: don't need the `Tracing` part of `Tracing.instance` anymore

src/java/org/apache/cassandra/tracing/PalantirTracing.java
```diff
@@ -49,6 +61,31 @@ public static void initializeTracerFromIncomingThriftMessage(String thriftOperat
...
+        }
+    }
+
+    public static Map<String, byte[]> serializeForMessage() {
+        Optional<com.palantir.tracing.TraceMetadata> traceMetadata = Tracer.maybeGetTraceMetadata();
+        if (traceMetadata.isPresent()) {
+            return ImmutableMap.of(
+            PALANTIR_TRACE_ID, traceMetadata.get().getTraceId().getBytes(StandardCharsets.UTF_8),
+            PALANTIR_PARENT_SPAN_ID, traceMetadata.get().getSpanId().getBytes(StandardCharsets.UTF_8),
+            PALANTIR_IS_SAMPLED, new byte[] { (byte) (Tracer.isTraceObservable() ? 1 : 0) }
```
d-guo, line 82: just making sure: the tracing downsampling added internally happens at a layer above, so even if this "isSampled" is true, the downsampling will still occur for inter-node messages

d-guo: i guess it "depends" on if the node receiving the message has the same downsampling config as the sending node. if the sender has 1/4 but the receiver has 1/2, it's possible for the receiver to emit traces for a request that the sender doesn't. is that correct?

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -678,8 +682,14 @@ public int sendRRWithFailure(MessageOut message, InetAddress to, IAsyncCallbackW
      */
     public int sendRR(MessageOut message, InetAddress to, IAsyncCallback cb, long timeout, boolean failureCallback)
     {
+        // n.b. this will be closed on the traced callback
```
d-guo, line 687: how does this work for `sendOneWay`? does the receiver node still receive the metadata to emit traces?

d-guo:
can we also add in the description what types of inter-node messages we're adding tracing for? is it all of them? also can you attach the results of your testing here when done

src/java/org/apache/cassandra/tracing/PalantirTracing.java
andybradshaw: Correct on both fronts, we both ignore span emission for trace IDs that get downsampled, and if downsampling is different between the services (sender more restrictive than reciever), we may end up with some odd span reporting.

src/java/org/apache/cassandra/concurrent/AbstractLocalAwareExecutorService.java
andybradshaw: `Tracers.wrap` doesn't actually start a span, it creates a `DeferredTracer`, which stores off all the state for the current context for spans to be started with later.

src/java/org/apache/cassandra/net/MessagingService.java
andybradshaw: Added an explicit span for `sendOneWay`, but yes, the receiver node will still interpret the tracing information on the message itself (the parent `spanId` will just be whoever created the message out, not the actual `send` 🙁)

src/java/org/apache/cassandra/net/TracedCallback.java
```diff
@@ -0,0 +1,65 @@
...
+import java.net.InetAddress;
+
+import com.google.common.collect.ImmutableMap;
+
+import com.palantir.tracing.CloseableSpan;
+import com.palantir.tracing.DetachedSpan;
+
+public class TracedCallback<T> implements IAsyncCallbackWithFailure<T>
+{
+    private final MessageOut message;
```
d-guo: `TracedCallback` doesn't need the entire message, right? the message id or verb should be enough?

src/java/org/apache/cassandra/net/TracedCallback.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public boolean isLatencyForSnitch()
+    {
+        return delegate.isLatencyForSnitch();
+    }
+
+    @Override
+    public void onFailure(InetAddress from)
+    {
+        // Trust that the caller has checked this for us.
```
d-guo, line 59: wow the callback failures api is terrible
```
public int addCallback(IAsyncCallback cb, MessageOut message, InetAddress to, long timeout, boolean failureCallback)
```
whether or not `cb` has a failure callback is separately determined by `failureCallback` ??

this change in the PR is fine from my read of the code but this whole thing is silly

d-guo (dismissed):
just the comment about `TracedCallback` containing the full message. otherwise PR LGTM

d-guo (approved):

PR #704, opened 2025-05-27
Log when KeyspaceTableOpStatePersister fails to persist and test CleanupStateTracker losing cache

fsxfreak:
This PR contains three items:
- additional logging of the exception whenever KeyspaceTableOpStatePersister fails to persist
- fix a bug where we attempt to delete the tmp file which has already successfully atomically moved to the actual json persisted file (and thus does not exist anymore)
- new unit test to test the behavior of what happens when CleanupStateTracker does not persist a cleanup time, loses the cached state, then tries to fetch the last cleanup timestamp. This confirms that the tracker knows that it is missing a cleanup timestamp entry for a valid table, and falls back to the min timestamp

### DIFF
--- a/src/java/org/apache/cassandra/service/opstate/CleanupStateTracker.java
+++ b/src/java/org/apache/cassandra/service/opstate/CleanupStateTracker.java
@@ -99,7 +99,10 @@ public class CleanupStateTracker
         Map<KeyspaceTableKey, Instant> updatedEntries = state.updateTsForEntry(key, value);
         updateCacheIfHasNotYetSuccessfullyReadFromPersister();
         if (!successfulReadFromPersister || !persister.updateStateInPersistentLocation(updatedEntries))
-            log.warn("Failed to update persistant cleanup state, but cache has been updated. Will retry at next update.", SafeArg.of("key", key));
+            log.warn("Failed to update persistant cleanup state, but cache has been updated. Will retry at next update.",
+                     SafeArg.of("keyspace", key.getKeyspace()),
+                     SafeArg.of("cf", key.getColumnFamily())
+            );
     }
 
     private void updateCacheIfHasNotYetSuccessfullyReadFromPersister()
--- a/src/java/org/apache/cassandra/service/opstate/CleanupStateTracker.java
+++ b/src/java/org/apache/cassandra/service/opstate/CleanupStateTracker.java
@@ -30,6 +30,7 @@ import com.google.common.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.palantir.logsafe.SafeArg;
 import org.apache.cassandra.config.DatabaseDescriptor;
 
 public class CleanupStateTracker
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/service/opstate/CleanupStateTracker.java
```diff
@@ -98,7 +99,7 @@ void updateTsForEntry(KeyspaceTableKey key, Instant value) throws IllegalArgumen
         Map<KeyspaceTableKey, Instant> updatedEntries = state.updateTsForEntry(key, value);
         updateCacheIfHasNotYetSuccessfullyReadFromPersister();
         if (!successfulReadFromPersister || !persister.updateStateInPersistentLocation(updatedEntries))
-            log.warn("Failed to update persistant cleanup state, but cache has been updated. Will retry at next update.");
+            log.warn("Failed to update persistant cleanup state, but cache has been updated. Will retry at next update.", SafeArg.of("key", key));
```
inespot: Nit: 
```
log.warn("Failed to update persistant cleanup state, but cache has been updated. Will retry at next update.", 
                     SafeArg.of("keyspace", key.getKeyspace()),
                     SafeArg.of("cf", key.getColumnFamily()));
```
to clarify the logs + that we are not safe-arging an actual key

src/java/org/apache/cassandra/service/opstate/KeyspaceTableOpStatePersister.java
```diff
@@ -124,6 +124,7 @@ private boolean writeStateToFile(File file, Map<KeyspaceTableKey, Instant> updat
         }
         catch (IOException e)
         {
+            log.warn("Failed to write state to file.", file.getAbsolutePath(), e);
```
inespot: In this log and other instances of it, can we switch to:
```
log.warn("Failed to write state to file.", SafeArg.of("file", file.getName()), e);
```
to be able to get them in Loki

test/unit/org/apache/cassandra/service/opstate/CleanupStateTrackerTest.java
```diff
@@ -133,4 +139,64 @@ public void getLastSuccessfulCleanupTsForNodeReturnsMinTsIfNoEntriesExist()
...
+        assertThat(tracker.getLastSuccessfulCleanupTsForNode()).isEqualTo(instant1);
+
+        Instant instant2 = instant1.plusSeconds(30);
+        Instant instant3 = instant1.plusSeconds(35);
+        doReturn(false).when(persister).updateStateInPersistentLocation(
+            argThat(map -> map.containsKey(OpStateTestConstants.KEYSPACE_TABLE_KEY_2) && map.get(OpStateTestConstants.KEYSPACE_TABLE_KEY_2).equals(instant3)
+        ));
+        try (MockedStatic<Instant> mockInstant = mockStatic(Instant.class)) {
+            mockInstant.when(Instant::now).thenReturn(instant2);
+            tracker.recordSuccessfulCleanupForTable(OpStateTestConstants.KEYSPACE1, OpStateTestConstants.TABLE1);
```
inespot: Instead of mockingStatic could you use `updateTsForEntry` ?

fsxfreak: somehow missed that, that works much better

inespot (approved):

PR #705, opened 2025-05-28
Do not purge tombstones if keyspace is repairing

inespot:
Derived from: https://github.com/palantir/cassandra/pull/634

**Corruption case**

- `keyspace1/cf1` is repairing and a neighboring node starts streaming the list of file snapshotted at time T to this node. Those sstables contain `key1` with `col1:value1`. 
- At T+1 a tombstone is written for `col1:val1`, deleting the column. 
- At T+1+gc_grace, the sstable on the node containing `col1:value1` + the tombstone is compacted, and the two cells (value + tombstone) are purged. 
- At T+2+gc_grace, the neighboring node finally streams the file containing `col1:value1`. The data is then resurrected.

**Fix and justification**

The `RepairRunnable` class creates a `ParentRepairSession` which gets tracked by the `ActiveRepairService`. This happens both on the coordinator's end: [code ref](https://github.com/palantir/cassandra/blob/1.213.0/src/java/org/apache/cassandra/service/ActiveRepairService.java#L282) and neighboring node's: [code ref ](https://github.com/palantir/cassandra/blob/1.213.0/src/java/org/apache/cassandra/repair/RepairMessageVerbHandler.java#L91)
Those sessions get untracked after the repair sessions (and therefore the streams) finish, successfully or not: https://github.com/palantir/cassandra/pull/634#discussion_r1988430643

If a compaction gets triggered before the repair session starts and continues while the repair session is happening, the LifecycleTransaction associated with the transaction should not consider the tombstone in the example presented above

**Perf**

- Cassandra 2.2 does not automatically run full column family repairs. They have to be user triggered (which we usually only in disaster recovery circumstances). Because of this the `parentRepairSessions` map should be mostly empty.
- The purgeTombstone repair exception is targeting a single cf at a time. Considering we also already run bootstraps without purging tombstones, this is not expected to significantly affect the overall repair perf in disaster recovery scenarios

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/CompactionController.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionController.java
@@ -34,6 +34,7 @@ import org.apache.cassandra.db.ColumnFamilyStore;
 import org.apache.cassandra.db.DecoratedKey;
 import org.apache.cassandra.db.Memtable;
 import org.apache.cassandra.db.RowPosition;
+import org.apache.cassandra.service.ActiveRepairService;
 import org.apache.cassandra.service.StorageService;
 import org.apache.cassandra.utils.AlwaysPresentFilter;
 
@@ -79,7 +80,7 @@ public class CompactionController implements AutoCloseable
 
     void maybeRefreshOverlaps()
     {
-        if (doNotPurgeTombstones(cfs.keyspace.getName()))
+        if (doNotPurgeTombstones(cfs))
         {
             logger.debug("not refreshing overlaps - doNotPurgeTombstones returned true for keyspace {}",
                          SafeArg.of("keyspace", cfs.keyspace.getName()));
@@ -98,7 +99,7 @@ public class CompactionController implements AutoCloseable
 
     private void refreshOverlaps()
     {
-        if (doNotPurgeTombstones(cfs.keyspace.getName()))
+        if (doNotPurgeTombstones(cfs))
         {
             logger.debug("not refreshing overlaps - doNotPurgeTombstones returned true for keyspace {}",
                          SafeArg.of("keyspace", cfs.keyspace.getName()));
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/db/compaction/CompactionController.java
```diff
@@ -145,9 +147,9 @@ public static Set<SSTableReader> getFullyExpiredSSTables(ColumnFamilyStore cfSto
         if (compacting == null)
             return Collections.<SSTableReader>emptySet();
 
-        if (doNotPurgeTombstones(cfStore.keyspace.getName())) {
-            logger.debug("not looking for droppable sstables - doNotPurgeTombstones returned true for keyspace {}",
-                         SafeArg.of("keyspace", cfStore.keyspace.getName()));
+        if (doNotPurgeTombstones(cfStore))
+        {
+            logger.debug("Not looking for droppable sstables - doNotPurgeTombstones returned true for keyspace {}", cfStore.keyspace.getName());
```
rhuffy: why remove the safearg?

rhuffy (dismissed):
approved with nit

src/java/org/apache/cassandra/db/compaction/CompactionController.java
inespot: Mistake on the rebase deconflicting, nice catch!

src/java/org/apache/cassandra/db/compaction/CompactionController.java
```diff
@@ -19,6 +19,7 @@
 
 import java.util.*;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
```
rhuffy: could you also remove the unused imports?

rhuffy (approved):

inespot:
👍

PR #706, opened 2025-05-28
Use `assertWithError` instead of `assert` for LRRS code paths

d-guo:
Depending on the environment, cassandra may be run with assertions disabled in consideration of the perf hit to have them enabled. The assertions added for resumable range scans are likely to be desirable in all cases, so in this PR we throw an exception instead.


d-guo:
open to renaming `assertWithError` to something like `forceAssert` or something like that

rhuffy (approved):

PR #710, opened 2025-05-30
Feature: Instrument iterator equality for filter scans

lucaskjaero:
Most filter experiment failures seem to be by iterator. This most likely means that one of the returned cells was not `left.equals(right)` by way of however that `Cell` subclass (of which there are 15) implements equality.

I've added a new comparison method that collects and stores what is safe to log about any differences. This should work out of the box to explain most Cell types, but I will follow up with any other interesting parameters for the other types.

### DIFF
--- a/src/java/org/apache/cassandra/FilterExperiment.java
+++ b/src/java/org/apache/cassandra/FilterExperiment.java
@@ -18,18 +18,13 @@
 
 package org.apache.cassandra;
 
-import java.util.ArrayList;
 import java.util.Iterator;
-import java.util.List;
-import java.util.Optional;
 import java.util.concurrent.ThreadLocalRandom;
-import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Predicates;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.slf4j.Logger;
@@ -143,7 +138,7 @@ public enum FilterExperiment
             boolean areEqual = !iterator(legacy).hasNext();
             return areEqual ? ComparisonResult.EQUAL : ComparisonResult.MODERN_WAS_NULL;
         } else {
-            return compareByIterator(iterator(legacy), iterator(modern));
+            return compareByIterator(iterator(legacy), iterator(modern), legacy.metadata());
         }
     }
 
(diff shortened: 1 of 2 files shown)


test/unit/org/apache/cassandra/FilterExperimentTest.java
```diff
@@ -50,6 +50,13 @@ public void testAreEqual_trueIfEqual() {
         assertThat(FilterExperiment.areEqual(cf, cf).isEqual()).isTrue();
     }
 
+    @Test
+    public void testAreEqual_falseIfNotEqual() {
```
lucaskjaero, line 54: I partially created this to make it easy to see my generated log.

src/java/org/apache/cassandra/FilterExperiment.java
```diff
@@ -147,10 +150,67 @@ private static Iterator<Cell> iterator(ColumnFamily columnFamily) {
...
+        return iterator.hasNext() ? Optional.ofNullable(iterator.next()) : Optional.empty();
+    }
+
     static boolean areTrulyEqual(ColumnFamily legacy, ColumnFamily modern) {
         return ColumnFamily.digest(legacy).equals(ColumnFamily.digest(modern));
     }
 
+    static ComparisonResult compareByIterator(Iterator<Cell> legacyIterator, Iterator<Cell> modernIterator)
+    {
+        ImmutableList.Builder<String> differencesBuilder = ImmutableList.builder();
```
schlosna: Should this just use ArrayList to avoid extra array copies?

src/java/org/apache/cassandra/FilterExperiment.java
```diff
@@ -147,10 +150,67 @@ private static Iterator<Cell> iterator(ColumnFamily columnFamily) {
                                 Predicates.not(columnFamily.inOrderDeletionTester()::isDeleted));
     }
 
+    private static Optional<Cell> getNext(Iterator<Cell> iterator)
+    {
+        return iterator.hasNext() ? Optional.ofNullable(iterator.next()) : Optional.empty();
```
schlosna: If we're mapping null cells to empty, can we avoid optional allocation overhead and just return `@Nullable`?

src/java/org/apache/cassandra/FilterExperiment.java
```diff
@@ -147,10 +150,67 @@ private static Iterator<Cell> iterator(ColumnFamily columnFamily) {
...
+        if (differences.isEmpty()) {
+            return ComparisonResult.EQUAL;
+        } else {
+            log.warn("Column families returned different results via iterator: {}", SafeArg.of("differences", differences));
+            return ComparisonResult.NOT_EQUAL_BY_ITERATOR;
+        }
+    }
+
+    private static String missingItem(int index, boolean hasLegacyCell, boolean hasModernCell) {
+        return String.format("[%s: MissingItem(hasLegacyCell=%s, hasModernCell=%s)]",
```
schlosna: String.format will be expensive to parse and allocate on potentially hot comparison paths

lucaskjaero: This should not be a hot comparison path, we see maybe 5-10 failures per month across the fleet (excluding RCs under test). Since we only run the String.format on failures we should be good? Given this is so hard to reproduce I think we should prioritize getting as much information as possible.

Definitely open to cheaper ways of logging though.

andybradshaw (approved):
Will keep an eye on the metrics for the total volume of these... As @lucaskjaero said, we really don't expect to be hitting this frequently, so the cost of the happy path should be cheap enough that this isn't a problem.

PR #713, opened 2025-06-06
Allow Schema Announcements and Pulls Between Compatible Messaging Versions

d-guo:
Allow Schema Announcements and Pulls Between Compatible Messaging Versions

### DIFF
--- a/src/java/org/apache/cassandra/service/MigrationManager.java
+++ b/src/java/org/apache/cassandra/service/MigrationManager.java
@@ -231,10 +231,10 @@ public class MigrationManager
     public static boolean shouldPullSchemaFrom(InetAddress endpoint)
     {
         /*
-         * Don't request schema from nodes with an incompatible or unknown major version (may have incompatible schema)
+         * Don't request schema from nodes with an incompatible or unknown major version
          * Don't request schema from fat clients
          */
-        return compatibleVersionsForSchemaPull(endpoint) && !Gossiper.instance.isGossipOnlyMember(endpoint);
+        return compatibleMessagingVersionsForSchema(endpoint) && !Gossiper.instance.isGossipOnlyMember(endpoint);
     }
 
     public static boolean shouldPullSchemaFrom(InetAddress endpoint, UUID theirVersion)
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/service/MigrationManager.java
```diff
@@ -70,6 +70,9 @@ public class MigrationManager
     public static final int MIGRATION_DELAY_IN_MS = 60000;
     public static final int MAX_SCHEDULED_SCHEMA_PULL_REQUESTS = 3;
 
+    // These messaging service versions are known to have compatible schema formats so nodes on these versions are allowed to pull from each other
+    private static final Set<Integer> SCHEMA_COMPATIBLE_VERSIONS_RRS = new HashSet<>(Arrays.asList(MessagingService.VERSION_22, MessagingService.VERSION_22_PLTR));
```
rhuffy: Why not an ImmutableSet here?

d-guo: ah good point didn't realize that was available in java 8

rhuffy:
Looks good. Can you add some tests in `MigrationManagerTest`?

src/java/org/apache/cassandra/service/MigrationManager.java
```diff
@@ -70,6 +71,9 @@ public class MigrationManager
     public static final int MIGRATION_DELAY_IN_MS = 60000;
     public static final int MAX_SCHEDULED_SCHEMA_PULL_REQUESTS = 3;
 
+    // These messaging service versions are known to have compatible schema formats so nodes on these versions are allowed to pull from each other
+    private static final ImmutableSet<Integer> SCHEMA_COMPATIBLE_VERSIONS_RRS = ImmutableSet.of(MessagingService.VERSION_22, MessagingService.VERSION_22_PLTR);
```
rhuffy: nit:
```suggestion
    private static final Set<Integer> SCHEMA_COMPATIBLE_VERSIONS_RRS = ImmutableSet.of(MessagingService.VERSION_22, MessagingService.VERSION_22_PLTR);
```

rhuffy (approved):

PR #716, opened 2025-06-08
Experiment with New Optimized Collation Method

d-guo:
This PR modifies the filter experiment to test a new collation method that emits cells covered by a range tombstone from the iterator. This is one way to make the optimized collation method compatible with resumable range scans https://github.com/palantir/cassandra/pull/701. It does this by emitting cells that the previous method drops within the iterator. We need to keep these cells so they that they are counted in the `CountingCellIterator`, which is used for metrics as well as the resumable range scans threshold.

This PR also randomizes the order in which experiments are being run to avoid caching bias, which is suspected to exist currently.

### DIFF
--- a/src/java/org/apache/cassandra/FilterExperiment.java
+++ b/src/java/org/apache/cassandra/FilterExperiment.java
@@ -102,7 +102,7 @@ public enum FilterExperiment
         ColumnFamily optimizedEmitCellsResult = results.get(USE_OPTIMIZED_EMIT_CELLS);
 
         compareAndLogResults(legacyResult, optimizedResult, USE_OPTIMIZED, function, fallback);
-        compareAndLogResults(legacyResult, optimizedEmitCellsResult, USE_OPTIMIZED, function, fallback);
+        compareAndLogResults(legacyResult, optimizedEmitCellsResult, USE_OPTIMIZED_EMIT_CELLS, function, fallback);
 
         return legacyResult;
     }
--- a/src/java/org/apache/cassandra/FilterExperiment.java
+++ b/src/java/org/apache/cassandra/FilterExperiment.java
@@ -18,16 +18,20 @@
 
 package org.apache.cassandra;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Predicates;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
-import org.apache.cassandra.utils.Pair;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
--- a/src/java/org/apache/cassandra/FilterExperiment.java
+++ b/src/java/org/apache/cassandra/FilterExperiment.java
@@ -44,7 +44,7 @@ import org.apache.thrift.annotation.Nullable;
 
 public enum FilterExperiment
 {
-    USE_LEGACY, USE_OPTIMIZED;
+    USE_LEGACY, USE_OPTIMIZED, USE_OPTIMIZED_EMIT_CELLS;
 
     private static final Logger log = LoggerFactory.getLogger(FilterExperiment.class);
     private static final MetricNameFactory names = new DefaultNameFactory("FilterExperiment");
--- a/src/java/org/apache/cassandra/FilterExperiment.java
+++ b/src/java/org/apache/cassandra/FilterExperiment.java
@@ -18,14 +18,16 @@
 
 package org.apache.cassandra;
 
-import java.util.Iterator;
+import java.util.*;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Predicates;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
+import org.apache.cassandra.utils.Pair;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -52,12 +54,20 @@ public enum FilterExperiment
             CassandraMetricsRegistry.Metrics.timer(names.createMetricName("Legacy"));
     private static final Timer optimizedTimer =
             CassandraMetricsRegistry.Metrics.timer(names.createMetricName("Optimized"));
+    private static final Timer optimizedEmittingTimer =
+            CassandraMetricsRegistry.Metrics.timer(names.createMetricName("OptimizedEmitting"));
     private static final Counter successes =
             CassandraMetricsRegistry.Metrics.counter(names.createMetricName("Successes"));
     private static final Counter failures =
             CassandraMetricsRegistry.Metrics.counter(names.createMetricName("Failures"));
     private static final Counter indeterminate =
             CassandraMetricsRegistry.Metrics.counter(names.createMetricName("Indeterminate"));
+    private static final Counter successesForEmitting =
+            CassandraMetricsRegistry.Metrics.counter(names.createMetricName("SuccessesForEmitting"));
+    private static final Counter failuresForEmitting =
+            CassandraMetricsRegistry.Metrics.counter(names.createMetricName("FailuresForEmitting"));
+    private static final Counter indeterminateForEmitting =
+            CassandraMetricsRegistry.Metrics.counter(names.createMetricName("IndeterminateForEmitting"));
 
     public static ColumnFamily execute(
             Function<FilterExperiment, ColumnFamily> function,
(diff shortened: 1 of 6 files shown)


src/java/org/apache/cassandra/db/filter/QueryFilter.java
```diff
@@ -322,6 +317,98 @@ private boolean pendingTombstoneNotDroppable() {
...
+
+                    // we have a pending range tombstone, and the incoming atom is either a range tombstone or a cell
+                    // if it is a cell, it is in range of the range tombstone
+                    // if it is a range tombstone, it starts before the pending range tombstone ends
+                    // in this case, we may be able to merge the incoming range tombstone if it is droppable and
+                    // fully contained within the pending range tombstone's range
+                    if (nextAtom instanceof Cell)
+                    {
+                        // even if the next cell is overwritten by the range tombstone, still emit it
+                        // this is the key difference of this method vs filterTombstones
```
rynorris, line 376: Is this the only difference? Could they just be the same method but with a flag?

src/java/org/apache/cassandra/FilterExperiment.java
```diff
@@ -44,57 +46,101 @@
...
-            } else {
-                failures.inc();
-                log.warn("Comparison failure while experimenting; Legacy: {}, Optimized: {}, Comparison method: {}, Legacy metadata: {}, Optimized metadata: {}",
-                         legacyResult, optimizedResult, SafeArg.of("comparisonMethod", initialComparison.name()), safeLoggableColumnFamilyMetadata("legacyMetadata", legacyResult), safeLoggableColumnFamilyMetadata("optimizedMetadata", optimizedResult));
+
+        List<FilterExperiment> experiments = Arrays.asList(FilterExperiment.values());
+        Collections.shuffle(experiments); // shuffle to prevent caching bias
+
+        Map<FilterExperiment, ColumnFamily> results = new HashMap<>();
+        for (FilterExperiment experiment : experiments)
```
rynorris, line 87: Are we concerned at all with harming performance running all 3 of these? Should we rather just pick one experiment to run each time?

src/java/org/apache/cassandra/FilterExperiment.java
```diff
@@ -44,57 +46,101 @@
...
-                //  long as we still have signal if we do hit it.
-                log.warn("Query result changed under immediate compaction but results from 60 seconds ago are identical, result is indeterminate; Legacy: {}, Optimized: {}, Legacy metadata: {}, Optimized metadata: {}",
-                         legacyResult, optimizedResult, safeLoggableColumnFamilyMetadata("legacyMetadata", legacyResult), safeLoggableColumnFamilyMetadata("optimizedMetadata", optimizedResult));
-            } else {
-                failures.inc();
-                log.warn("Comparison failure while experimenting; Legacy: {}, Optimized: {}, Comparison method: {}, Legacy metadata: {}, Optimized metadata: {}",
-                         legacyResult, optimizedResult, SafeArg.of("comparisonMethod", initialComparison.name()), safeLoggableColumnFamilyMetadata("legacyMetadata", legacyResult), safeLoggableColumnFamilyMetadata("optimizedMetadata", optimizedResult));
+
+        List<FilterExperiment> experiments = Arrays.asList(FilterExperiment.values());
+        Collections.shuffle(experiments); // shuffle to prevent caching bias
```
lucaskjaero, line 84: Are we caching? My understanding was that we could only use this in cases where we were explicitly not using the cache.

src/java/org/apache/cassandra/db/filter/QueryFilter.java
d-guo: it's not the only difference: the new method doesn't use a peeking iterator because it can decide whether or not to emit an atom on the first pass

src/java/org/apache/cassandra/FilterExperiment.java
d-guo: you're right that we don't do any row caching. this was talking more about os level caching like page caching or something similar

src/java/org/apache/cassandra/FilterExperiment.java
d-guo: i'll test it on some stacks with high load before rolling it out

i like the direct comparisons we get from this approach

my thought is it's probably fine since most queries are small enough that this will succeed. and even if it does fail, the client will retry and most likely not run the experiment again

src/java/org/apache/cassandra/FilterExperiment.java
lucaskjaero: Wouldn't we expect page caching to apply to all experiments then? My mental model here is that we're changing how we process the page but not actually what we read off of disk.

d-guo: you're right that we don't preventing page caching from happening. before we were always running the legacy method first, then optimized, and page caching could explain why the optimized method seemed to be consistently faster than the legacy

this PR makes the three run in a random order so on average, they all benefit from the page cache the same amount

src/java/org/apache/cassandra/FilterExperiment.java
```diff
@@ -44,57 +46,101 @@
...
             CassandraMetricsRegistry.Metrics.timer(names.createMetricName("Optimized"));
+    private static final Timer optimizedEmittingTimer =
+            CassandraMetricsRegistry.Metrics.timer(names.createMetricName("OptimizedEmitting"));
     private static final Counter successes =
             CassandraMetricsRegistry.Metrics.counter(names.createMetricName("Successes"));
     private static final Counter failures =
             CassandraMetricsRegistry.Metrics.counter(names.createMetricName("Failures"));
     private static final Counter indeterminate =
             CassandraMetricsRegistry.Metrics.counter(names.createMetricName("Indeterminate"));
+    private static final Counter successesForEmitting =
```
lucaskjaero, line 69: Did you allowlist these? Indeterminate is not allowlisted for example so we don't actually get these metrics.

d-guo: yep! these are all allowlisted by default now

src/java/org/apache/cassandra/FilterExperiment.java
```diff
@@ -18,14 +18,16 @@
 
 package org.apache.cassandra;
 
-import java.util.Iterator;
+import java.util.*;
```
inespot: Nit: import should use full path

src/java/org/apache/cassandra/db/filter/QueryFilter.java
```diff
@@ -322,6 +317,98 @@ private boolean pendingTombstoneNotDroppable() {
...
+                        // we do have a pending range tombstone and the incoming atom is either a range tombstone or a cell
+                        // if it is a range tombstone, it starts after the pending range tombstone ends and therefore is not in range
+                        // if it is a cell, it is after the pending range tombstone ends and therefore not in range / covered
+                        if (nextAtom instanceof RangeTombstone)
+                        {
+                            // always immediately return the range tombstone to maintain ordering
+                            // even though we will hold onto it to merge with other range tombstones
+                            // this is because we will emit cells covered by this range tombstone after
+                            setPendingRangeTombstone((RangeTombstone) nextAtom);
+                            return maybePendingRangeTombstone;
```
inespot, line 359: Practically, I wonder if we we would be able to just merge tombstones, if they are intersecting. Although considering what timestamp AtlasDB uses to emit them, we should mostly fall into the superseding case and it might not actually be worth reasoning about intersections.

inespot:
Can you detail the description of this PR (or the next internal one) to explain why we need to retain those cells for future reference?

src/java/org/apache/cassandra/db/filter/QueryFilter.java
```diff
@@ -322,6 +317,98 @@ private boolean pendingTombstoneNotDroppable() {
...
+                    }
+                    else
+                    {
+                        RangeTombstone tombstone = (RangeTombstone) nextAtom;
+
+                        // if the incoming range tombstone is droppable and superceded by the pending range tombstone,
+                        // we expect that we can skip emitting the incoming range tombstone
+                        if (maybePendingRangeTombstone.supersedes(tombstone, comparator) && tombstoneIsDroppable(tombstone))
+                        {
+                            continue;
```
inespot, line 387: So this is really the only true optimization we get in with this method? Dropping superseding tombstones early

src/java/org/apache/cassandra/db/filter/QueryFilter.java
```diff
@@ -322,6 +317,98 @@ private boolean pendingTombstoneNotDroppable() {
...
+                    // we have a pending range tombstone, and the incoming atom is either a range tombstone or a cell
+                    // if it is a cell, it is in range of the range tombstone
+                    // if it is a range tombstone, it starts before the pending range tombstone ends
+                    // in this case, we may be able to merge the incoming range tombstone if it is droppable and
+                    // fully contained within the pending range tombstone's range
+                    if (nextAtom instanceof Cell)
+                    {
+                        // even if the next cell is overwritten by the range tombstone, still emit it
+                        // this is the key difference of this method vs filterTombstones
+                        return nextAtom;
```
inespot, line 377: In the case where you have a couple cells covered by the same range tombstone, you could theoretically only the last one, correct? Which would still get you the pageToken you need? 
You also do not need to return already processed droppable range tombstones that covered no cells.
Maybe as a V1, I am not actually sure how much memory we are optimizing here by dropping additional cells?

src/java/org/apache/cassandra/FilterExperiment.java
```diff
@@ -44,57 +46,101 @@
...
+        if (initialComparison.isEqual())
+        {
+            getSuccessCounter(experiment).ifPresent(Counter::inc);
+        }
+        else if (!areTrulyEqual(legacyResult, function.apply(USE_LEGACY)))
+        {
+            // This case means that the data was modified between getting the legacy and optimized result. As a result, this experiment is indeterminate.
+            getIndeterminateCounter(experiment).ifPresent(Counter::inc);
+        }
+        else if ((legacyResult.metadata().getGcGraceSeconds() == 0
```
inespot, line 128: Is this case trying to account for compactions between two successive queries? If yes, this should be an issue in the general case, correct? If so, why the `legacyResult.metadata().getGcGraceSeconds() == 0` condition?

src/java/org/apache/cassandra/db/filter/QueryFilter.java
```diff
@@ -322,6 +317,98 @@ private boolean pendingTombstoneNotDroppable() {
...
+                        if (maybePendingRangeTombstone.supersedes(tombstone, comparator) && tombstoneIsDroppable(tombstone))
+                        {
+                            continue;
+                        }
+                        else
+                        {
+                            // cannot optimize so we return the tombstone
+                            // however we only want to set the pending range tombstone to the incoming range tombstone
+                            // if the incoming one's end comes after the pending one's end
+                            if (!maybePendingRangeTombstone.includes(tombstone, comparator))
```
inespot, line 394: This is an optimization correct? Not a correctness fix to the original JB experiment?

src/java/org/apache/cassandra/db/filter/QueryFilter.java
d-guo: the idea makes sense to me. also agree with your statement that with atlas's deletion patterns we most will likely just supersede the next tombstone

we should keep this in mind. i think there's probably more optimizations similar to your idea that we can do on this code path but it's nonobvious how useful they would actually be, so i think this is out of scope for this PR

src/java/org/apache/cassandra/db/filter/QueryFilter.java
d-guo: yep, this probably helps a good bit considering our deletion pattern w atlas tho

src/java/org/apache/cassandra/FilterExperiment.java
d-guo: i took this code from the original experiment. i spent some time to try to understand it but didn't really get anymore. my goal in this PR was just to introduce the new filtering method

src/java/org/apache/cassandra/db/filter/QueryFilter.java
d-guo: ah you still need to return all cells. otherwise they aren't counted by the `CountingCellIterator` which is necessary to count toward the resumable range scan threshold (and metrics)

src/java/org/apache/cassandra/FilterExperiment.java
d-guo: i took this code directly from the original experiment — tbh i don't really understand what it's doing but my goal in this PR was just to introduce the new filtering method and not really examine the original experimentation method

src/java/org/apache/cassandra/db/filter/QueryFilter.java
d-guo: i think this might be a correctness fix, but it's a bit difficult to concretely reasonable about. i think the best way to know is just to roll out the experiment

inespot: Got it, not blocking to experiment but would be curious for us to dig more into this

src/java/org/apache/cassandra/FilterExperiment.java
inespot: Did we end up testing this? If my understanding is correct, this might add 50% more latency on that code path for 1/100 queries which seems not that negligible 🤔

